PR #3, opened 2017-04-10
Removed ERT building from travis.

joakim-hove:


.travis.yml
```diff
@@ -49,7 +49,7 @@ install:
     - conda update -q conda
     # Useful for debugging any issues with conda
     - conda info -a
-    - conda install pyqt=4 scipy=0.16.1 numpy pandas matplotlib future
```
pgdr: we can also remove pandas, can't we?  I thought that's only used in UI.

.travis.yml
```diff
@@ -58,7 +58,7 @@ env:
 before_script:
   - mkdir build
   - cd build
-  - cmake -DBUILD_TESTS=ON -DBUILD_PYTHON=ON -DERT_BUILD_CXX=ON ..
+  - cmake -DBUILD_TESTS=ON -DBUILD_PYTHON=ON -DERT_BUILD_CXX=ON -DBUILD_APPLICATIONS=ON ..
```
pgdr: Sorry for removing this.  What's applications, again?

joakim-hove: Applications is just a set of small utilities for working with binary eclipse files, could probably to a large extent be replaced by 10 line Python scripts.

PR #5, opened 2017-04-18
Pylint testing for travis

pgdr:
This PR adds linting as a test case.

PR #6, opened 2017-04-18
Removed ping testing based on server list.

joakim-hove:


pgdr:
I wonder if it's not travis that blocks pinging to `"localhost"`.

Anyway, I don't know why the `util_ping` test isn't run, when it's run in #5 

This will also invalidate my commit b3605e5

pgdr:
merged!

PR #10, opened 2017-04-19
WIP lint test all of ert/ecl

pgdr:
work in progress while we lint the rest of ert/ecl


pgdr:
Okay, there is a bug in `pylint` (at PyCQA/pylint#779) which blocks this PR.

PR #14, opened 2017-04-21
avoid leaking memory in EclFilename.cpp

joakim-hove:
the memory returned by ecl_util_alloc_filename() must be
free()'d. this solution is obviously not very performant and also not
very elegant, but it seems to work.

this was found by address sanitizer.


pgdr:
Forward ported from Ensembles/ert#1493

Issue #16, opened 2017-04-21
CMake errors out when requesting static libs even if python is not available


blattms:
I do not have the requested python version on my system. Nevertheless CMake issues an error when run with -DBUILD_SHARED_LIBS=OFF
```
CMake Error at CMakeLists.txt:140 (message):
  The Python wrappers require shared libraries


-- Configuring incomplete, errors occurred!

joakim-hove:
Thank you - will look into.

markusdregi:
You should disable python if you want to run it without Python. Please reopen if you need any further assistance.

PR #17, opened 2017-04-21
Do not search for exact 2.7 python version

### LINKED ISSUE #15: Is requiring exact 2.7 python version really intended?
CMake deactivates python on my system with
```
-- Found PythonInterp: /usr/bin/python (found version "2.7.9") 
-- Could NOT find PythonInterp: Found unsuitable version "2.7.9", but required is exact version "2.7" (found /usr/bin/python)
```
Is this intended?

blattms:
Closes #15 
BTW I am not sure whether this will find python3 or not and what would be intended here.

### DIFF
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -1,7 +1,7 @@
 include(cmake/find_python_module.cmake)
 include(cmake/python_package_versions.cmake) # finds version
 
-FIND_PACKAGE(PythonInterp 2.7 EXACT)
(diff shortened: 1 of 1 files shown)


joakim-hove:
Thank you - but this will happily find Python3; which we can not (yet) use. However  - I think #18 should fix it, also including #16.

joakim-hove:
Observe that if you use this library with OPM you must *install* it - when https://github.com/OPM/opm-common/pull/236 has been merged you can also this library in "sibling mode".

PR #18, opened 2017-04-21
Python 2.7

joakim-hove:


CMakeLists.txt
```diff
@@ -179,12 +168,16 @@ endif()
 include_directories( ${PROJECT_SOURCE_DIR}/libecl_well/include )
 add_subdirectory( libecl_well )
 
-
 if (BUILD_PYTHON)
    if (ERT_WINDOWS)
       message(STATUS "Python is not supported on Windows")
```
pgdr: Could you replace `STATUS` with either `WARNING` or `SEND_ERROR`?

python/CMakeLists.txt
```diff
@@ -1,16 +1,24 @@
...
 include(cmake/python_package_versions.cmake) # finds version
 
-FIND_PACKAGE(PythonInterp 2.7 EXACT)
+FIND_PACKAGE(PythonInterp)
 if (NOT DEFINED PYTHON_EXECUTABLE)
-   message("Python2.7 interpreter not found - Python wrappers not enabled")
+   message(WARNING "Python2 interpreter not found - Python wrappers not enabled")
+   set( BUILD_PYTHON OFF PARENT_SCOPE )
+   return()
+endif()
```
pgdr: I don't think  `Python2` makes sense.  We don't support Python 2.6, and there will never be a Python 2.8, hence Python 2.7 is the only real alternative.  It should say `"Python 2.7 interpreter not found"`.

python/CMakeLists.txt
```diff
@@ -1,16 +1,24 @@
...
+
+if (NOT "${PYTHON_VERSION_MAJOR}.${PYTHON_VERSION_MINOR}" STREQUAL "2.7")
+   message(WARNING "Need Python version 2.7.x, found version: ${PYTHON_VERSION_MAJOR}.${PYTHON_VERSION_MINOR} - Python wrappers not enabled")
```
pgdr: `Python version 2.7.x` = `Python 2.7`.

joakim-hove:
OK - fixups applied and force pushed.

PR #23, opened 2017-04-25
Added __repr__ to ExtendedTestCase

joakim-hove:


joakim-hove:
This is trivial - I'll self merge!

PR #24, opened 2017-04-25
Extendedtestcase: __str__ instead of __repr__

joakim-hove:
Seems it will not climb the class hierarchy looking for `__repr__` - makes sense.

### DIFF
--- a/python/python/ert/test/extended_testcase.py
+++ b/python/python/ert/test/extended_testcase.py
@@ -56,10 +56,10 @@ class ExtendedTestCase(TestCase):
         self.__share_root = None
         installAbortSignals()
         super(ExtendedTestCase , self).__init__(*args , **kwargs)
-        self.__str__ = self.__repr__
 
-    def __repr__(self):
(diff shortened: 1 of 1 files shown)


python/python/ert/test/extended_testcase.py
```diff
@@ -56,10 +56,10 @@ def __init__(self , *args , **kwargs):
         self.__share_root = None
         installAbortSignals()
         super(ExtendedTestCase , self).__init__(*args , **kwargs)
-        self.__str__ = self.__repr__
 
-    def __repr__(self):
-        return 'ExtendedTestCase( TESTADATA_ROOT=%s, SOURCE_ROOT=%s, SHARE_ROOT=%s, BUILD_ROOT=%s)' % (TESDATA_ROOT,
+
+    def __str__(self):
```
pgdr: This should be `__repr__`.  `str(obj)` will fallback to `__repr__` if `__str__` is not defined, not the other way.

joakim-hove: Well; I don't disagree with that - the problem was with inheritance:
```python
class Base:
     def __repr__(self):
           ...
    
Class Child(Base):

c = Child( )
print c  
```
The `print c`will *not* use `Base.__repr__( )` but rather fall back to the default `__repr__( )` representation - that was at least my understanding of the situation?

PR #25, opened 2017-04-27
WIP: Changes in running Python tests

joakim-hove:
Clean up of the cmake + python machinery. The intention is that `libres` (and other) modules should be able to use the cmake files from this module.


pgdr:
Except failing tests, good to merge.

PR #33, opened 2017-05-03
Added libecl-config file.

joakim-hove:


joakim-hove:
With this in place downstream modules should be able to do:
```
find_package( libecl )
```
and be done with it!

joakim-hove:
This has been reviewed in private email - merging.

PR #48, opened 2017-05-07
Added ert Python package shim.

joakim-hove:
Added ert/ package root which will transparently import all symbols
from the ecl/ root, with this existing code which imports from the ert
root package should continue to work.


joakim-hove:
OK - after some discussion I have been granted permission to merge this - I hope it will be a blessing.

PR #51, opened 2017-05-09
WIP Respect grid orientation in cell containment calculations

pgdr:
If a grid is "mirrored", i.e. (exactly one of) decreasing `i` index increases east coordinates, or decreasing `j` index increases north coordinates, the cell containment did not work correctly for points close to the walls.

PR #53, opened 2017-05-10
Ignored sign of tetrahedron decomposition

pgdr:


joakim-hove:
Is this essentially #51?

PR #54, opened 2017-05-11
Refactored cmake

jokva:
Refactors the makefiles in order to accomplish (primarily) three things:

* Leverage cmake dependency tracking and hacks to make cmake do the
  heavy lifting
* Allow cmake to generate good *-config.cmake export files to make
  libecl play nicer with other projects
* Allow libecl to use pycmake and other third-party cmake modules that
  expect a fairly vanilla cmake setup

A few guiding principles have been applied, most of which are described
in this blog post: https://rix0r.nl/blog/2015/08/13/cmake-guide/

In summary:
- All third party libraries are registered as cmake targets, which means
  no more magic ${LIBRARY_INCLUDE_DIRS}. Optional libraries are always
  targets and are always linked to, but may not actually contain any
  physical libraries. This makes dependency tracking nice and smooth and
  uniform, and should carry well to downstream users
- All libraries export targets, install and set up proper metadata both
  for install and build, with no hard coded paths. cmake keeps track of
  this for us.
- The top level cmake deals with global configuration, options, imports
  etc., and orchestrates includes and add_subdirectory.
- Individual libraries are still in their own directories, and only
  specify how to build and test themselves, and their dependencies. No
  option or include checking.
- One cmake file per component, not per directory. This leads to larger,
  but fewer files. cmake files are kept simply and mostly just list
  source files for a target.
- All applications are moved to the applications/ directory, instead of
  being organised under individual modules.

This is stil a WIP and is *not* ready for merge - I'm making it public for early feedback. The plan is to, unless this is nipped in the bud, to possibly introduce and cross-pollinate with pycmake and use that to generate the python package. The python sub library has not yet been touched.

This introduces a LONG list of changes (6k+ diff), but really the only interesting things to look at are the new, complete cmake files. Everything else can mostly be for now.


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "pycmake"]
```
joakim-hove: Is this just a preparation, or does this PR actually introduce a dependency on pycmake.

CMakeLists.txt
```diff
@@ -84,89 +86,206 @@ if (MSVC)
...
+add_library(openmp INTERFACE)
+if (OPENMP_FOUND AND ERT_USE_OPENMP)
+    target_compile_options(openmp INTERFACE OpenMP_C_FLAGS)
+else ()
+     set(ERT_USE_OPENMP OFF)
      message(STATUS "OpenMP package not found - OpenMP disabled")
-  endif()
+endif ()
+
+add_library(pthread INTERFACE)
```
joakim-hove: What does the `INTERFACE` imply?

.gitmodules
jokva: It does with the second commit in the series. It's trivial to postpone it to a separate PR though

joakim-hove: OK - I would prefer if this PR was C level only.

CMakeLists.txt
jokva: It's a pretty cool cmake feature covered slightly in the blog I linked. From cmake doc:

> Creates an Interface Library. An INTERFACE library target does not directly create build output, though it may have properties set on it and it may be installed, exported and imported

What it does it that it creates a target (like any other `add_library` call) but *without any source files*. Then include dirs, libraries etc are set on this target, and it's used as any other `add_library`, which means we can link to it with `target_link_libraries`. What's neat about it is that it sets up *transitive* linking and include directories, so if `ecl` depends on `util`, and `util` depends on, say `lapack`, then `ecl` also depends on lapack *and* linking and include dirs are set up accordingly.

joakim-hove: OK - that is nifty!

jokva:
Alright, I've updated this PR to just include the C-related changes.

joakim-hove:
OK - this is cool; it will be ready to merge when the windows build has been resurrected,

jokva:
The Windows build is working, actually, but travis dies for some reason.

However, I **think** this will break downstream builds as-is, so it warrants some more testing, including opm and internal tests.

jokva:
Two new things before the weekend:

1. For backwards compatibility, or at least until after the python revamp, non-default library and executable paths are reintroduced. It hopefully makes merging easier because some things seem to depend on it. That's why the `add_tests` are updated to take the `NAME <name> COMMAND <cmd>` form, because it will resolve targets, whereas the shorthand won't. It is its own commit for ease-of-revert.
2. I've added the `export(PACKAGE)` command, which means that when libecl is built from source on a development system, it's automatically registered with cmake so that downstream projects *won't* have to specify some dir for target imports. This will really only affect developers, but I think it's pretty cool. Demonstrated-ish in https://github.com/Statoil/libres/pull/6

PR #56, opened 2017-05-11
Extract grid

markusdregi:
The goal of this PR is to implement functionality for extracting a subgrid from a larger grid. This is to enable one to extract interesting cases from larger data files.

### DIFF
--- a/python/python/ecl/ecl/ecl_grid_generator.py
+++ b/python/python/ecl/ecl/ecl_grid_generator.py
@@ -94,8 +94,9 @@ class EclGridGenerator:
         return grid
 
     @classmethod
-    def createZcorn(cls, dims, dV, offset, escape_origo_shift,
-                    irregular_offset, irregular, concave, faults):
+    def createZcorn(cls, dims, dV, offset=1, escape_origo_shift=(1,1,0),
+            irregular_offset=False, irregular=False, concave=False,
+            faults=False):
 
         cls.__assertZcornParameters(dims, dV, offset, escape_origo_shift,
                     irregular_offset, irregular, concave, faults)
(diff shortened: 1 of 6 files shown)


python/python/ecl/ecl/ecl_grid_generator.py
```diff
@@ -20,6 +20,21 @@
 from ecl.util import IntVector
 from ecl.ecl import EclGrid, EclKW, EclDataType, EclPrototype
 
+def flatten(l):
+    return [elem for sublist in l for elem in sublist]
+
+def divide(l, size):
+    return [l[i:i+size:] for i in range(0, len(l), size)]
+
+def verbose(l):
```
joakim-hove: `verbose` was a really funny name for this function?

python/python/ecl/ecl/ecl_grid_generator.py
```diff
@@ -20,6 +20,21 @@
...
+    return [elem for elem in l for i in range(2)][1:-1:]
+
+def constructFloatKW(name, values):
+     kw = EclKW(name, len(values), EclDataType.ECL_FLOAT)
+     for i in range(len(values)):
```
joakim-hove: Alternatively:
```python
for i,value in enumerate(values):
    kw[i] = value
```

joakim-hove:
The integer field `ACTNUM` has `nx*ny*nz` values and is used to differentiate between *active* and *inactive* cells. `ACTNUM` is optional - if it is not present all cells are assumed to be active. I guess the `extract_grid` method could take an optional `ACTNUM` a sub copy of `ACTNUM` in addition to the `zcorn` and `coord` keywords.

PR #61, opened 2017-05-14
Fixed mem-leak caused by failing realloc

epa095:
If the call to realloc fails then realloc returns null, giving us no pointer to the original data in "work", giving us no way to free it. Introduced a temporary variable which lets us free the original data in case realloc fails.


joakim-hove:
Looks good - the failing test is a "one out of twenty fails" test - we ignore that now.

PR #63, opened 2017-05-15
Exposing mapaxes

markusdregi:
Exposes mapaxes in Python. Made the C-function consistent with the corresponding functions for the other keywords..

### DIFF
--- a/python/python/ecl/ecl/ecl_grid_generator.py
+++ b/python/python/ecl/ecl/ecl_grid_generator.py
@@ -559,18 +559,21 @@ class EclGridGenerator:
         actnum = grid.export_actnum()
         cls.assert_actnum(nx, ny, nz, actnum)
 
+        mapaxes = grid.export_mapaxes()
+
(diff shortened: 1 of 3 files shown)


python/python/ecl/ecl/ecl_grid_generator.py
```diff
@@ -559,18 +559,21 @@ def extract_subgrid(cls, grid, ijk_bounds,
...
                                     decomposition_change=decomposition_change,
-                                    translation=translation)
+                                    translation=translation
+                                    )
 
         sdim = tuple([b-a+1 for a,b in ijk_bounds])
         sub_coord, sub_zcorn, sub_actnum = sub_data
 
-        return EclGrid.create(sdim, sub_zcorn, sub_coord, sub_actnum)
+        return EclGrid.create(sdim, sub_zcorn, sub_coord, sub_actnum, mapaxes)
```
pgdr: please use named keyword: `..., mapaxes=mapaxes)`

PR #64, opened 2017-05-16
Mapaxes translation

markusdregi:
The translation given by the user is adjusted according to the MAPAXES of the grid.


python/python/ecl/ecl/ecl_grid_generator.py
```diff
@@ -34,6 +35,26 @@ def construct_floatKW(name, values):
         kw[i] = value
     return kw
 
+def pre_mapaxes_translation(translation, mapaxes):
+    if not mapaxes:
```
joakim-hove: `if mapaxes is None`?

PR #67, opened 2017-05-16
Require cmake >= 2.8.12

jokva:


joakim-hove:
1. Why do we need this.
2. I fear that the default travis build environment does not satisfy this requirement. 

@pgdr: I have a feeling this repo is not on the default travis build environment?

jokva:
1. `target_include_directories`
2. Travis supports it (with ppa for linux and native with trusty images, OS X is out of the box)

PR #68, opened 2017-05-16
Extract test

markusdregi:


python/tests/ecl/test_fk_user_data.py
```diff
@@ -0,0 +1,47 @@
...
+    def test_cell_containment(self):
+
+        grid_location = "local/ECLIPSE/faarikaal/faarikaal%d.EGRID"
+        well_location = "local/ECLIPSE/faarikaal/faarikaal%d.txt"
+
+        for i in range(1, 8):
+            grid_file = self.createTestPath(grid_location % i)
+            well_file = self.createTestPath(well_location % i)
+
+            grid = EclGrid.loadFromFile(grid_file)
```
joakim-hove: Ehhh - the `__init__( )`based: `grid = EclGrid( grid_file )` still works no?

python/tests/ecl/test_fk_user_data.py
```diff
@@ -0,0 +1,47 @@
...
+
+            # Load well data
+            with open(well_file, "r") as f:
+                lines = [line.split() for line in f.readlines()]
+                points = [map(float, line[:3:]) for line in lines]
```
joakim-hove: The two lines:
```python
points = ...
exp_cells = ...
```
could be taken out of the `with open() as f:` scope?

python/tests/ecl/test_fk_user_data.py
markusdregi: You're absolutely right! But then I think we should make EclGrid.loadFromFile private and deprecate the current public version of the function. This to be in line with the "only one obvious way to do it" Python mantra..

PR #75, opened 2017-05-22
Nnc new data structure

stefoss23:
New ecl_nnc_data_type. Can read TRANNNC_KW from init files.


libecl/src/ecl_nnc_data.c
```diff
@@ -0,0 +1,116 @@
...
+struct ecl_nnc_data_struct {
+   UTIL_TYPE_ID_DECLARATION;
+   int size;
+   double * values;
+};
+
+
+
+
+static void ecl_nnc_data_set_trans_simple(ecl_nnc_data_type * data, ecl_file_view_type * view_file) {
```
joakim-hove: This is not in use and should be removed?

libecl/src/ecl_nnc_data.c
```diff
@@ -0,0 +1,116 @@
...
+
+static ecl_kw_type * ecl_nnc_data_get_trangl_kw( const ecl_file_view_type * view_file, int lgr_nr ) {
+  ecl_kw_type * tran_kw = NULL;
+  return tran_kw;
+}
+
+
+static void ecl_nnc_data_set_trans(ecl_nnc_data_type * data, ecl_nnc_geometry_type * nnc_geo, ecl_file_view_type * view_file) {
+ 
+   int nnc_size = ecl_nnc_geometry_size( nnc_geo );
```
joakim-hove: Remove this line.

libecl/src/ecl_nnc_data.c
```diff
@@ -0,0 +1,116 @@
...
+   data->size = nnc_size;
+   data->values = util_malloc( data->size * sizeof(double));
+
+   ecl_kw_type * trannnc_kw;
```
joakim-hove: The declaration on line 59, the `if` test on line 61 and the `for` loop on line 65 should be reorganized into one unit,

libecl/src/ecl_nnc_data.c
```diff
@@ -0,0 +1,116 @@
...
+      else 
+         data->values[nnc_index] = -12; 
+      
+   }
+   
+
+}
+
+
+ecl_nnc_data_type * ecl_nnc_data_alloc_tran(ecl_nnc_geometry_type * nnc_geo, ecl_file_view_type * view_file) {
```
joakim-hove: Argument should have `init` in name - to communicate that this view should be from an `INIT` file.

libecl/src/ecl_nnc_data.c
```diff
@@ -0,0 +1,116 @@
...
+    return data->size;
+}
+
+double * ecl_nnc_data_get_values( const ecl_nnc_data_type * data ) {
+    return data->values;
+}
+
+
+double ecl_nnc_data_iget_value(const ecl_nnc_data_type * data, int index) {
+    return data->values[index];
```
joakim-hove: Add a guard;
```C
if (index < data->size)
   return data->values[index];
else
   util_abort("%s: index value:%d out range: [0,%d) \n",__func__ , index , data->size);
```

stefoss23:
I have adressed the changes.

libecl/tests/test_ecl_nnc_data.c
```diff
@@ -0,0 +1,94 @@
...
+         int nnc_data_size = ecl_nnc_data_get_size( nnc_geo_data );
+         test_assert_true(nnc_data_size == 3);
+
+         double * values = ecl_nnc_data_get_values( nnc_geo_data );
+         test_assert_true(values[0] == 0);
+         test_assert_true(values[1] == 1.5);
+         test_assert_true(values[2] == 3.0);
+         
+         ecl_nnc_data_free( nnc_geo_data );
+         ecl_file_view_free( view_file );
```
joakim-hove: The view shall *not* be freed - the `ecl_file` will discard the view when it is freed; observe that the view has been aquired with a *get* function, then it should (as a general rule) not be freed either.

libecl/src/ecl_nnc_data.c
```diff
@@ -0,0 +1,101 @@
...
+   free(data);
+}
+
+
+
+int ecl_nnc_data_get_size(ecl_nnc_data_type * data) {
+    return data->size;
+}
+
+double * ecl_nnc_data_get_values( const ecl_nnc_data_type * data ) {
```
joakim-hove: This should be `const double *`

libecl/tests/test_ecl_nnc_data.c
```diff
@@ -0,0 +1,94 @@
...
+         ecl_file_view_type * view_file = ecl_file_get_global_view( init_file );
+
+         test_assert_true( ecl_file_view_has_kw( view_file, TRANNNC_KW) );
+         
+         ecl_nnc_data_type * nnc_geo_data = ecl_nnc_data_alloc_tran(nnc_geo, view_file);
+         int nnc_data_size = ecl_nnc_data_get_size( nnc_geo_data );
+         test_assert_true(nnc_data_size == 3);
+
+         double * values = ecl_nnc_data_get_values( nnc_geo_data );
+         test_assert_true(values[0] == 0);
```
joakim-hove: For floating point comparison you should (typically) use `test_assert_double_equal( a , b )` which uses some heuristics for an approximate comparison.

libecl/src/ecl_nnc_data.c
```diff
@@ -0,0 +1,101 @@
...
+  ecl_kw_type * tran_kw = NULL;
+  return tran_kw;
+}
+
+
+static void ecl_nnc_data_set_trans(ecl_nnc_data_type * data, ecl_nnc_geometry_type * nnc_geo, ecl_file_view_type * init_file) {
+ 
+   data->size = ecl_nnc_geometry_size( nnc_geo );
+   
+   data->values = util_malloc( data->size * sizeof(double));
```
joakim-hove: The `date->size = ... ` and `data->values = ...` assignments should be in the `ecl_nnc_data_alloc_tran( )` function.

libecl/src/ecl_nnc_data.c
```diff
@@ -0,0 +1,101 @@
...
+
+   for (int nnc_index = 0; nnc_index < data->size; nnc_index++) {
+      ecl_nnc_pair_type * pair = ecl_nnc_geometry_iget( nnc_geo, nnc_index );
+      int grid1 = pair->grid_nr1;
+      int grid2 = pair->grid_nr2;
+      if (grid1 == grid2) {
```
joakim-hove: The body of this loop assumes that this is `grid1 == 0`- i.e. I would suggest:
```C
if ((grid1 == grid2) && (grid1 == 0)) {
...
}
```

PR #83, opened 2017-05-25
grid.get_lgr: Can do numbered lookup.

joakim-hove:


kudhru:
Thanks!

python/python/ecl/ecl/ecl_grid.py
```diff
@@ -78,7 +79,8 @@ class EclGrid(BaseCClass):
     _cell_contains                = EclPrototype("bool ecl_grid_cell_contains_xyz1( ecl_grid , int , double , double , double )")
     _cell_regular                 = EclPrototype("bool ecl_grid_cell_regular1( ecl_grid , int)")
     _num_lgr                      = EclPrototype("int  ecl_grid_get_num_lgr( ecl_grid )")
-    _has_lgr                      = EclPrototype("bool ecl_grid_has_lgr( ecl_grid , char* )")
+    _has_numered_lgr_             = EclPrototype("bool ecl_grid_has_lgr_nr( ecl_grid , int)")
```
markusdregi: Should be _has_numbered_lgr?

joakim-hove: Thank you - yes.

joakim-hove:
> it seems like the numbered version of the get_lgr-function is never tested. 

Yes - that is unfortunately correct - quite large job to create valid testdata. It is tested in the Statoil data (and that test had not been run .... ). Will look into it.

PR #88, opened 2017-05-29
Add OS X setup for Travis

markusdregi:
Adds a basic OS X setup. We might want to include several builds later on, but I think this is good for now..

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,17 @@
-dist: trusty
-sudo: false
 language: c
+
+matrix:
+  fast_finish: true
+  allow_failures:
+    - os: osx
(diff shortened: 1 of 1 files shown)


joakim-hove:
Hmmm - the title of the PR seems good; we want this - but I do not really understand why it fails.

markusdregi:
@joakim-hove I have this going as a side project and will have a second look at it as soon as possible. Should be easy to fix..

jokva:
You can have a look at `segyio`'s travis.yml which builds both linux and os x in multi configs:

https://github.com/Statoil/SegyIO/blob/master/.travis.yml

PR #90, opened 2017-05-30
Adding log level enums and a simplified log function

epa095:
Added the function:
 void log_add_message_str(log_type *logh, message_level_type message_level , const char* message);
which does not take in a dup_stream or a flag for freeing (it will not free the message as it is const)

### DIFF
--- a/libert_util/include/ert/util/log.h
+++ b/libert_util/include/ert/util/log.h
@@ -26,12 +26,23 @@ extern "C" {
 #include <stdbool.h>
 #include <stdlib.h>
 
+//Same as pythons default log levels, but with different numeric values.
+typedef enum {
+  LOG_CRITICAL=0, //OOM.
+  LOG_ERROR=1, //When something we really expected to work does not, e.g. IO failure.
+  LOG_WARNING=2, //Important, but not error. E.g. combination of settings which can be intended, but probably are not.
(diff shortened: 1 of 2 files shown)


libert_util/include/ert/util/log.h
```diff
@@ -26,12 +26,23 @@ extern "C" {
 #include <stdbool.h>
 #include <stdlib.h>
 
+//Same as pythons default log levels, with numeric values equal to python / 10.
```
epa095: I set them to python/10 to keep a bit of compability with the old levels, which is not defined anywhere, but it seems like 1 is used a lot, and that is also the default level. I am not able to see far enough in to the future to determine if it is better to change those and standarize on the same numeric values as python uses.

libert_util/src/log.c
```diff
@@ -128,6 +128,17 @@ bool log_include_message(const log_type *logh , int message_level) {
     return false;
 }
 
+/**
+ * Adds a string to the log if message_level is below the threshold
+ * @param logh The log  to add to
```
epa095: My IDE adds this doxygen formating, but I see that they are not present elsewhere in the code.

joakim-hove: No - and they are dang ugly :-)

PR #91, opened 2017-05-30
Track all source files & build only libecl

jokva:
This PR does two things:

1. Building is modified slightly so that every source file is tracked by cmake, instead of source files being `#include`d. This means everything will trigger rebuilds, and that every source file is explicitly built.
2. This is bigger. `libgeometry`, `libert_util`, and `libecl_well` are compiled into `libecl`, i.e. this module with this patch applied only builds into one target. This is intended to make downstream use easier, pip distribution simpler and tears down a somewhat misleading separation of the four libraries. If C++ is enabled, the C++ based code is also compiled into `libecl`.


joakim-hove:
OK - this *IS* a big one; I fast merged some small PR's; hopefully that does not create problems for you.

jokva:
Since it's mostly moves, rebase is usually painless. Anything else you want in?

joakim-hove:
> Anything else you want in?

No - this should be fine.

jokva:
'twas easy peasy

pgdr:
Good job!

PR #93, opened 2017-05-30
Install header files from source dir

jokva:


jokva:
This is pretty harmless and an embarrassing oversight, so I'll cowboy merge.

PR #94, opened 2017-05-31
Include Windows.h in util_env.c

jokva:
If Windows.h is not included, SetEnvironmentVariable linking does not
work as expected.


jokva:
Cowboy merge to fix some builds.

PR #98, opened 2017-06-01
PR template

markusdregi:
A suggestion for a very simple PR template for creating a pull request. It contains two sections, Task and Approach. In addition is has a check box that one is to check, confirming that the Statoil tests did pass.

I think this is a good idea, but I'm open to suggestions regarding the headers of the sections and possible extensions of the checklist.


joakim-hove:
How should this be used?

joakim-hove:
>How should this be used?

I understand now - good idea.

PR #103, opened 2017-06-06
WIP: Writing keywords with ecl_file on OS X fails

markusdregi:
#### Task
_The EclKW test fails on OS X._

#### Approach
_It turned out to be an problem with ecl_file/fortio in C. The problem occurred in ecl_file_save_kw and the solution was to flush the fortio stream after skipping the header, but before writing the data._

**Pre un-WIP checklist**

### DIFF
--- a/lib/ecl/fortio.c
+++ b/lib/ecl/fortio.c
@@ -488,6 +488,17 @@ void fortio_data_fseek(fortio_type* fortio, offset_type data_offset, size_t data
     }
 }
 
+int fortio_fclean(fortio_type * fortio) {
+  long current_pos = ftell(fortio->stream);
+  if(current_pos == -1)
+    return -1;
+
+  int flush_status = fflush(fortio->stream);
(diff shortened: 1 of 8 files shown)


lib/ecl/fortio.c
```diff
@@ -488,6 +488,17 @@ void fortio_data_fseek(fortio_type* fortio, offset_type data_offset, size_t data
     }
 }
 
+int fortio_fclean(fortio_type * fortio) {
+  long current_pos = ftell(fortio->stream);
+  if(current_pos == -1)
+    return -1;
+
+  int flush_status = fflush(fortio->stream);
+  if(flush_status != 0)
```
joakim-hove: Are the `ftell( )` and `fseek(  )` calls necessary?

markusdregi: Yes, because the position in the stream after a flush is undefined. And in particular, it results in funky behaviour with clang on OS X..

PR #105, opened 2017-06-06
Ecl unpack fixup

joakim-hove:
#### Task
The `ecl_unpack` application failed + it was not even built.

#### Approach
Minor cmake fix to ensure that it was built + fix bug in ecl_file_view / ecl_file usage.

#### Pre un-WIP checklist
- [x] All Statoil tests passes locally

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -306,6 +306,10 @@ if (ERT_LINUX)
 endif()
 
 add_subdirectory( lib )
+if (BUILD_APPLICATIONS)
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -306,6 +306,10 @@ if (ERT_LINUX)
 endif()
 
 add_subdirectory( lib )
+if (BUILD_APPLICATIONS)
+   add_subdirectory( applications )
```
jokva: Move logic inside of `applications`

PR #106, opened 2017-06-07
Python refactor + pip deploy

jokva:
#### Task

#### Approach

**Pre un-WIP checklist**


joakim-hove:
Hmmm; when I have checked out a clean version of this branch I get:

```txt
CMake Error: File /private/joaho/work/ERT/libecl/python/setup.py.in does not exist.
CMake Error at /private/joaho/work/ERT/install/share/cmake/Modules/PythonPackage.cmake:719 (configure_file):
  configure_file Problem configuring file
  python/CMakeLists.txt:192 (add_setup_py)


CMake Error: File /private/joaho/work/ERT/libecl/python/setup.py.in does not exist.
CMake Error at /private/joaho/work/ERT/install/share/cmake/Modules/PythonPackage.cmake:719 (configure_file):
  configure_file Problem configuring file
  python/CMakeLists.txt:193 (add_setup_py)
```

does that make sense?

joakim-hove:
Post and close - much the same. Sorry.

joakim-hove:
> Hmmm; when I have checked out a clean version of this branch I get:

Sorry - I checked out the wrong branch; seems you are actually developing on your own `master`?

jokva:
Yeah, I was for a bit anyway, mainly to get some travis logic to work.

joakim-hove:
Currently if you pass `-DCMAKE_PREFIX_PATH=/my-path` the Python loader will look for Python modules in `/my-path/lib/python${version}/site-package` - and that path will also be added to `PYTHONPATH` manipulations prior to testing. Is that behaviour retained in this PR, or do you have to set up `PYTHONPATH`correctly to find all Python dependencies before invoking `cmake`?

The current behavior has been convenient, but it is not essential to retain it.

joakim-hove:
> It's fairly reasonable to assume that the system building it has the right numpy (probably the big dependency in question)

That I agree; but for this case the main usage is to tell downstream modules, i.e. libres and ert where the libecl module is installed - that is typically done with `-DCMAKE_PREFIX_PATH` which will be used in `find_package( ecl )`to find headers and librarie(s); but it will not be used in the find python functionality.

joakim-hove:
@jokva: Can you rebase this - and I will merge it.

pgdr:
@jokva what is the status here?

PR #109, opened 2017-06-09
Nnc data w flux

stefoss23:
**Task**
Creating functionality in ecl_nnc_data for reading dynamic flux data. 

**Approach**
Existing functionality has been generalized to accept flux data. 
Note:
for nnc in which grid1 != 0, grid2 != 0, grid1 != grid2, ecl_nnc_data CANNOT read flux data. 

**Pre un-WIP checklist**

### DIFF
--- a/lib/include/ert/ecl/ecl_nnc_data.h
+++ b/lib/include/ert/ecl/ecl_nnc_data.h
@@ -27,6 +27,11 @@ extern "C" {
 #include <ert/ecl/ecl_nnc_geometry.h>
 #include <ert/ecl/ecl_nnc_export.h>
 
+#define TRANS_DATA     1
+#define WTR_FLUX_DATA  2
+#define OIL_FLUX_DATA  4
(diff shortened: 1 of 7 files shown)


lib/include/ert/ecl/ecl_nnc_data.h
```diff
@@ -27,9 +27,14 @@ extern "C" {
 #include <ert/ecl/ecl_nnc_geometry.h>
 #include <ert/ecl/ecl_nnc_export.h>
 
+#define TRANS_DATA     1
```
joakim-hove: 1. Move this to the `ecl_nnc_data` file - no need to let this leak into the public api.
2. Define an `enum` instead of just using `#define`
3. When you use a `2^N` series that is a very clear hint that these values can be used as a bitmask; as long as that is actually not the case I would prefer if you used the values `1,2,3,4`.

stefoss23:
There is no functionality for reading LL keywords other than for tranll_kw.

PR #113, opened 2017-06-14
Rewrote binary search in _get_index_from_sim_time, improve error log

pgdr:
**Task**
Debugging an error regarding `ecl_sum_data_init_interp_from_sim_time` with an error message `Hmm internal error??`.

Now we output `No ministep prior to %s (idx=%d)`.

**Approach**
* rewrote binary search to have a more classic layout
* improve error message in ecl_sum_data_init_interp_from_sim_time
* renamed _variable to variable in aforementioned function

**Pre un-WIP checklist**
- [x] Have completed graphical integration test steps

### DIFF
--- a/lib/ecl/ecl_sum_data.c
+++ b/lib/ecl/ecl_sum_data.c
@@ -526,10 +526,10 @@ bool ecl_sum_data_check_sim_days( const ecl_sum_data_type * data , double sim_da
 
 
 static int ecl_sum_data_get_index_from_sim_time( const ecl_sum_data_type * data , time_t sim_time) {
-  time_t data_start_time = time_interval_get_start( data->sim_time );
-  time_t sim_end         = time_interval_get_end( data->sim_time );
+  time_t data_start_time = time_interval_get_start(data->sim_time);
+  time_t sim_end = time_interval_get_end(data->sim_time);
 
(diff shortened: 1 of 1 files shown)


lib/ecl/ecl_sum_data.c
```diff
@@ -526,15 +526,15 @@ bool ecl_sum_data_check_sim_days( const ecl_sum_data_type * data , double sim_da
 
 
 static int ecl_sum_data_get_index_from_sim_time( const ecl_sum_data_type * data , time_t sim_time) {
-  time_t data_start_time = time_interval_get_start( data->sim_time );
-  time_t sim_end         = time_interval_get_end( data->sim_time );
+  time_t sim_start = time_interval_get_start(data->sim_time);
```
joakim-hove: There is a conceptual difference between `data_start_time` and `sim_start`. The summary case will always have a correct value for `sim_start` - but for restarted simulations there can be a finite time interval: `[sim_start, data_start_time]` where we do not have data. Hence I would prefer to keep the name `data_start_time`.

pgdr:
Thanks, @joakim-hove, I have addressed the comment.

joakim-hove:
Good - I saw nothing wrong with the code and it was certainly easier to read than what was there before, but I was a little worried about the the rest failures in libres.Have restarted Travis.

PR #117, opened 2017-06-18
Made datetime function public + added test.

joakim-hove:
**Task**
Previously static function made public.

**Approach**
Removed `static` + added simple test; and fixed a bug!

**Pre un-WIP checklist**

### DIFF
--- a/lib/include/ert/util/util.h
+++ b/lib/include/ert/util/util.h
@@ -99,10 +99,10 @@ typedef enum {left_pad   = 0,
               center_pad = 2} string_alignement_type;
 
   //#define UTIL_CXX_MALLOC(var , num_elm) (typeof (var)) util_malloc( (num_elm) * sizeof var)
-
   void         util_bitmask_on(int *  , int );
   char       * util_get_timezone(void);
   time_t       util_make_datetime_utc(int , int , int , int , int , int );
+  bool         util_make_datetime_utc_validated(int sec, int min, int hour , int mday , int month , int year, time_t * t);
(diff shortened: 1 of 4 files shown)


lib/include/ert/util/util.h
```diff
@@ -99,10 +99,10 @@ typedef enum {left_pad   = 0,
               center_pad = 2} string_alignement_type;
 
   //#define UTIL_CXX_MALLOC(var , num_elm) (typeof (var)) util_malloc( (num_elm) * sizeof var)
-
   void         util_bitmask_on(int *  , int );
   char       * util_get_timezone(void);
   time_t       util_make_datetime_utc(int , int , int , int , int , int );
+  bool         util_make_datetime_utc__(int sec, int min, int hour , int mday , int month , int year, time_t * t);
```
markusdregi: This is not a very good function name.. How about following the `util_set_....`-syntax used by some of the other util-functions?

joakim-hove: Well - the function name might not be very good; but there is some consistency to it:

- `util_make_xxxx` : `(year, day, month , [hour,min,sec]) -> time_t`
- `util_set_xxx` : `time_t -> (year , month , day , [hour , min, sec])`

Where the `util_set_xxx` use pointer references to "return" multiple values.

joakim-hove:
OK - the name is been changed, the new public function is: 
```C
bool util_make_datetime_utc_validated( ... , time_t * t)
```
which will return a `bool` if the input arguments correspond to a *valid* datetime, and the actual generated time will be in the last `time_t *t` argument.

PR #119, opened 2017-06-20
Fortio write failure

joakim-hove:
**Task**
Possible to mark a writable fortio instance as "failed" - it will be removed when closing the fortio instance.

**Approach**
Call to `unlink( )`

**Pre un-WIP checklist**


joakim-hove:
This is contained in #120

PR #127, opened 2017-06-23
Fix compiler warnings

markusdregi:
**Task**
_Get rid of compiler warnings._

**Approach**
_These fixes allows compiling libecl with -Werror=all._

**Pre un-WIP checklist**

### DIFF
--- a/lib/util/util_unlink.c
+++ b/lib/util/util_unlink.c
@@ -1,0 +1,37 @@
+/*
+   Copyright (C) 2017  Statoil ASA, Norway.
+
+   The file 'util_unlink.c' is part of ERT - Ensemble based Reservoir Tool.
+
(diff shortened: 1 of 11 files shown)


lib/util/util_unlink.c
```diff
@@ -0,0 +1,31 @@
...
+   for more details.
+*/
+
+#ifdef ERT_WINDOWS
+  #include <io.h>
+#else
+  #include <unistd.h>
+#endif
+
+int util_unlink(const char * filename) {
```
joakim-hove: Have tried to do these tests by looking for the existence of a specific function - i.e. `unlink( )` - instead of the catch all `ERT_WINDOWS` - see e.g.: https://github.com/Statoil/libecl/blob/master/lib/util/util.c#L3329

PR #133, opened 2017-06-27
Exported header should not include build_config.h

joakim-hove:
**Task**
The exported header `util_unlink.h` included the private header `build_config.h`

**Approach**
Moved #define symbols.

**Pre un-WIP checklist**

### DIFF
--- a/lib/util/util_unlink.c
+++ b/lib/util/util_unlink.c
@@ -34,4 +34,8 @@ int util_unlink(const char * filename) {
  return unlink(filename);
 }
 
+#else
+
+#error "No unlinke implementation on this platform"
+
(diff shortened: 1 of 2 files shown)


lib/util/util_unlink.c
```diff
@@ -34,4 +34,11 @@ int util_unlink(const char * filename) {
  return unlink(filename);
 }
 
+#else
+
```
markusdregi: Maybe we could just do an #error instead? To avoid it compiling all together if unlink is missing?

joakim-hove: Yes - that is better.

PR #135, opened 2017-06-30
Changes to subst_list.__getitem__( )

joakim-hove:
**Task**
Changed the `subst_list.__getitem__()` implementation to behave more like a regular map.

**Approach**
Just remapped some existing function prototypes.

**Pre un-WIP checklist**


markusdregi:
Do we even need to expose the `iget`-functions? How about just the `__getitem__` by `key` and then a method to expose all keys, just like a standard Python dictionary?

joakim-hove:
> Do we even need to expose the iget-functions? How about just the __getitem__ by key and then a method to expose all keys, just like a standard Python dictionary?

I was very in doubt whether to do as I did - or as you suggest. The reason I ended up there was for the returning of doc strings; but I don't know.

PR #147, opened 2017-07-28
Move files from libecl to libres

ErlendHaa:
Files to be moved to libres
    /lib/util/subst_list.c
    /lib/util/block.c
    /applications/block_fs
    
Headerfiles, tests and dependencies are also moved

**Task**
When splitting the ert repository into libecl and libres some files ended up in a slightly illogical place. The purpose of this task is to move files from the libecl repository to the libres repository. The files in question include:

/lib/util/subst_list.c
/lib/util/block.c
/applications/block_fs

**Approach**
Removed mentioned files with their include files and tests from libecl repo.  Some files in the libecl repo was dependent on /lib/util/subst_list.c. These files was also removed. The files in question include:

/util/template.c
/util/template_loop.c
template.h

All files are added to the libres repo.

**Pre un-WIP checklist**

**Depends on**
* Statoil/libres#68


joakim-hove:
This is a good start; the subst_list alsa has a Python wrapper: https://github.com/Statoil/libecl/blob/master/python/python/ecl/util/substitution_list.py with a corresponding test: https://github.com/Statoil/libecl/blob/master/python/tests/util/test_substitution_list.py which must be moved along.

joakim-hove:
In this wrapper: https://github.com/Statoil/libecl/blob/master/python/python/legacy/ert/util/__init__.py the 
```
from ecl.util import SubstitutionList
```
can be replaced with:
```
try:
    from res.util import SubstitutionList
except ImportError:
    pass
```

joakim-hove:
Maybe you have not enabled the Python wrappers? That must be done by passing:
```
-DBUILD_PYTHON=ON
```
when running cmake.

joakim-hove:
Good work!

PR #148, opened 2017-08-02
Removed from API util_fread_from_buffer

pgdr:
**Task**

The symbol `util_fread_from_buffer` is only used one place in all of the ert ecosystem, from `enkf_util_assert_something`.

That does not warrant its existence.  The implementation was three lines, and they have been moved up to call site.

This depends on  Statoil/libres#76.

**Pre un-WIP checklist**
- [X] Statoil tests pass locally
- [X] libres Statoil tests pass locally

### DIFF
--- a/lib/util/buffer.c
+++ b/lib/util/buffer.c
@@ -194,98 +194,52 @@ void buffer_clear( buffer_type * buffer ) {
 }
 
 
-/*****************************************************************/
-/**
-    Observe that it is the functions with _safe_ in the name which
-    most closely mimicks the behaviour of fread(), and fwrite() -
-    these functions will *NOT* abort if the buffer is to small,
-    instead they will just return the number of items read/written,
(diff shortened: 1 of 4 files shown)


lib/util/buffer.c
```diff
@@ -194,98 +194,52 @@ void buffer_clear( buffer_type * buffer ) {
...
-    return values.
-
-    The functions buffer_fread() and buffer_fwrite() will abort if
-    read/write to buffer failed.
-*/
-
-
-
-static size_t buffer_fread__(buffer_type * buffer , void * target_ptr , size_t item_size , size_t items, bool abort_on_error) {
+size_t buffer_fread(buffer_type * buffer,
```
markusdregi: The return value of this function gives no information as the function either `util_aborts`, or returns the number of items requested. I suggest to make the function `void`.

PR #151, opened 2017-08-07
Initialize var to remove incompatible pointer type

pgdr:
**Task**

Jenkins is complaining about errors.  Initialize an `unsigned int`.


pgdr:
I cannot reopen?

joakim-hove:
Seems GitHub considers #152 as an equivalent replacement.

PR #152, opened 2017-08-07
WIP Remove jenkins warning

pgdr:
@joakim-hove Can we remove the `/dev/random` and `/dev/urandom` from `util.c` and use `rand()`?

### DIFF
--- a/lib/util/rng.c
+++ b/lib/util/rng.c
@@ -121,8 +121,8 @@ void rng_init( rng_type * rng , rng_init_mode init_mode ) {
     case(INIT_DEV_RANDOM):
       util_fread_dev_random( rng->state_size * sizeof * seed_buffer , seed_buffer );
       break;
-    case(INIT_DEV_URANDOM):
-      util_fread_dev_urandom( rng->state_size * sizeof * seed_buffer , seed_buffer );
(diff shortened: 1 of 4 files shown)


joakim-hove:
> Can we remove the /dev/random and /dev/urandom from util.c and use rand()?

Well - we need some "random" way to seed the generator; just calling `rand( )` will create the same sequence every time.

lib/util/rng.c
```diff
@@ -236,8 +219,8 @@ double rng_get_double( rng_type * rng ) {
 }
 
 int rng_get_int( rng_type * rng , int max_value ) {
-  rng_safe_cast( rng );
-  return rng->forward( rng->state ) % max_value;
+  int x = (int) util_dev_urandom_seed();
```
joakim-hove: Ehhhh - no no; the `rng_get_int( )` function should get *the next* integer in a deterministic fashion, from the stateful `rng_type` instance. The proposed changes will just get a "truly random" value - that is *not* what we want.

joakim-hove:
Hmmm - must say I am bit skeptical to this one; the changes when it comes to RNG behavior are not just warning polish.

joakim-hove:
What is wrong with `/dev/random` and `/dev/urandom`?

pgdr:
removing this playground mumble, never intended to be seen

PR #153, opened 2017-08-10
util: don't crash when a NULL buffer is passed to sscanf methods

andreabrambilla:
When passing a NULL buffer to sscanf functions in util, the software crashes.
I've added a check at the beginning of those functions so that, if the input buffer is NULL, they simply return false

**Pre un-WIP checklist**

### DIFF
--- a/lib/util/util.c
+++ b/lib/util/util.c
@@ -1481,6 +1481,9 @@ char * util_fscanf_alloc_token(FILE * stream) {
 */
 
 bool util_sscanf_double(const char * buffer , double * value) {
+  if(!buffer)
+    return false;
+
   bool value_OK = false;
   char * error_ptr;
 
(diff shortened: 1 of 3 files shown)


joakim-hove:
Please add tests - e.g. here: https://github.com/Statoil/libecl/blob/master/lib/util/tests/ert_util_sscan_test.c

lib/util/util.c
```diff
@@ -1504,7 +1507,10 @@ bool util_sscanf_double(const char * buffer , double * value) {
    Base 8
 */
 
-bool util_sscanf_octal_int(const char * buffer , unsigned int * value) {
+bool util_sscanf_octal_int(const char * buffer , int * value) {
```
andreabrambilla: @joakim-hove I am not completely sure about this change. I think using `int` is the correct solution (for consistency with other methods, and I see no reason why octal should not be negative). On the other hand, I found at least one place in libres where this is used with a variable of type mode_t (which is unsigned - this leads only to a warning, things seems to work fine). 

Any suggestion? Maybe we could add a second version `util_sscanf_octal_uint` that handles unsigned int

joakim-hove: Well - I would prefer if it were unchanged, the only current use for it is for reading modes.

joakim-hove:
The testing in this PR is *very good*; do what you think best regarding the octal parsing - and apart from that: Just merge it!

andreabrambilla:
(y)

I don't think I have write access to the repo. Can you please either merge it or give me the suitable access rights? thanks!

joakim-hove:
I'll merge.

PR #155, opened 2017-08-11
WIP: Ecl file cache

joakim-hove:
**Task**
When loading an ECLIPSE binary we start by seeking through the whole file and building an index. In the case of very large ( ~ 500GB ) this takes prohibitively long time, and the task is to create a on-disk cache of the index. This is mainly for Resinsight.

**Approach**
Added functions to dump an index, and to open a file from an on-disk index.

**Pre un-WIP checklist**

### DIFF
--- a/lib/ecl/ecl_file_kw.c
+++ b/lib/ecl/ecl_file_kw.c
@@ -135,7 +135,7 @@ UTIL_IS_INSTANCE_FUNCTION( ecl_file_kw , ECL_FILE_KW_TYPE_ID )
 
 
 
-static ecl_file_kw_type * ecl_file_kw_alloc__( const char * header , ecl_data_type data_type , int size , offset_type offset) {
+ecl_file_kw_type * ecl_file_kw_alloc0( const char * header , ecl_data_type data_type , int size , offset_type offset) {
   ecl_file_kw_type * file_kw = util_malloc( sizeof * file_kw );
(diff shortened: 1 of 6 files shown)


lib/ecl/ecl_file_kw.c
```diff
@@ -323,4 +336,88 @@ void ecl_file_kw_inplace_fwrite( ecl_file_kw_type * file_kw , fortio_type * fort
...
 
 
+void ecl_file_kw_fwrite( const ecl_file_kw_type * file_kw , FILE * stream ) {
+  int header_length = strlen( file_kw->header );
+  for (int i=0; i < ECL_STRING8_LENGTH; i++) {
+    if (i < header_length)
+      fputc( file_kw->header[i], stream );
+    else
+      fputc( ' ' , stream );
+  }
```
andreabrambilla: Just to help me understand... by reading the piece of code above, I got the impression that the header_length is expected to be less than  ECL_STRING8_LENGTH (which seems to be 8). I was wondering, does the header have usually a specific length?

joakim-hove: The header will have `len(header) <= ECL_STRING8_LENGTH`. For this particular we pad with spaces if the header is shorter - to guarantee that the on-disk size of one `ecl_file_kw_type` instance will have a deterministic size - that should lead to better performance when many (~ 100000) instances should be loaded.

PR #158, opened 2017-08-16
util: fixed tests failing when conflicting file is found in /tmp

andreabrambilla:
**Task**
Fix a test_path_context since it fails when /tmp/file already exists and the user has no access to it

**Approach**
~~If the file exist, create the file required for the test in a subfolder of /tmp with a pseudorandom name~~
Use a dedicated TestAreaContext for performing the test

**Pre un-WIP checklist**

### DIFF
--- a/python/tests/util/test_path_context.py
+++ b/python/tests/util/test_path_context.py
@@ -3,18 +3,29 @@ from ecl.test import ExtendedTestCase, PathContext,TestAreaContext
 
 
 class PathContextTest(ExtendedTestCase):
+    
     def test_error(self):
-        with self.assertRaises(OSError):
-            with PathContext("/usr/lib/testing"):
-                pass
-
-        with open("/tmp/file" , "w") as f:
-            f.write("xx")
-
-        with self.assertRaises(OSError):
-            with PathContext("/tmp/file"):
-                pass
-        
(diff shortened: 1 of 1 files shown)


python/tests/util/test_path_context.py
```diff
@@ -8,12 +9,29 @@ def test_error(self):
             with PathContext("/usr/lib/testing"):
                 pass
 
-        with open("/tmp/file" , "w") as f:
-            f.write("xx")
+        tmp_dir = None
```
joakim-hove: Can I suggest: 
```Python
import getpass
...
tmp_dir = os.path.join("/tmp" , getpass.user())
tmp_file = os.path.join( tmp_dir , "file")
```

python/tests/util/test_path_context.py
```diff
@@ -26,10 +37,6 @@ def test_chdir(self):
     def test_cleanup(self):
         with TestAreaContext("pathcontext"):
             os.makedirs("path/1")
-
-            with self.assertRaises(OSError):
-                with PathContext("path/1"):
-                    pass
```
andreabrambilla: Removed since it's redundant now

andreabrambilla:
@joakim-hove fixed as we discussed, please have a look when you have time ;)

PR #159, opened 2017-08-16
Ecl file cache (continued)

stefoss23:
Task
When loading an ECLIPSE binary we start by seeking through the whole file and building an index. In the case of very large ( ~ 500GB ) this takes prohibitively long time, and the task is to create a on-disk cache of the index. This is mainly for Resinsight.

Approach
Added functions to dump an index, and to open a file from an on-disk index.

Pre un-WIP checklist

 Statoil tests pass locally

### DIFF
--- a/lib/ecl/ecl_file.c
+++ b/lib/ecl/ecl_file.c
@@ -1048,14 +1048,12 @@ ecl_file_type * ecl_file_fast_open(const char * file_name, const char * index_fi
       ecl_file_kw_type * file_kw;
       for (int i = 0; i < index_size; i++) {
         file_kw = ecl_file_kw_fread_alloc( istream );
-        printf(" ********************************** %s: %s\n", __func__, ecl_file_kw_get_header(file_kw) );
+        fprintf(stderr, " ********************************** %s: %s\n", __func__, ecl_file_kw_get_header(file_kw) );
    
         ecl_file_view_add_kw( ecl_file_index->global_view , file_kw );
       }   
       ecl_file_view_make_index( ecl_file_index->global_view );
       ecl_file_select_global(ecl_file_index);
-
-      ecl_file_view_check_flags( ecl_file_index->flags , ECL_FILE_CLOSE_STREAM);     //debug
 
     {
     
--- a/lib/ecl/ecl_file.c
+++ b/lib/ecl/ecl_file.c
@@ -569,8 +569,8 @@ void ecl_file_select_global( ecl_file_type * ecl_file ) {
 */
 
 
-ecl_file_type * ecl_file_open( const char * filename , int flags) {
-  fortio_type * fortio;
+static fortio_type * ecl_file_alloc_fortio(const char * filename, int flags) {
+  fortio_type * fortio = NULL;
   bool          fmt_file;
 
   ecl_util_fmt_file( filename , &fmt_file);
@@ -579,6 +579,13 @@ ecl_file_type * ecl_file_open( const char * filename , int flags) {
     fortio = fortio_open_readwrite( filename , fmt_file , ECL_ENDIAN_FLIP);
   else
     fortio = fortio_open_reader( filename , fmt_file , ECL_ENDIAN_FLIP);
+
+  return fortio;
+}
+
+
+ecl_file_type * ecl_file_open( const char * filename , int flags) {
+  fortio_type * fortio = ecl_file_alloc_fortio(filename, flags);
 
   if (fortio) {
     ecl_file_type * ecl_file = ecl_file_alloc_empty( flags );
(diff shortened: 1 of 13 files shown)


ANerd:
I don't know if this is relevant, but I just want to point out that ECLIPSE already provides INSPEC/RSSPEC files with very similar information, including the field name, type, size and file offsets of all the fields in the init/restart files.

joakim-hove:
> but I just want to point out that ECLIPSE already provides INSPEC/RSSPEC files with very similar information

Yes - I know; although I must admit I had forgotten them. I think we will stick to the current approach, changing the format of the underlying indexfile should be quite simple.

lib/ecl/ecl_file.c
```diff
@@ -992,3 +992,75 @@ static ecl_file_type * ecl_file_iopen_rstblock__( const char * filename , int se
...
+}
+
+
+ecl_file_type * ecl_file_fast_open(const char * file_name, const char * index_file_name, int flags) {
+  if ( !(   util_file_exists( file_name ) && util_file_exists (index_file_name)   )   )
+    return NULL;
+
+  FILE * istream = util_fopen(index_file_name , "r");
+  char * file_name_ptr = util_fread_alloc_string(istream);
+  ecl_file_type * ecl_file_index = NULL;
```
joakim-hove: This is not an index - and should be named `ecl_file`.

lib/ecl/ecl_file.c
```diff
@@ -992,3 +992,75 @@ static ecl_file_type * ecl_file_iopen_rstblock__( const char * filename , int se
...
+    
+    }
+      
+      ecl_file_index = ecl_file_alloc_empty( flags );
+      ecl_file_index->fortio = NULL;
+      ecl_file_index->global_view = ecl_file_view_alloc( ecl_file_index->fortio , &ecl_file_index->flags , ecl_file_index->inv_view , true );   
+    
+      int index_size = util_fread_int(istream);
+      ecl_file_kw_type * file_kw;
+      for (int i = 0; i < index_size; i++) {
```
joakim-hove: Instead of multiple calls to `ecl_file_kw_fread_alloc( istream )` we should have *one* `call to ecl_file_kw_fread_alloc_multiple( istream , index_size)`:
```C
ecl_file_kw_type ** file_kw_list = ecl_file_file_kw_fread_alloc_multiple( istream, index_size);
for (int i=0; i < index_size; i++)
     ecl_file_view_add_kw(ecl_file_index->global_view , file_kw_list[i]);
```

lib/ecl/ecl_file.c
```diff
@@ -992,3 +992,75 @@ static ecl_file_type * ecl_file_iopen_rstblock__( const char * filename , int se
...
+        file_kw = ecl_file_kw_fread_alloc( istream );
+        fprintf(stderr, " ********************************** %s: %s\n", __func__, ecl_file_kw_get_header(file_kw) );
+   
+        ecl_file_view_add_kw( ecl_file_index->global_view , file_kw );
+      }   
+      ecl_file_view_make_index( ecl_file_index->global_view );
+      ecl_file_select_global(ecl_file_index);
```
joakim-hove: The lines 1045 - 1056 can be replaced with one function `ecl_file_view_fread_alloc( FILE * stream )`  and then that return value can be assigned to `ecl_file->global_view`.

lib/ecl/ecl_file.c
```diff
@@ -992,3 +992,75 @@ static ecl_file_type * ecl_file_iopen_rstblock__( const char * filename , int se
 ecl_file_type * ecl_file_iopen_rstblock( const char * filename , int seqnum_index , int flags) {
   return ecl_file_iopen_rstblock__(filename , seqnum_index , flags );
 }
+
+void  ecl_file_write_index( ecl_file_type * ecl_file , const char * filename , const char * index_filename) {
```
joakim-hove: This meat of this method should be in a function `ecl_file_view_write_index( const ecl_file_view_type * file_view, FILE * ostream)`.

lib/ecl/ecl_file.c
```diff
@@ -992,3 +992,75 @@ static ecl_file_type * ecl_file_iopen_rstblock__( const char * filename , int se
...
+  fclose( ostream );
+}
+
+
+ecl_file_type * ecl_file_fast_open(const char * file_name, const char * index_file_name, int flags) {
+  if ( !(   util_file_exists( file_name ) && util_file_exists (index_file_name)   )   )
+    return NULL;
+
+  FILE * istream = util_fopen(index_file_name , "r");
+  char * file_name_ptr = util_fread_alloc_string(istream);
```
joakim-hove: Remove the `_ptr`suffix - we have no tradition for using that; call it e.g. `source_file`.

lib/ecl/ecl_file.c
```diff
@@ -992,3 +992,75 @@ static ecl_file_type * ecl_file_iopen_rstblock__( const char * filename , int se
...
+  ecl_file->active_view = current_ecl_file_view;
+  
+  fclose( ostream );
+}
+
+
+ecl_file_type * ecl_file_fast_open(const char * file_name, const char * index_file_name, int flags) {
+  if ( !(   util_file_exists( file_name ) && util_file_exists (index_file_name)   )   )
+    return NULL;
+
```
joakim-hove: This should have a check of timestamps - to verify that the index file is newer than the source file:
```
if (util_file_difftime( file_name , index_file_name) > 0)
   return NULL;
```

joakim-hove: The existence of files - and timestamp checking - should be factored out in a separate function `bool ecl_file_index_valid( )` which can be used from calling scope.

lib/ecl/ecl_file.c
stefoss23: I cannot find that function.

joakim-hove: No - you should take the code on lines ~1045-1056 and assemble it in the new function `ecl_file_view_fread_alloc( )`

stefoss23: I think ecl_file_view_fread_alloc(..) should also contain ecl_file_view_alloc(..).

lib/ecl/ecl_file.c
```diff
@@ -992,3 +999,42 @@ static ecl_file_type * ecl_file_iopen_rstblock__( const char * filename , int se
 ecl_file_type * ecl_file_iopen_rstblock( const char * filename , int seqnum_index , int flags) {
   return ecl_file_iopen_rstblock__(filename , seqnum_index , flags );
 }
+
+bool ecl_file_index_valid(const char * file_name, const char * index_file_name) {
+  if ( !(   util_file_exists( file_name ) && util_file_exists (index_file_name)   )   )
```
joakim-hove: Simpler to read:
```C
if (util_file_exists( file_name ))
   return false;
if (!util_file_exists( index_file_name))
   return false;
```

lib/ecl/ecl_file.c
```diff
@@ -992,3 +999,42 @@ static ecl_file_type * ecl_file_iopen_rstblock__( const char * filename , int se
...
+    return false;
+  if (util_file_difftime( file_name , index_file_name) > 0)
+    return false;
+  return true;
+}
+
+void  ecl_file_write_index( ecl_file_type * ecl_file , const char * filename , const char * index_filename) {
```
joakim-hove: Not necessary to take in the `filename` argument - the `ecl_file` instance knows it's own filename.

lib/ecl/ecl_file_view.c
```diff
@@ -777,3 +777,34 @@ ecl_file_view_type * ecl_file_view_add_summary_view( ecl_file_view_type * file_v
...
+
+  int index_size = util_fread_int(istream);
+  ecl_file_kw_type ** file_kw_list = ecl_file_kw_fread_alloc_multiple( istream, index_size);
+  if (file_kw_list) {
+    ecl_file_view_type * file_view = ecl_file_view_alloc( fortio , flags , inv_map , true ); 
+    for (int i=0; i < index_size; i++)
+      ecl_file_view_add_kw(file_view , file_kw_list[i]);
+
+    ecl_file_view_make_index( file_view );
+    return file_view;
```
joakim-hove: Need a `free( file_kw_list);`

lib/ecl/tests/test_ecl_file_index.c
```diff
@@ -0,0 +1,99 @@
...
+      ecl_file_close( ecl_file_index );
+      
+   }
+   test_work_area_free( work_area );
+}
+
+
+int main( int argc , char ** argv) {
+   util_install_signals();
+   test_load_nonexisting_file();
```
joakim-hove: Could add the the following tests:
1. `test_assert_false( ecl_file_index_valid( "CASE" , "index_does_not_exist"));`
2. `test_assert_false( ecl_file_index_valid( "source_file_does_not_exists, "index"));`
3. `test_assert_false( ecl_file_index_valid( "CASE_IS_NEWER_THAN_INDEX" , "index"));`

joakim-hove:
Add `index_file` as an optional argument to the `EclFile.__init__( )` constructor and a `EclFile::write_index( )` method - so that this functionality is also available from Python.

PR #162, opened 2017-08-21
Make util compilable with C++

jokva:
**Task**

Compile `util` with `g++`

**Approach**

Renamed all `util/*.c` files to `.cpp` and compiled with `g++`. Lots of errors, most of them crying about implicit `void*` casts.

As a consequence, some `extern C`s are added and fixed.

The patch isn't warning free with C++, but does compile and tests do work.

**Pre un-WIP checklist**
- [X] Statoil tests pass locally

### DIFF
--- a/lib/include/ert/util/matrix_stat.h
+++ b/lib/include/ert/util/matrix_stat.h
@@ -22,6 +22,10 @@
 
 #include <ert/util/matrix.h>
 
+#ifdef __cplusplus
+extern "C" {
+#endif //__cplusplus
+
 typedef enum {
   LLSQ_SUCCESS = 0,
   LLSQ_INVALID_DIM = 1,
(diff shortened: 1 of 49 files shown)


joakim-hove:
Will it continue to build and compile with C - or is this a one-way-street?

jokva:
It's still valid C, and it will be for quite a while.

jokva:
Oh, and I should add, the first commit in this series fixes a problem with the `version_*` functions returning non-const pointers to static strings, which is a violation in C++. I don't _think_ it is in C, but it's pretty bad practice anyway.

joakim-hove:
As I understand it: These are changes to the source and header files which allow compilation with a C++ compiler, but the build system will still compile as C?  In that case this should be merged right away - as soon as the merge conflict is resolved.

jokva:
> As I understand it: These are changes to the source and header files which allow compilation with a C++ compiler, but the build system will still compile as C? In that case this should be merged right away - as soon as the merge conflict is resolved.

100% correct.

jokva:
Travis didn't want to restart.

PR #164, opened 2017-08-25
Allow empty variables in CMake string functions

magnesj:
Allow empty variables in CMake string functions

See
https://stackoverflow.com/questions/26585513/cmake-error-with-string-sub-command-strip-requires-two-arguments


CMakeLists.txt
```diff
@@ -17,7 +17,7 @@ set( ECL_VERSION_MICRO git ) # with "new in Ert Version X.X.X"!
 # development version leading towards version MAJOR.MINOR.0
 
 execute_process(COMMAND date "+%Y-%m-%d %H:%M:%S" OUTPUT_VARIABLE RES_BUILD_TIME )
-string(STRIP ${RES_BUILD_TIME} RES_BUILD_TIME)
+string(STRIP "${RES_BUILD_TIME}" RES_BUILD_TIME)
```
joakim-hove: When you are here anyway - could you just please fix a bug `RES_BUILD_TIME` -> `ECL_BUILD_TIME`

PR #170, opened 2017-08-31
WIP: Ecl big file

stefoss23:
**Task**
When using well_info_add_UNRST_wells() on very large (60 GB+) files memory overload may occur. 

**Approach**
The function  well_info_add_UNRST_wells() uses ecl_file_view_iget_named_kw. When the get_kw functions are used, data from the file is stored in the file_kw objetcs owned by ecl_file and retained in memory. 

The approach is to add the following functionality: It will be possible to obtain ecl_kw from ecl_file or eci_file_view as an independent object. This can be done in two different ways:
- simply using alloc functions
- using a flag ECL_FILE_RETURN_COPY when using ecl_file_open. This will cause ecl_file / ecl_file_view to 
  return independent ecl_kw objects using get functions. 

**Pre un-WIP checklist**

### DIFF
--- a/lib/ecl/ecl_file_kw.c
+++ b/lib/ecl/ecl_file_kw.c
@@ -244,6 +244,14 @@ static void ecl_file_kw_load_kw( ecl_file_kw_type * file_kw , fortio_type * fort
   }
 }
 
+ecl_kw_type * ecl_file_kw_alloc_kw( ecl_file_kw_type * file_kw, fortio_type * fortio) {
+  if (fortio == NULL)
+    util_abort("%s: trying to load a keyword after the backing file has been detached.\n",__func__);
+
+  fortio_fseek( fortio , file_kw->file_offset , SEEK_SET );
+  return ecl_kw_fread_alloc( fortio );   
+}
+
 ecl_kw_type * ecl_file_kw_get_kw_ptr( ecl_file_kw_type * file_kw , fortio_type * fortio , inv_map_type * inv_map ) {
   return file_kw->kw;
 }
--- a/lib/ecl/ecl_file_kw.c
+++ b/lib/ecl/ecl_file_kw.c
@@ -244,7 +244,7 @@ static void ecl_file_kw_load_kw( ecl_file_kw_type * file_kw , fortio_type * fort
   }
 }
 
-ecl_kw_type * ecl_file_kw_alloc_kw( ecl_file_kw_type * file_kw, fortio_type * fortio) {
+ecl_kw_type * ecl_file_kw_alloc_kw( const ecl_file_kw_type * file_kw, fortio_type * fortio) {
   if (fortio == NULL)
     util_abort("%s: trying to load a keyword after the backing file has been detached.\n",__func__);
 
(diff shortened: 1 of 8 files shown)


lib/ecl/tests/ecl_file.c
```diff
@@ -83,10 +83,58 @@ void test_truncated() {
...
+    fortio_fclose(fortio); 
+    //finished creating data file
+
+    ecl_file_type * ecl_file = ecl_file_open(file_name, ECL_FILE_RETURN_COPY);
+    ecl_kw_type * kw1_copy = ecl_file_iget_kw( ecl_file , 0 );
+    ecl_kw_type * kw2_copy = ecl_file_iget_kw( ecl_file , 1 );
+    ecl_kw_type * kw_alloc1 = ecl_file_alloc_kw( ecl_file , 0 );
+    ecl_kw_type * kw_named = ecl_file_iget_named_kw( ecl_file, "TEST1_KW", 0);
+    ecl_kw_type * kw_alloc2 = ecl_file_alloc_named_kw( ecl_file, "TEST2_KW", 0);
+
```
joakim-hove: One more test: `test_assert_pointer_not_equal( kw1_copy, kw_named)`

lib/ecl/ecl_file_kw.c
```diff
@@ -244,6 +244,19 @@ static void ecl_file_kw_load_kw( ecl_file_kw_type * file_kw , fortio_type * fort
   }
 }
 
+ecl_kw_type * ecl_file_kw_alloc_kw( const ecl_file_kw_type * file_kw, fortio_type * fortio) {
```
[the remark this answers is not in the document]
stefoss23: The ecl_file_kw_alloc_kw and ecl_file_get_kw  are the functions from getting ecl_kw via the ecl_file_kw object. They cannot be static.

lib/include/ert/ecl/ecl_file.h
```diff
@@ -38,7 +38,8 @@ extern "C" {
 
 #define ECL_FILE_FLAGS_ENUM_DEFS \
   {.value =   1 , .name="ECL_FILE_CLOSE_STREAM"}, \
-  {.value =   2 , .name="ECL_FILE_WRITABLE"}
+  {.value =   2 , .name="ECL_FILE_WRITABLE"}, \
+  {.value =   4 , .name="ECL_FILE_RETURN_COPY"}.
 #define ECL_FILE_FLAGS_ENUM_SIZE 2
```
pgdr: should this be 3?

joakim-hove: No - these values are used as bitmasks, so it should indeed be 4 - or 8, or ...

pgdr: Joakim, I'm thinking about this:
```c
#define ECL_FILE_FLAGS_ENUM_SIZE 2
```

joakim-hove: Ahhh yes - that is right :-)

joakim-hove:
The approach has changed.

PR #175, opened 2017-09-05
WIP Transaction type

stefoss23:
**Task**
Added suggested useful object to monitor use of ecl_kw_type objects. 

**Pre un-WIP checklist**

### DIFF
--- a/lib/ecl/ecl_file_kw.c
+++ b/lib/ecl/ecl_file_kw.c
@@ -247,6 +247,8 @@ static void ecl_file_kw_load_kw( ecl_file_kw_type * file_kw , fortio_type * fort
 }
 
 ecl_kw_type * ecl_file_kw_get_kw_ptr( ecl_file_kw_type * file_kw , fortio_type * fortio , inv_map_type * inv_map ) {
+  if (file_kw->kw)
+    file_kw->ref_count++;
   return file_kw->kw;
 }
 
@@ -268,7 +270,8 @@ ecl_kw_type * ecl_file_kw_get_kw_ptr( ecl_file_kw_type * file_kw , fortio_type *
 ecl_kw_type * ecl_file_kw_get_kw( ecl_file_kw_type * file_kw , fortio_type * fortio , inv_map_type * inv_map ) {
   if (file_kw->kw == NULL)
     ecl_file_kw_load_kw( file_kw , fortio , inv_map);
-
+  if(file_kw->kw)
+    file_kw->ref_count++;
   return file_kw->kw;
 }
 
@@ -425,10 +428,6 @@ ecl_file_kw_type * ecl_file_kw_fread_alloc( FILE * stream ) {
   return file_kw;
 }
 
-
-void ecl_file_kw_increment_ref_count(ecl_file_kw_type * file_kw) {
-  file_kw->ref_count++;
-}
 
 int ecl_file_kw_get_ref_count(ecl_file_kw_type * file_kw) {
   return file_kw->ref_count;
--- a/lib/ecl/ecl_file_kw.c
+++ b/lib/ecl/ecl_file_kw.c
@@ -424,3 +424,12 @@ ecl_file_kw_type * ecl_file_kw_fread_alloc( FILE * stream ) {
 
   return file_kw;
 }
+
+
+void ecl_file_kw_increment_ref_count(ecl_file_kw_type * file_kw) {
+  file_kw->ref_count++;
+}
+
+int ecl_file_kw_get_ref_count(ecl_file_kw_type * file_kw) {
+  return file_kw->ref_count;
+}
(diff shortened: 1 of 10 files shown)


lib/ecl/ecl_file_kw.c
```diff
@@ -422,3 +424,12 @@ ecl_file_kw_type * ecl_file_kw_fread_alloc( FILE * stream ) {
 
   return file_kw;
 }
+
+
+void ecl_file_kw_increment_ref_count(ecl_file_kw_type * file_kw) {
```
joakim-hove: This should *not* be a publicly visible function. The use of a reference counter is an implementation detail.

lib/ecl/ecl_file_kw.c
```diff
@@ -422,3 +424,12 @@ ecl_file_kw_type * ecl_file_kw_fread_alloc( FILE * stream ) {
...
+  file_kw->ref_count++;
+}
+
+int ecl_file_kw_get_ref_count(ecl_file_kw_type * file_kw) {
```
joakim-hove: Should not be visible. We should not test reference counts - but rather the final finished functionality.

lib/ecl/ecl_file_view.c
```diff
@@ -159,9 +157,16 @@ ecl_kw_type * ecl_file_view_iget_kw( const ecl_file_view_type * ecl_file_view ,
         fortio_fclose_stream( ecl_file_view->fortio );
     }
   }
+  if (ecl_kw)
+    ecl_file_kw_increment_ref_count(file_kw);
```
joakim-hove: This should just be `file_kw->ref_count++;`

lib/ecl/ecl_file_view.c
```diff
@@ -146,9 +146,7 @@ void ecl_file_view_add_flag( ecl_file_view_type * file_view , int flag)  {
   *file_view->flags |= flag;
 }
 
-
-ecl_kw_type * ecl_file_view_iget_kw( const ecl_file_view_type * ecl_file_view , int index) {
-  ecl_file_kw_type * file_kw = ecl_file_view_iget_file_kw( ecl_file_view , index );
+static ecl_kw_type * ecl_file_view_get_kw(const ecl_file_view_type * ecl_file_view, ecl_file_kw_type * file_kw) {
```
joakim-hove: The test here should use the ref_count:
```C
if (file_kw->ref_count == 0) {
     // Get ecl_kw from file.
}
```

lib/include/ert/ecl/ecl_file_kw.h
```diff
@@ -56,6 +56,9 @@ typedef struct inv_map_struct inv_map_type;
   void                ecl_file_kw_fwrite( const ecl_file_kw_type * file_kw , FILE * stream );
   ecl_file_kw_type ** ecl_file_kw_fread_alloc_multiple( FILE * stream , int num);
   ecl_file_kw_type *  ecl_file_kw_fread_alloc( FILE * stream );
+
+  void                ecl_file_kw_increment_ref_count(ecl_file_kw_type * file_kw);
+  int                 ecl_file_kw_get_ref_count(ecl_file_kw_type * file_kw);
```
joakim-hove: Do not want these out of `ecl_file_kw` To be able to do that the implementation of `ecl_file_transaction` must be in the `ecl_file_view.c` file.

lib/include/ert/ecl/transaction.h
```diff
@@ -0,0 +1,40 @@
...
+
+#include <ert/ecl/ecl_file_view.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+typedef struct transaction_struct transaction_type;
+
+transaction_type * transaction_start(ecl_file_view_type * file_view);
```
joakim-hove: Call it e.g. `ecl_file_transaction` not just `transaction`.

joakim-hove: No need to have both `transaction_end()` and `transaction_free()` - it should be enough with one of them.

lib/ecl/transaction.c
```diff
@@ -0,0 +1,56 @@
...
+}
+
+int transaction_iget_value(transaction_type * tran, int index) {
+  return tran->value[index];
+}
+
+void transaction_end(transaction_type * tran) {
+  for (int i = 0; i < tran->size; i++) {
+    ecl_file_kw_type * file_kw = ecl_file_view_iget_file_kw( tran->file_view, i);
+    tran->value[i] = ecl_file_kw_get_ref_count(file_kw) - tran->value[i];
```
joakim-hove: This is kind of oppsite; when the transaction is ended it is the `ecl_file_kw` instance which should be modified. Both the creation and the destruction of the transaction object can be from the `ecl_file_view` class:

```C
ecl_file_transaction_type * t = ecl_file_view_start_transaction( file_view );
...
...
ecl_file_view_end_transaction( file_view , t );
```

PR #179, opened 2017-09-10
Added cwrap have_function test.

joakim-hove:
Proper implementation is here: https://github.com/Statoil/cwrap/pull/7

### DIFF
--- a/python/python/cwrap/prototype.py
+++ b/python/python/cwrap/prototype.py
@@ -71,7 +71,7 @@ class PrototypeError(Exception):
 class Prototype(object):
     pattern = re.compile(PROTOTYPE_PATTERN)
 
-    def __init__(self, lib, prototype, bind=False):
+    def __init__(self, lib, prototype, bind=False, allow_attribute_error=False):
         super(Prototype, self).__init__()
         self._lib = lib
         self._prototype = prototype
@@ -79,6 +79,8 @@ class Prototype(object):
         self._func = None
         self.__name__ = prototype
         self._resolved = False
+        self._allow_attribute_error = allow_attribute_error
+
 
 
     def _parseType(self, type_name):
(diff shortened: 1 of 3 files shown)


jepebe:
Suggestion: 

Handling of **not implemented** can be moved to the `Prototype` class. For example add a keyword argument to the prototyping, such as `raise_not_implemented=True`, and then `raise` on **calling** of the method instead of `AttributeError` while prototyping. 

The benefit of this is that you do not need to check for`None` and do the conditional prototyping in the end of the file.

joakim-hove:
> The benefit of this is that you do not need to check forNone and do the conditional prototyping in the end of the file.

I agree - I also considered a solution along the lines you suggested. Have implemented that now.

python/python/ecl/util/matrix.py
```diff
@@ -68,8 +70,8 @@ def matmul(cls, m1,m2):
             return cls._alloc_matmul( m1, m2)
         else:
             raise ValueError("Matrix size mismathc")
```
jepebe: **mismatch** ← spelling

python/python/cwrap/prototype.py
```diff
@@ -148,7 +152,11 @@ def __call__(self, *args):
             self._resolved = True
 
         if self._func is None:
-            raise PrototypeError("Prototype has not been properly resolved!")
+            if self._allow_attribute_error:
+                raise NotImplementedError("Function:%s has not been properly resolved - missing library symbol?" % self.__name__)
+            else:
+                raise PrototypeError("Prototype not proper")
```
jepebe: I think it would be nicer to have the old message in the `PrototypeError` or something else that is a bit more informative.

joakim-hove:
All review comments adressed - merging.

PR #181, opened 2017-09-11
Stringlist unique

joakim-hove:
**Task**
Small utility function required in libres to check if all elements in a stringlist are unique.

**Approach**
O(n^2) sorting :-1: 

**Pre un-WIP checklist**

### DIFF
--- a/lib/util/stringlist.c
+++ b/lib/util/stringlist.c
@@ -790,6 +790,25 @@ int stringlist_append_matching_elements(stringlist_type * target , const stringl
   return stringlist_append_matching_elements( target , src , pattern );
 }
 
+bool stringlist_unique(const stringlist_type * stringlist )
+{
+  bool unique = true;
+  stringlist_type * cpy = stringlist_alloc_shallow_copy(stringlist);
+
+  stringlist_sort(cpy, strcmp);
+  for (int i = 0; i < stringlist_get_size(cpy) - 1; i++) {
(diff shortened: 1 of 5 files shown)


lib/util/stringlist.c
```diff
@@ -790,6 +790,26 @@ int stringlist_append_matching_elements(stringlist_type * target , const stringl
...
+
+  for (int index1=0; index1 < size - 1; index1++) {
+    const char * s1 = stringlist_iget( stringlist , index1 );
+    for (int index2=index1 + 1; index2 < size; index2++) {
+      const char * s2 = stringlist_iget( stringlist , index2 );
+      if (strcmp(s1,s2) == 0)
+        return false;
+    }
+  }
+  return true;
```
pgdr:
```c
bool stringlist_unique(const stringlist_type* lst)
{
  stringlist* cpy = stringlist_alloc_shallow_copy(lst);
  stringlist_sort(cpy, strcmp);
  int size = stringlist_get_size(cpy);
  for (int i = 0; index1 < size - 1; i++) {
    const char* s1 = stringlist_iget(cpy, i);
    const char* s2 = stringlist_iget(cpy, i+1);
    if (strcmp(s1,s2) == 0)
      return false;
  }
  return true;
}
```
Wouldn't this be almost simpler to write and read, _and_ you get `O(n log n)`?

pgdr:
Sweet.  Ready to merge.

PR #183, opened 2017-09-13
Allow Python 3 failure

markusdregi:
**Task**
_Parallel branch of https://github.com/Statoil/libecl/pull/182._

**Approach**
_Remove `;` as separators in `env matrix` as they become a part of the variable. Ensure that the match between set environment variables and the set up one wants to exclude is a perfect match._

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,13 +16,15 @@ env:
   global:
     - ERT_SHOW_BACKTRACE=1
   matrix:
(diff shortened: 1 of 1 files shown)


markusdregi:
I believe that the builds that are allowed to fail will start to fail when https://github.com/Statoil/ert/pull/73 is merged. After that, this PR is ready for review + merging as well.

PR #185, opened 2017-09-20
Fix a few buildsystem issues

akva2:
**Task**
While building packages I ran into a few buildsystem issues.

**Approach**
Fix according to my understanding of what options are meant to imply. In particular, the BUILD_ECL_SUMMARY used to imply building only that application, while currently it's a switch to disable/enable the single application, requiring all others to be built. Also fix some OpenMP linker issues.

**Pre un-WIP checklist**

### DIFF
--- a/applications/CMakeLists.txt
+++ b/applications/CMakeLists.txt
@@ -79,6 +79,12 @@ if (BUILD_ECL_SUMMARY)
 endif()
 
 if(apps)
+     if (ERT_USE_OPENMP)
+        foreach(app ${apps})
+            target_compile_options(${app} PUBLIC ${OpenMP_C_FLAGS})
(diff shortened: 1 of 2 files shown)


akva2:
the openmp fix is probably not the way to do it. currently, the gomp library isn't even linked into the main library. i see you use some OpenMP_EXE_LINKER_FLAGS var, is this a windows thing? it's not set by a vanilla FindOpenMP.cmake on linux.

akva2:
applied the same thing to the library target. now it is fine on linux, let's hope it doesn't hamper windows.

joakim-hove:
Ahhh trying to maintain a build system; the hair you have left certainly turns grey, I will try to take the opportunity to clean up things a bit - hopefully have an alternative ready tomorrow.

akva2:
more info about the openmp issue; https://gitlab.kitware.com/cmake/cmake/issues/9075 - been a known bug for 8 years.

PR #201, opened 2017-10-11
Fix ecl nnc data alloc wat flux

stefoss23:
**Task**
ecl_nnc_data_alloc_wat_flux shall return NULL if any wat_flux nnc type kws are missing when loaded from file. 

**Approach**
Changes in ecl_nnc_data.c causing ecl_nnc_data_alloc_XXX (including wat_flux) to return NULL if keywords necessary to build a complete ecl_nnc_data structure are missing. 
Also the data attribute of the ecl_nnc_data structure is set to be the same as for the ecl_nnc_geometry_type input. 

**Pre un-WIP checklist**

### DIFF
--- a/lib/ecl/ecl_nnc_data.c
+++ b/lib/ecl/ecl_nnc_data.c
@@ -173,7 +173,7 @@ static void assert_correct_kw_count(ecl_kw_type * kw, const char * function_name
 }
 
 
-static void ecl_nnc_data_set_values(ecl_nnc_data_type * data, const ecl_grid_type * grid, const ecl_nnc_geometry_type * nnc_geo, const ecl_file_view_type * init_file, int kw_type) {
+static bool ecl_nnc_data_set_values(ecl_nnc_data_type * data, const ecl_grid_type * grid, const ecl_nnc_geometry_type * nnc_geo, const ecl_file_view_type * init_file, int kw_type) {
 
    int current_grid1 = -1;
    int current_grid2 = -1;
@@ -182,6 +182,7 @@ static void ecl_nnc_data_set_values(ecl_nnc_data_type * data, const ecl_grid_typ
    int kw_count = 0;
    bool check_kw_count = false;
    int nnc_size = ecl_nnc_geometry_size( nnc_geo );   
+   bool all_nnc_kw_present = true;
 
    for (int nnc_index = 0; nnc_index < nnc_size; nnc_index++) {
       const ecl_nnc_pair_type * pair = ecl_nnc_geometry_iget( nnc_geo, nnc_index );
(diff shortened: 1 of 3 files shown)


lib/ecl/ecl_nnc_data.c
```diff
@@ -197,36 +198,37 @@ static void ecl_nnc_data_set_values(ecl_nnc_data_type * data, const ecl_grid_typ
          if (current_kw) {            
             correct_kw_count = ecl_kw_get_size( current_kw );
             check_kw_count = true;
-            data->size = nnc_index + correct_kw_count;
          }
          else {
-            check_kw_count = false;
-            printf("Warning: failed to obtain kw from file in function %s. ", __func__);
-            printf("Grid1: %d, Grid2 %d\n", current_grid1, current_grid2);
+            all_nnc_kw_present = false;
```
joakim-hove: Instead of setting this `bool` variable to `false` the whole function should just return `false` immediately.

lib/ecl/ecl_nnc_data.c
```diff
@@ -182,6 +182,7 @@ static void ecl_nnc_data_set_values(ecl_nnc_data_type * data, const ecl_grid_typ
    int kw_count = 0;
    bool check_kw_count = false;
    int nnc_size = ecl_nnc_geometry_size( nnc_geo );   
+   bool all_nnc_kw_present = true;
```
joakim-hove: The `all_nnc_kw_present` variable can be removed completely.

lib/ecl/ecl_nnc_data.c
```diff
@@ -197,36 +198,37 @@ static void ecl_nnc_data_set_values(ecl_nnc_data_type * data, const ecl_grid_typ
...
-         kw_count++;
+        data->values[nnc_index] = ecl_kw_iget_as_double(current_kw, pair->input_index);
+        kw_count++;
       }
-      else
-         data->values[nnc_index] = -1;
+      
    }
    assert_correct_kw_count(current_kw, __func__, check_kw_count, correct_kw_count, kw_count);
+   return all_nnc_kw_present;
```
joakim-hove: If you have come all the way down here it *is* `return true;`

lib/ecl/ecl_nnc_data.c
```diff
@@ -197,36 +198,37 @@ static void ecl_nnc_data_set_values(ecl_nnc_data_type * data, const ecl_grid_typ
...
-         data->values[nnc_index] = ecl_kw_iget_as_double(current_kw, pair->input_index);
-         kw_count++;
+        data->values[nnc_index] = ecl_kw_iget_as_double(current_kw, pair->input_index);
+        kw_count++;
       }
-      else
-         data->values[nnc_index] = -1;
+      
    }
    assert_correct_kw_count(current_kw, __func__, check_kw_count, correct_kw_count, kw_count);
```
joakim-hove: The `bool check_kw_count` should be removed from `assert_correct_kw_count( )` function - if you call the `assert_xxx` the check should be applied!

PR #202, opened 2017-10-11
Reimplement nexus2ecl in C++

JensGM:


joakim-hove:
I appreciate the C++ implementaction of the Nexus plotting as a kick in the butt towords a C++ implementarien of libecl, but I am very concierge that the new code integrates well with the existing code. For this particular PR I would like you to use (or modify) the existing endian flip code in util.c

pgdr:
> For this particular PR I would like you to use (or modify) the existing endian flip code in util.c

I'm sorry, but I disagree here.  Why would we use a (potentially buggy) endian flip implementation in util when we can use standard tools?  On the contrary, I think we rather could go through the call sites for endian_flip and use POSIX tools, and then trim down the size of `util.c`.

But for now, I think that Jens' approach sounds like a more sustainable solution.

PR #210, opened 2017-10-17
Added some small Python examples.

joakim-hove:
**Task**
Some examples for OPM + Python presentation.


pgdr:
I won't interfere with the presentation, but I would appreciate the `__name__ == '__main__'` idiom in all of these examples.

PR #212, opened 2017-10-17
POC: Cell from Grid

pgdr:
It is somewhat cumbersome to get the volume, coordinates and indices of a bunch of cells.

This adds a small Python class `Cell` that you can obtain from a grid by either using `grid[123]`, or from `grid.cell(i=3, j=4, k=5)`.

A `Cell` can give you volume, dz, coordinates, corners, indices etc.

This is a _proof of concept_ so please be constructive and think carefully about what a good API for a user would look like.

One example was updated to see some uses:
```python
vols = [c.volume for c in grid if c.active]

upper = grid[123].corners[:4] # upper or lower
# ...
        c = grid.cell(active_index=ai)
        print('dz(%2d, %2d, %2d) = %.3f' % (c.i, c.j, c.k, c.dz))

# supports __bool__ and __eq__:
vols = [c.volume for c in grid if c]  # c is same as c.active
c = grid[123]
d = grid[1,2,3]
if c == d:
    print('Yes, c and d are the same')
if c:
    print('c is active')
else:
    print('c is inactive')

xyz = (601292.32, 1412.231, 3242.3134)
if xyz in c:
    print('c contains world coordinate given')
```

### DIFF
--- a/python/python/ecl/ecl/ecl_grid.py
+++ b/python/python/ecl/ecl/ecl_grid.py
@@ -34,7 +34,7 @@ import itertools
 from cwrap import CFILE, BaseCClass
 from ecl.util import monkey_the_camel
 from ecl.util import IntVector
-from ecl.ecl import EclPrototype, EclDataType, EclKW, FortIO, EclUnitTypeEnum
+from ecl.ecl import EclPrototype, EclDataType, EclKW, FortIO, EclUnitTypeEnum, Cell
 
 
 class EclGrid(BaseCClass):
@@ -385,6 +385,33 @@ class EclGrid(BaseCClass):
         """
         n = self._get_name()
         return str(n) if n else ''
+
+    def cell(self, global_index=None, active_index=None, i=None, j=None, k=None):
+        if global_index is not None:
+            return Cell(self, global_index)
+        if active_index is not None:
+            return Cell(self, self.global_index(active_index=active_index))
+        if i is not None:
+            return Cell(self, self.global_index(ijk=(i,j,k)))
+
+    def __getitem__(self, global_index):
+        if isinstance(global_index, tuple):
+            i,j,k = global_index
+            return self.cell(i=i, j=j, k=k)
+        return self.cell(global_index=global_index)
+
+    def __iter__(self):
+        for i in range(len(self)):
+            yield self[i]
+
+    def cells(self, active=False):
+        """Iterator over all the (active) cells"""
+        if not active:
+            for c in self:
+                yield c
+        else:
+            for i in range(self.get_num_active()):
+                yield self.cell(active_index=i)
 
     def global_index(self, active_index=None, ijk=None):
         """
--- a/docs/examples/grid_info.py
+++ b/docs/examples/grid_info.py
@@ -3,32 +3,29 @@ import sys
 from ecl.ecl import EclGrid, EclRegion
 
 
-def volum_min_max(grid):
-    vmin = 10000000
-    vmax = 0
-    for a in range(grid.getNumActive()):
-        v = grid.cell_volume( active_index = a )
-        vmin = min(vmin,v)
-        vmax = max(vmax,v)
+def volume_min_max(grid):
+    vols = [c.volume for c in grid if c.active]
+    return min(vols), max(vols)
 
-    return vmin,vmax
+def main(grid):
+    vmin,vmax = volume_min_max(grid)
 
-
-if __name__ == "__main__":
-    case = sys.argv[1]
-    grid = EclGrid( case )
-
-    vmin,vmax = volum_min_max( grid )
-
-    dz_limit = 0.1
-    region = EclRegion( grid, False )
-    region.select_thin( dz_limit )
+    dz_limit = 0.3
+    region = EclRegion(grid, False)
+    region.select_thin(dz_limit)
 
     print "Smallest cell     : %g" % vmin
     print "Largest cell      : %g" % vmax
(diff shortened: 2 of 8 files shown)


python/python/ecl/ecl/ecl_grid.py
```diff
@@ -386,6 +386,21 @@ def get_name(self):
...
 
+    def cell(self, global_index=None, active_index=None, i=None, j=None, k=None):
+        if global_index is not None:
+            return Cell(self, global_index)
+        if active_index is not None:
+            return Cell(self, self.global_index(active_index=active_index))
+        if i is not None:
+            return Cell(self, self.global_index(ijk=(i,j,k)))
+
+    def __getitem__(self, global_index):
```
joakim-hove: This could check if the index argument was a tuple, i.e. ijk and then calculate global index from that - that way we could also support: `cell = grid[1,2,3]` - which looks kind of nice.

joakim-hove:
I think the basic idea - hooking a small `Cell`object onto `EclGrid.__getitem__` is *very good*.

joakim-hove:
The `Cell` class could possibly be implemented by wrapping the `ecl_cell_struct` which is already present in the C code.

pgdr:
I will add support for `__getitem__[1,2,3]`!

When it comes to using `ecl_cell_struct` as a backend for `Cell` I am not completely sure of the aim.  My goal was simply to expose a POD object representing a cell to the API.  If we notice that this implementation is too slow, it is trivial to update it to use the C backend later.

One thing I want to see whether is possible is to support _active indexing_ in the grid in the form of a context manager;  e.g.:

```python
active_volume = 0
with grid.active as a_grid:
    # a_grid is now a grid where every index is using active size and active index
    num = len(a_grid)  # this is num_active
    for cell in a_grid:
        active_volume += cell.volume  # cell is in this context guaranteed active
    a_grid[14].volume  # a_grid[14] is the 14th active cell
```

joakim-hove:
> When it comes to using ecl_cell_struct as a backend for Cell I am not completely sure of the aim

No big point - I just get itches when things are implemented more than time.

joakim-hove:
> One thing I want to see whether is possible is to support active indexing in the grid in the form of a context manager; e.g.:

Interesting idea; the point that both active indexing and global indexing make perfect sense - and that they are both just plain integers comes a bit in the way for a natural & nice api.

Instead of the `__iter__(self)` I have thought of either:
```Python
def cells(self, active=False):
     """ Returns an iterator over all cells; only include active cells is @active is True """ 
``` 
or alternatively two properties:

```Python
@property
def active_cells(self):
     ...

def global_cells(self)
    ...
```
where the latter might be named `all_cells` or just `cells`?

pgdr:
> I just get itches when things are implemented more than time.

I wholeheartedly agree, but I would wager that this is also in the nature of ctypes/cwrap.  We can choose to have the "backend" of `Cell` being `EclGrid` or to be "`cwrap(ecl_cell_struct)`".

pgdr:
Please comment on whether you think implementation of `__bool__/__nonzero__` and `__contains__` is over the top, or if it is intuitive.

Is it obvious that the following sums up active cell's volume?
```python
vol = 0
if c:
    vol += c.volume
return vol
```

Is it obvious that the following finds a cell containing a certain coordinate?
```python
for c in grid:
    if (x,z,y) in c:
        return c
return None
```

joakim-hove:
> Please comment on whether you think implementation of __bool__/__nonzero__ 

Binding `cell._active` to truth testing certainly makes sense.

> and __contains__ [...]

A bit more uncertain on this; I guess the reason I am a bit uncertain is that `__contains__` usually(?) takes an integer-like argument - but by all means `__contains__` certainly matches the intended check - in the traditional meaning of the word *contains* - I would say go for it.

pgdr:
Well, in a `list`  `__contains__` checks for membership, in a `string` `__contains__` checks substring and in a `dict`, `__contains__` checks if the given argument is a key.

But I agree that in all these cases, the element provided is something that is in a sense enumerable in the type.  For a `Cell`, that would mean that we consider a `Cell` defined as all the Cartesian points it contains.  I don't know if that is sensible or not?

joakim-hove:
I would say go for the __contains__

markusdregi:
> For a Cell, that would mean that we consider a Cell defined as all the Cartesian points it contains. I don't know if that is sensible or not?

This is exactly how I at least would define a cell formally. Hence, I'm also in favour of contains.

This is a very nice PR by the way, I've numerous times ended up with the `[grid.getCellCorner(i, k) for k in range(8)]` construction (Was the indices the other way around? Which again proves the point of this PR..).

I have one additional comment and that is that I personally would expect the `__bool__` to also be hooked up to the `valid_cell_geometry` or similar. These are cells that are odd creatures and are treated accordingly by the grid. Volumes nor containment is well-behaved for these ones.

python/python/ecl/ecl/ecl_grid.py
pgdr: Done

pgdr:
> I personally would expect the `__bool__` to also be hooked up to the valid_cell_geometry or similar. These are cells that are odd creatures and are treated accordingly by the grid.

Okay, I mapped up `valid` and `valid_geometry`, but your comment makes me think that it's not obvious what `__bool__` is supposed to mean.

I will therefore remove the `__bool__` functionality until we all agree that there is _one_ intuitive meaning.

pgdr:
This is now ready for reviewed.  But please keep API in mind!

PR #213, opened 2017-10-19
Fixed bug in ecl_kw.assign

andreabrambilla:
**Task**
If an empty mask is passed to assign, no cell of the grid should be affected

**Approach**
Explicitly check if the the mask is None instead of relying on the implicit conversion to bool.

From a more philosophical point of view, one might argue that an empty region is not really an "invalid" (False) region. So, an alternative solution could be to just get rid of the `__nonzero__` method in ecl_region.py

**Pre un-WIP checklist**

### DIFF
--- a/python/python/ecl/ecl/ecl_kw.py
+++ b/python/python/ecl/ecl/ecl_kw.py
@@ -719,7 +719,7 @@ class EclKW(BaseCClass):
         """
         if self.isNumeric():
             if type(value) == type(self):
-                if mask:
+                if mask is not None:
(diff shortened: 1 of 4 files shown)


joakim-hove:
Thank you - this should be merged; just one very minor detail - I guess it is a matter of preference, but I would personally like to flip the testing - i.e:
```python
if mask is None:
    # Assign to all cells
else:
    # Do masked assign
```

andreabrambilla:
I agree, but find/replace was much easier ;)

PR #214, opened 2017-10-19
WIP: Read nexus data

JensGM:


lib/include/ert/nexus/nexus_plot.hpp
```diff
@@ -41,17 +42,38 @@ struct unexpected_eof : public std::runtime_error {
...
     int32_t num_classes = 0;
     int32_t day, month, year;
     int32_t nx, ny, nz;
     int32_t ncomp;
-    std::vector<std::string> class_names;
-    std::vector<int32_t> vars_in_class;
-    std::vector< std::vector<std::string> > var_names;
+    std::vector< std::string > class_names;
+    std::map< std::string, int32_t > vars_in_class;
+    std::map< std::string, std::vector<std::string> > var_names;
```
andreabrambilla: unless there is some weirdness hidden somewhere, I'd expect `class_names == va_names.keys()` (no, there is no keys() method, but that's the idea) and also `vars_in_class["class"] = var_names["class"].size()`. So it is sufficient to store the `var_names` map, and you don't have to bother with keeping these 3 variables consistent with each other

lib/include/ert/nexus/nexus_plot.hpp
```diff
@@ -41,17 +42,38 @@ struct unexpected_eof : public std::runtime_error {
...
 
+private:
+    std::map< int, timestep_data > plot_data;
```
andreabrambilla: I would provide both the const and non-const version of this method. That would allow you to create const refs to an instance of this class to pass around to functions

lib/include/ert/nexus/nexus_plot.hpp
```diff
@@ -41,17 +42,38 @@ struct unexpected_eof : public std::runtime_error {
...
+
+public:
+    std::map< int, timestep_data >& data();
```
andreabrambilla: I would provide both the const and non-const version of this method. That would allow you to create const refs to an instance of this class to pass around to functions

lib/nexus/nexus_plot.cpp
```diff
@@ -39,6 +41,22 @@ const std::string NEXUS_PLOT_TYPE_HEADER = "PLOT  BIN ";
...
+    int32_t frac_bits = i & 0x007FFFFFL;
+    double c = (exp_bits ? 1.0 : 0.0) + frac_bits / double(0x007FFFFFL);
+
+    if (exp_bits == 0x000000FFL)
+        return frac_bits ? NAN : (sign_bit ? -INFINITY : INFINITY);
+    else if (frac_bits | exp_bits)
+        return float(sign * c * std::pow(2, exp));
+    else
+        return 0.0f;
+}
```
andreabrambilla: maybe I'm missing something here, but wouldn't it be sufficient to `return static_cast<float>(i)`

lib/nexus/nexus_plot.cpp
```diff
@@ -85,30 +101,87 @@ hinfo headerinfo( std::istream& stream ) {
...
+        //Skipped
+        stream.seekg(8, std::ios::cur);
+
+        std::array< int32_t, 5 > buf {};
+        stream.read( (char*)buf.data(), buf.max_size() * 4 );
+        if ( !stream.good() ) throw nex::unexpected_eof("");
+        std::transform( buf.begin(), buf.end(), buf.begin(), ntohl );
+
+        int timestep  = (int) int32_t_bits_to_float(buf[0]);
+        int num_items = (int) int32_t_bits_to_float(buf[2]);
```
andreabrambilla: not really an issue, but I think the C-style type casting is kind of frowned upon...

lib/nexus/nexus_plot.cpp
```diff
@@ -85,30 +101,87 @@ hinfo headerinfo( std::istream& stream ) {
...
+        if (data.find(timestep) == data.end())
+            data[timestep] = nex::NexusPlot::timestep_data {};
+        if (data[timestep].find(class_name) == data[timestep].end())
+            data[timestep][class_name] =
+                std::vector<nex::NexusPlot::NexusClassItem> {};
```
andreabrambilla: Not really a performance killer, but every time you write ` data[timestep]`, you have a lookup through the map. 
An alternative could be
```C++
auto& tsData = data[timestep];  // if it doesn't exist yet, it's default constructed
auto& classData = tsData[class_name]; // same

File: lib/nexus/nexus_plot.cpp [hunk 7]
```diff
@@ -122,22 +195,24 @@ nex::NexusPlot::NexusPlot( std::istream& stream ) {
...
+
+        ~stream_guard() { this->s.exceptions(this->mask); }
+
         std::ios::iostate mask;
-        std::istream& s;
-    } g { stream };
-    stream.exceptions( std::ios::goodbit );
+        std::istream &s;
+    } g{stream};
+    stream.exceptions(std::ios::goodbit);
```
andreabrambilla: just a curiosity: I looked up the documentation for `exceptions()` but it's fairly fuzzy. What does that function do??

lib/nexus/tests/nexus_plot_load.cpp
```diff
@@ -125,18 +126,59 @@ void test_spe1_header(char *argv) {
...
+
+    for (auto it : plt.data()) {
+        std::cout <<  "time(" << it.first << ") -> {" << std::endl;
+        for (auto it1 : it.second) {
+            std::cout << "  " << it1.first << ":" << it1.second.size();
         }
+        std::cout << std::endl << "}" << std::endl;
     }
+
+    std::cout << plt.data().at(33)["CONN    "][0]["CGAS"] << std::endl;
```
andreabrambilla: I still think that `plt.data() @33 @"CONN    " @0 @"CGAS"` would be amazing ;)

PR #215, opened 2017-10-21
Exposing `grid.region`

pgdr:
A user shouldn't have to know about `EclRegion` to use it.

A user should be able to write `region = grid.region()` or `grid.region(preselect=True)`, and then go on with her life.

Now you can write
```python
from ecl.ecl import EclGrid
grid = EclGrid(fname)
reg = grid.region()
reg.select_all()
print(len(reg))
```

Next up:

```python
import ecl
grid = ecl.parse_grid(fname)
# ...
```

### DIFF
--- a/docs/examples/avg_pressure.py
+++ b/docs/examples/avg_pressure.py
@@ -3,81 +3,84 @@ import sys
 import matplotlib.pyplot as plt
 
 # Import the required symbols from the ecl.ecl package.
-from ecl.ecl import EclFile, EclGrid, EclRegion, EclRestartFile
+from ecl.ecl import EclFile, EclGrid, EclRestartFile
 
 
-# Calculate the average pressure for all the cells in the region using
-# three different methods.
-def avg_pressure(p, sw, pv, region, region_id, result):
-    if region:
-        p_pv = p*pv
-        p_hc_pv = p*pv*(1 - sw)
-        hc_pv = pv*(1 - sw)
+def compute_avg_pressure(p, sw, pv, region, region_id, result):
+    """Calculate the average pressure for all the cells in the region using three
+     different methods.
 
-        total_pv = pv.sum( mask = region )
-        total_hc_pv = hc_pv.sum( mask = region )
+    """
+    if not region:
+        return None, None, None
+    p_pv = p*pv
+    p_hc_pv = p*pv*(1 - sw)
+    hc_pv = pv*(1 - sw)
 
-        p1 = p.sum( mask = region) / region.active_size( )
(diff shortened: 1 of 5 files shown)


joakim-hove:
I agree that `reg = grid.region()` is better than the current apporach, and would like to merge it in some way. But there are many users of the current approach out in the wild - and that must continue to work - although it probably should be deprecated. Currently there seems to be some circular import problems  between `ecl_region.py`and `ecl_grid.py`.

A bit unfortunate with all the formatting changes which are part of this PR - that makes it difficult to see the changes, specially in `ecl_region.py`.

joakim-hove:
Maybe a bit premature, but this:

```python
import ecl
grid = ecl.parse_grid(fname)
```

I do not like. The user does *not* want to parse - she wants a grid; the fact some parsing is involved is not really relevant to the user.

pgdr:
> A bit unfortunate with all the formatting changes which are part of this PR - that makes it difficult to see the changes, specially in ecl_region.py.

Actually, I believe there are no changes in `ecl_region` except a fix of the `hash` function.  But I can make it clearer if you want to.

My point about parse (or load or read or grid, call it what you will), is that for a user, having to write
```python
import ecl
grid = ecl.ecl.EclGrid(fname)         # or move ecl.ecl up into import area
reg = ecl.ecl.EclRegion(grid, False)
```
is not the way it should be done.

In numpy, you would write `numpy.array(data)` and in pandas you would write `pandas.read_csv(fname)` or similarly.

The Pythonic way is not for a user to get the data they want by instantiating a class, we should expose a function, e.g.
```python
import ecl
grid = ecl.grid(fname)
reg = grid.region()  # reg could even be a numpy array
```

joakim-hove:
>The Pythonic way is not for a user to get the data they want by instantiating a class, we should expose a function, e.g.

We agree on the region thing - but changing: 
```python
from ecl.ecl import EclGrid
grid = EclGrid( fname )
```
to
```
import ecl.ecl
grid = ecl.ecl.load_grid( fname )
```
is in my opinion close to making changes just to make changes. Due to the required backwards compatibility there is certainly a code-complexity price related to this.

pgdr:
You are right, of course we're not going to have the users write
```python
import ecl.ecl
grid = ecl.ecl.load_grid(fname)
```

> Due to the required backwards compatibility

There is no required backward compatibility once we start our journey to ecl 3 :)

joakim-hove:
I liked this - why was it closed?

joakim-hove:
Is there a reason this was abondoned?

PR #216, opened 2017-10-22
WIP: Can ignore errors when opening ecl_file.

joakim-hove:
**Task**

**Approach**

**Pre un-WIP checklist**

### DIFF
--- a/lib/ecl/ecl_file.c
+++ b/lib/ecl/ecl_file.c
@@ -138,6 +138,7 @@ struct ecl_file_struct {
   int             flags;
   vector_type   * map_stack;
   inv_map_type  * inv_view;
+  bool            input_error;
 };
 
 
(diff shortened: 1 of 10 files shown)


lib/ecl/ecl_file.c
```diff
@@ -542,10 +542,8 @@ static bool ecl_file_scan( ecl_file_type * ecl_file ) {
 
     ecl_kw_free( work_kw );
   }
-  if (scan_ok)
-    ecl_file_view_make_index( ecl_file->global_view );
-
-  return scan_ok;
+  ecl_file_view_make_index( ecl_file->global_view );
+  return (scan_ok || ignore_errors);
```
andreabrambilla: It's maybe unnecessary that this method handles the `ignore_errors` flag. Whoever is calling this method is aware of `ignore_errors` so it can do
```
void whatever( ..., bool ignore_errors) {
  ....
   bool result = ecl_file_scan( .... )
   if(result || ignore_errors)
      ...
}
```

joakim-hove: Yes - that is a good idea.

andreabrambilla:
Looks good to me. In general, I think it's acceptable to allow users to ignore errors, but those errors should be anyway reported. The user should be aware that a problem has occurred and that there is no guarantee that the loaded data is correct

joakim-hove:
> but those errors should be anyway reported. The user should be aware that a problem has occurred and that there is no guarantee that the loaded data is correct

I agree - getting this really Right&trade; is tricky - now it will just print a warning on `stderr`.

PR #217, opened 2017-10-23
WIP: Fix RFT bug

stefoss23:
**Task**
ERT-1412: When using the command ecl_rft_file_alloc_case() with parameter : filename = /d/proj/bg/enkf/ErtTestData/ECLIPSE/RFT/NORNE_ATW2013_RFTPLT_V2.RFT, ERT will crash hard. 

**Approach**
 ecl_rft_file_alloc_case crashed is set to crash if for a certain blockview, the blockview:
1) Does not contain a conpres_kw
2) Has a conpres_kw, but all values are 0, and there is no PRESSURE_kw. 

The case 2 applies for the filename mentioned under "Task".  For the case 2, the program will not crash hard, but instead issue a warning. 

**Pre un-WIP checklist**


joakim-hove:
When the Statoil tests have been run this should be merged.

PR #224, opened 2017-10-27
Read nexus data

JensGM:
Read data from nexus plot files. The data is stored in a sequential manner.

### DIFF
--- a/lib/include/ert/nexus/nexus_plot.hpp
+++ b/lib/include/ert/nexus/nexus_plot.hpp
@@ -19,13 +19,18 @@
 #ifndef NEXUS_PLOT_H
 #define NEXUS_PLOT_H
 
+#include <algorithm>
+#include <array>
 #include <cinttypes>
-#include <stdexcept>
 #include <fstream>
+#include <map>
+#include <memory>
+#include <stdexcept>
 #include <string>
 #include <vector>
 
-typedef struct ecl_sum_struct ecl_sum_type;
+#include <ert/ecl/ecl_sum.h>
+
 
 namespace nex {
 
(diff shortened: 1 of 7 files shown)


lib/include/ert/nexus/nexus_plot.hpp
```diff
@@ -39,25 +44,270 @@ struct unexpected_eof : public std::runtime_error {
...
+               this->max_perfs     == rhs.max_perfs &&
+               this->classname     == rhs.classname &&
+               this->instancename == rhs.instancename &&
+               this->varname       == rhs.varname &&
+               this->value         == rhs.value;
+    }
+    constexpr bool operator!=(const NexusData& rhs) const noexcept {
+        return ! (*this == rhs);
+    }
+    // bool operator< (const NexusData& rhs) { }
```
joakim-hove: Just remove commented out code.

lib/include/ert/nexus/nexus_plot.hpp
```diff
@@ -39,25 +44,270 @@ struct unexpected_eof : public std::runtime_error {
...
+    // bool operator<=(const NexusData& rhs) { }
+    // bool operator> (const NexusData& rhs) { }
+    // bool operator>=(const NexusData& rhs) { }
 };
 
+namespace get {
+
+    std::array< char, 8 > classname( const NexusData& nd ) {
```
joakim-hove: Do we really need all these methods?

lib/include/ert/nexus/nexus_plot.hpp
```diff
@@ -39,25 +44,270 @@ struct unexpected_eof : public std::runtime_error {
...
-    std::vector< std::vector<std::string> > var_names;
+};
+
 
-    ecl_sum_type* ecl_summary( const std::string& ecl_case );
+struct NexusData {
+    int32_t timestep;
+    float time;
+    int32_t max_perfs;
+    std::array< char, 8 > classname;
```
joakim-hove: Why not just use `std::string` - if this is a performance optimization - *it is not worth it*.

lib/include/ert/nexus/nexus_plot.hpp
```diff
@@ -39,25 +44,270 @@ struct unexpected_eof : public std::runtime_error {
...
+    // bool operator> (const NexusData& rhs) { }
+    // bool operator>=(const NexusData& rhs) { }
 };
 
+namespace get {
+
+    std::array< char, 8 > classname( const NexusData& nd ) {
+        return nd.classname;
+    }
+    std::array< char, 8 > instancename( const NexusData& nd ) {
```
joakim-hove: The `NexusData` is already a struct with all members public; I think the attributes should just be accessed through `.xxx` notatiation.

JensGM: These are used for filtering and transforming iterables, they are just convenience functions. I use them several places.

lib/include/ert/nexus/nexus_plot.hpp
```diff
@@ -39,25 +44,270 @@ struct unexpected_eof : public std::runtime_error {
...
+        return nd.max_perfs;
+    }
+    float value( const NexusData& nd ) {
+        return nd.value;
+    }
+}
+
+namespace is {
+
+    struct timestep {
```
joakim-hove: Lo and behold - this was much code for not so much functionality?

lib/include/ert/nexus/nexus_plot.hpp
JensGM: It is both for performance reasons and the fact that the we know that all classnames and instancenames are 8 characters long. This makes construction simple, and comparisons can be done without having to worry about null-terminators. Strings can, but does not need to, contain null-terminators. When comparing two equal strings, one with null-terminator and one without, c++ will say they are different. It is also easier to read into an std::array as std::array::data is guaranteed to return a sequential pointer, std::string is not (until c++14).

lib/include/ert/nexus/nexus_plot.hpp
JensGM: Same as the reply below, convenience.

lib/include/ert/nexus/nexus_plot.hpp
JensGM: Fixed

lib/include/ert/nexus/nexus_plot.hpp
JensGM: I agree it seems a bit much, but it does make the rest of the code much less messy.

lib/include/ert/nexus/nexus_plot.hpp
JensGM: They let you write things like 
```std::copy_if( field.begin(), field.end(), std::back_inserter( cop ), is::varname( "COP" ));```
rather than
```std::copy_if( field.begin(), field.end(), std::back_inserter( cop ), [](const auto& d) { return d.varname == "COP "; });```

In my opinion, this makes the intent of the line of code easier to understand.

Comment (joakim-hove) [reply lib/include/ert/nexus/nexus_plot.hpp]: OK - I agree.

File: lib/include/ert/nexus/nexus_plot.hpp [hunk 6]
```diff
@@ -39,25 +44,266 @@ struct unexpected_eof : public std::runtime_error {
...
+        return lhs.varname == rhs.varname;
+    }
+    bool value( const NexusData& lhs, const NexusData& rhs ) {
+        return lhs.value == rhs.value;
+    }
+}
+
+struct NexusPlot {
+    NexusHeader header;
+    std::vector< NexusData > data;
```
joakim-hove: When the `NexusData` is implemented as a `struct` with all public members I think the container should use `const NexusData`

lib/nexus/nexus_plot.cpp
```diff
@@ -32,128 +35,203 @@
...
 
-const std::string NEXUS_PLOT_TYPE_HEADER = "PLOT  BIN ";
 
 namespace {
 
+uint32_t ntoh(uint32_t ui) { return ntohl(ui); }
+
+float interpret_float(int32_t i) {
+    float f;
+    std::memcpy( &f, &i, sizeof( float ) );
```
joakim-hove: `sizeof f`

lib/include/ert/nexus/nexus_plot.hpp
```diff
@@ -39,25 +44,266 @@ struct unexpected_eof : public std::runtime_error {
...
+        auto last = std::unique(vec.begin(), vec.end());
+        std::sort(vec.begin(), last);
+        last = std::unique(vec.begin(), last);
+        vec.erase(last, vec.end());
+        return vec;
+    }
+};
+
+NexusPlot load( const std::string& );
+NexusPlot load( std::istream& );
```
joakim-hove: I do not see the `istream` overload as being relevant - please remove it.

lib/nexus/tests/nexus_plot_spe1.cpp
```diff
@@ -0,0 +1,223 @@
...
+
+#include <ert/util/test_util.hpp>
+#include <ert/nexus/nexus_plot.hpp>
+
+using namespace nex;
+
+
+namespace {
+    const std::vector< std::string > spe1_classnames {{
+        "WELL    ", "NODE    ", "CONN    ", "REGION  ", "FIELD   ", "CONNLIST"
```
joakim-hove: These spaces are on the way out in a subsequent PR?

lib/include/ert/nexus/nexus_plot.hpp
JensGM: The istream overload is used for tests. I could be removed but it makes testing the exceptions easier. The alternative is to just create additional test data that hold the same information as the streams.

lib/nexus/tests/nexus_plot_spe1.cpp
JensGM: I'll fix this before the merge.

lib/include/ert/nexus/nexus_plot.hpp
JensGM: I don't think this is actually possible, but by constructing the the plot as `const plt = nex::load(...);`, you essentially achieve the same thing (which is how it's used in all the current code).

PR #227, opened 2017-10-30
Write ECL summary for FIELD from nexus data

JensGM:


joakim-hove:
> No description provided.

Please fill out a bit; this PR seems to start on top of: #224 - that is at least something which should be written.

lib/nexus/nexus_plot.cpp
```diff
@@ -230,10 +231,17 @@ NexusPlot nex::load(std::istream& stream) {
 }
 
 ecl_sum_type* nex::ecl_summary(const std::string& ecl_case, const NexusPlot& plt) {
+    std::time_t base_time{};
+    std::time( &base_time );
+    std::tm *tm = std::localtime ( &base_time );
+    tm->tm_mday = plt.header.day;
+    tm->tm_mon  = plt.header.month - 1;
+    tm->tm_year = plt.header.year - 1900;
```
joakim-hove: Please use the util function:
```
std::time_t sime_start = util_make_date_utc( plt.header.day, plt.header.month , plt.header.year);
```

lib/nexus/nexus_plot.cpp
```diff
@@ -255,20 +263,56 @@ ecl_sum_type* nex::ecl_summary(const std::string& ecl_case, const NexusPlot& plt
...
+        { "QGI ", "FGIR" },
+        { "CWI ", "FWIT" },
+        { "CGI ", "FGIT" },
+        { "QPR ", "FCPR" },
+        { "CPP ", "FCPC" },
+        { "CTP ", "FTPR" },
+        { "TC  ", "FTPC" }
+    };
+
+    auto field_class_vars = plt.varnames("FIELD   ");
```
joakim-hove: Kill the space padding.

lib/nexus/nexus_plot.cpp
```diff
@@ -255,20 +263,56 @@ ecl_sum_type* nex::ecl_summary(const std::string& ecl_case, const NexusPlot& plt
...
+    std::vector< smspec_node_type* > smspec_nodes;
+    for (const auto& kw : field_class_vars) {
+        auto it = kw_nex2ecl.find(kw);
+        if (it != kw_nex2ecl.end()) {
+            field_vars.push_back(kw);
+            const auto& ecl_kw = it->second;
+            auto* node = ecl_sum_add_var( ecl_sum, ecl_kw.c_str(), NULL, -1,
+                                          "x", 0.0 );
+            smspec_nodes.push_back(node);
+        }
```
joakim-hove:
```C++
else
     std::cerr << "Do not know how to translate " << kw << " to eclipse" << std::endl;
```

Alternatively - if the `if (it != kw_nex2ecl.end())` should *never* fail - I suggest we rather use `at` - which signals exactly that.

lib/nexus/nexus_plot.cpp
```diff
@@ -255,20 +263,56 @@ ecl_sum_type* nex::ecl_summary(const std::string& ecl_case, const NexusPlot& plt
...
+    }
 
     auto nex_timesteps = plt.get_unique(get::timestep);
     auto nex_times = plt.get_unique(get::time);
     std::vector< ecl_sum_tstep_type* > timesteps;
     for (size_t i = 0; i < nex_timesteps.size(); i++) {
-        auto* ts = ecl_sum_add_tstep(ecl_sum, nex_timesteps[i], nex_times[i] );
+        float sim_time = i == 0 ? nex_times[0] : nex_times[i] - nex_times[i-1];
```
joakim-hove: This test looks wrong to me; I would assume the equivalent of:
```
if (i == 0)
   sim_time = 0;
else
   sim_time = nex_times[i] - nex_times[i-1];
```

lib/nexus/tests/nexus2ecl.cpp
```diff
@@ -44,6 +44,12 @@ void test_create_ecl_sum(char *root_folder) {
     test_assert_true( util_file_exists( "ECL_CASE.FSMSPEC"));
 
     test_work_area_free(work_area);
+
+
+    ecl_sum = nex::ecl_summary( "/tmp/necl/ECL_CASE", plt );
```
joakim-hove: This test was quite basic:
1. Please use the `test_area_context` instead of writing to a hardcoded path.
2. Should try to load - and at least do a very basic - something like:
```C
{
   ecl_sum_type * ecl_sum_loaded = ecl_sum_fread_alloc_case("ECL_CASE", ":");

   test_assert_true( ecl_sum_has_key(ecl_sum_loaded, "FOPR"));
   test_assert_time_t_equal( ecl_sum_get_start_time( ecl_sum_loaded ), util_make_data_utc( day, month ,year));
   test_assert_int_equal( ecl_sum_get_data_length( ecl_sum_loaded ), timesteps.size());
   
   for (size_t tstep = 0; tstep < ecl_sum_get_data_ength( ecl_sum_loaded ); tstep++)
        test_assert_double_equal( ecl_sum_get_general_var( ecl_sum_loaded, tstep , "FOPR"), fopr_values[tstep].value);

   ecl_sum_free( ecl_sum_loaded);
}
```

lib/nexus/nexus_plot.cpp
JensGM: This can fail without being wrong. There are some of variables in nexus not supported by eclipse. These could however be specified in eclupse by using user-defined quantities, for example `FUWOR` (Field, user-defined, water, oil, ratio).

joakim-hove: I see - then; I suggest we stick to a `std::cerr` warning if the test fails.

lib/nexus/tests/nexus2ecl.cpp
JensGM: Removed the temporary test that wrote a filed to /tmp/necl/ and added the test you wrote for point 2.

JensGM:
@joakim-hove All the points you brought up should have been addressed now. There is a major revision of the file structure in a new commit https://github.com/Statoil/libecl/pull/227/commits/0ee3567147322490be9c0ed600f3b2f44840f322. The reason for this change is that the header files should not be installed as they are not ready to be exposed as a part of the ecl library.

All the convenience functions from the PR https://github.com/Statoil/libecl/pull/224 are now in `<nexus/util.hpp>` and the part of the API that could be exposed later are in `<nexus/nexus_plot.hpp>`. This choice of file structure is inspired by segyio. These headers are available for libecl only, and are not installed with `make install`.

joakim-hove:
> These headers are available for libecl only, and are not installed with make install.

Very good - I fully support the spltting in header files only used during the build, and header files installed publicly.

PR #228, opened 2017-10-30
Make sure TestArea is destructed as soon as a `with` block ends

andreabrambilla:
Destructing the TestArea most likely include a change of directory.
If the object is not destructed explicitly, the garbage collector would
do it, but asynchronously. And changing directory at an undefined point
in time might affect other part of the code

**Pre un-WIP checklist**


joakim-hove:
Good - merge when Travis is Green.

PR #229, opened 2017-10-30
Add nex2ecl executable

ReedOnly:
**Task**
- Generate executable for converting Nexus summary to Eclipse summary.

How to use:    `./build/bin/nex2ecl /path/to/summary.plt`

**Pre un-WIP checklist**
- Make AppVeyor find ntohl for Windows compilation


joakim-hove:
> No description provided.

Please add description - that template initial text is there for a reason!

applications/CMakeLists.txt
```diff
@@ -10,6 +10,11 @@ if (BUILD_APPLICATIONS)
     target_link_libraries(grdecl_grid   ecl)
     target_link_libraries(summary       ecl)
 
+    if (BUILD_NEXUS)
+        add_executable(nex2ecl   nexus/nex2ecl.cpp)
+        target_link_libraries(nex2ecl   ecl)
+    endif ()
```
joakim-hove: Need to add the `nex2ecl` binary to the `apps` list to ensure it is installed:
```txt
list( APPEND apss nex2ecl )
```

ReedOnly: OK. Fixed now. But the executable was already in `build/bin` and working?

applications/nexus/nex2ecl.cpp
```diff
@@ -0,0 +1,47 @@
...
+    std::string filename( file_path );
+    auto ending = filename.rfind( ".plt" );
+    if (ending == std::string::npos) {
+        filename += ".plt";
+    }
+
+    nex::NexusPlot plt = nex::load( filename );
+
+    std::string basename( filename );
+    basename.erase(basename.find_last_of( '.' ), basename.size());
```
joakim-hove: For Eclipse output the basename should be `ALL_CAPS`. In addition it is the question of what to with a possible directory part - as I see it there are two possibilities:

1. The ECLIPSE output will be generated in cwd, irrespective of where the nexus case is located. In that case it is important to strip off the leading path component.
2. The ECLIPSE output will be generated in the same location as the nexus case is located, in that case it is important that *only* the base name is UPPERCASED - and not the directory part.

applications/CMakeLists.txt
joakim-hove: Yes - but when you later did `make install` it would not have installed in `$PREFIX/bin`

applications/nexus/nex2ecl.cpp
ReedOnly: Now option 2 is implemented, output file will be given the same name as input, but in upper case. It will be generated in the same folder as of now, I can add an option to output in spesified folder with syntax like: `nex2ecl input.plt -o /output/directory/` if desired.

applications/nexus/nex2ecl.cpp
```diff
@@ -0,0 +1,55 @@
...
+#include <string>
+
+#include <iostream>
+
+#include <ert/nexus/nexus_plot.hpp>
+
+
+void create_ecl_sum(char *file_path) {
+    std::string filename(file_path);
+    auto ending = filename.rfind(".plt");
```
joakim-hove: I would prefer if the path manipulations were done with the util functions - yes they are quite ugly plain C, but they work - even on Windows:
```C
char * basename;
util_alloc_file_components( file_path , NULL, &basename , NULL);
util_strupr( basename );

ecl_sum_type * ecl_sum = nex::ecl_summary( basename, plt);
ecl_sum_fwrite( ecl_sum );
ecl_sum_free( ecl_sum );
free( basename );
```

joakim-hove:
OK - please change the path manipulations to use the `util` variety, then rebase and merge!

applications/nexus/nex2ecl.cpp
ReedOnly: OK. Fixed now

PR #232, opened 2017-11-02
Added function util_alloc_normal_path()

joakim-hove:
**Task**
Added a function which will normalize a path like: "../../path/subpath/./../../". Convenient for testing in libres.

**Pre un-WIP checklist**

### DIFF
--- a/lib/util/tests/ert_util_normal_path.c
+++ b/lib/util/tests/ert_util_normal_path.c
@@ -1,0 +1,73 @@
+/*
+   Copyright (C) 2017  Statoil ASA, Norway.
+
+   The file 'ert_util_normal_path.c' is part of ERT - Ensemble based Reservoir Tool.
+
+   ERT is free software: you can redistribute it and/or modify
+   it under the terms of the GNU General Public License as published by
+   the Free Software Foundation, either version 3 of the License, or
+   (at your option) any later version.
+
+   ERT is distributed in the hope that it will be useful, but WITHOUT ANY
+   WARRANTY; without even the implied warranty of MERCHANTABILITY or
+   FITNESS FOR A PARTICULAR PURPOSE.
+
+   See the GNU General Public License at <http://www.gnu.org/licenses/gpl.html>
+   for more details.
+*/
+
+#include <stdlib.h>
+#include <stdbool.h>
+
+#include <ert/util/test_util.h>
+#include <ert/util/test_work_area.h>
+#include <ert/util/util.h>
+
+
+void test_path(const char * input_path, const char * expected_path) {
+  char * normal_path = util_alloc_normal_path( input_path );
+  test_assert_string_equal( normal_path , expected_path );
+  free( normal_path );
+}
+
+
+void test_relative() {
+  test_work_area_type * work_area = test_work_area_alloc("Work");
+  util_make_path("level0/level1/level2");
+
+  test_path( "level0/level1/../", "level0");
+  test_path( "level0", "level0");
+  test_path( "level0/././level1/../", "level0");
+  test_path( "level0/level1/../level1/level2/../", "level0/level1");
+  test_path( "level0/level1/../level1/level2/..", "level0/level1");
+  test_path( "level0/level1/../level1/level2/../file.txt", "level0/level1/file.txt");
+  test_path( "level0/level1/../level1a" , "level0/level1a");
+
+  test_path( "a/b/c/../c/../.." , "a");
+  test_path( "a/b/c/d/e/f" , "a/b/c/d/e/f");
+  util_chdir("level0/level1");
+  test_path("../../level0/level1/level2/../file.txt" , "file.txt");
+  test_path("../../level0/level1/level2/../" , "");
+  test_work_area_free( work_area );
+}
+
+void test_beyond_root() {
+  test_work_area_type * work_area = test_work_area_alloc("Work");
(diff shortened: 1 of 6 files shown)


pgdr:
> Convenient for testing in libres

Perhaps it should be in _libres_  then?

lib/util/util.c
```diff
@@ -895,6 +895,13 @@ char * util_alloc_realpath__(const char * input_path) {
         if (path_index == path_len)
           break;
 
+        /* We have backrefed us out of the current filesystem. */
+        if (prev_index < 0) {
+          free( abs_path );
+          free( real_path);
+          return NULL;
```
jokva: Is this _really_ desired behaviour? That it returns `nullptr` when it hits "beyond root"?

pgdr: I agree that I would expect `../../../../../../../..adinfinitum` to return root since `/.. = /`.

joakim-hove: OK - I just checked this:
```
cd /tmp/path
cd ../../../../../../tmp/path
pwd 
/tmp/path
```
I must say I was a little surprised by that - but then yes I fully agree that the semantics should be changed; thank you.

lib/util/tests/ert_util_normal_path.c
```diff
@@ -0,0 +1,71 @@
...
+  test_work_area_type * work_area = test_work_area_alloc("Work");
+  util_make_path("level0/level1/level2");
+
+
+  test_path( "level0/level1/../", "level0");
+  test_path( "level0", "level0");
+  test_path( "level0/././level1/../", "level0");
+  test_path( "level0/level1/../level1/level2/../", "level0/level1");
+  test_path( "level0/level1/../level1/level2/..", "level0/level1");
+  test_path( "level0/level1/../level1/level2/../file.txt", "level0/level1/file.txt");
```
markusdregi: Please add a test that does not move linearly in the file system. Something like:
`test_path( "level0/level1/../level1a", "level0/level1a" );`
I expect it to pass, but I would like it to there.

lib/util/tests/ert_util_normal_path.c
```diff
@@ -0,0 +1,71 @@
...
+  free( normal_path );
+}
+
+
+void test_relative() {
+  test_work_area_type * work_area = test_work_area_alloc("Work");
+  util_make_path("level0/level1/level2");
+
+
+  test_path( "level0/level1/../", "level0");
```
markusdregi: Do we expect this to work at Windows as well? In that case I would appreciate a test with a Windows back slash path as well.

lib/util/tests/ert_util_normal_path.c
joakim-hove: New test added and force-pushed

lib/util/tests/ert_util_normal_path.c
joakim-hove: > Do we expect this to work at Windows as well?

Ehhh yes but no; we *expect* it to work on Windows, but we do not run the tests on Windows - so how naive is that?!

The way windows path support is implemented is that all operations working with paths use the symbol `UTIL_PATH_SEP_STRING` - and that resolves to `"/"` and `"\"` on Posix and Windows respectively. So - although the code is assumed to be portable to windows - and compile-tested for that - the tests are not written portably.

lib/util/util.c
```diff
@@ -889,26 +890,33 @@ char * util_alloc_realpath__(const char * input_path) {
...
           break;
 
         if (strcmp(path_list[path_index] , BACKREF ) == 0) {
           mask[path_index] = false;
-          mask[prev_index] = false;
-          prev_index--;
           path_index++;
+          if (depth > 0) {
+            mask[prev_index] = false;
+            prev_index--;
```
markusdregi: I believe there is a bug in this function (see #239). But it was here before this PR as well, so I just created an issue for it.

joakim-hove: Yes - you are right; thank you for the thorough review.

lib/util/tests/ert_util_normal_path.c
markusdregi: Btw, are these tests ran at Windows as well? I would expect them to fail due to the forward slash and `util_path_sep` using `UTIL_PATH_SEP_STRING` to split files.

joakim-hove: Ehh - no; as commented here: https://github.com/Statoil/libecl/pull/232#discussion_r149655688

joakim-hove:
OK - implementation completely rewritten; should fix #239

lib/util/tests/ert_util_normal_path.c
markusdregi: Yeah, that is an interesting approach. But beyond the scope of this PR if it is ever to be fixed.

joakim-hove:
> I would be very happy if you would make a test out of the example I made

I have done that.

lib/util/util.c
```diff
@@ -875,78 +875,67 @@ char * util_alloc_realpath__(const char * input_path) {
...
+      path_stack[i] = NULL;
 
     {
-      int path_index = 1;  // Path can not start with ..
-      int prev_index = 0;
-      while (true) {
-        if (path_index == path_len)
-          break;
+      int stack_size = 0;
+      int path_index = 0;
```
markusdregi: This path_index needs to go for libecl to compile.

PR #235, opened 2017-11-06
Nexus to ecl units

JensGM:
Map nexus variable names to unit based on input file unit system.
Only the MATBAR nexus unit system is currently supported. Supported variables
are QOP, QWP, QGP, GOR, WCUT, COP, CWP, CGP, QWI, QGI, CWI, CGI, QPP,
and CPP.

Support for additional variables is done by adding an entry to the map
varname_to_unit_str in `lib/nexus/unit.cpp`.

## TODO
Squash commits to single commit


joakim-hove:
Ehhh; sorry I was a bit slow when reviewing this. This PR handles the unit system for Nexus very well, but it misses the point slightly when it comes to the ECL units; currently this PR passes the unit strings from Nexus directly to Eclipse, but what we are missing is a unit translation system which takes a Nexus unit as input and outputs an eclipse unit string; they are markedly different from the ones you have in this PR.

PR #237, opened 2017-11-07
Ecl sum keyword vector

joakim-hove:
**Task**
When exporting multiple cases in tabular form it is essential that they have the same keys. With this PR we will emit an empty string for missing keys, to ensure that alignment is correct.

Part of: ERT-1435

**Pre un-WIP checklist**

### DIFF
--- a/lib/ecl/ecl_sum_data.c
+++ b/lib/ecl/ecl_sum_data.c
@@ -1180,12 +1180,10 @@ double ecl_sum_data_interp_get(const ecl_sum_data_type * data , int time_index1 
 }
 
 
-static double ecl_sum_data_vector_iget(const ecl_sum_data_type * data,  time_t sim_time, const ecl_sum_vector_type * keylist, int key_index,
+static double ecl_sum_data_vector_iget(const ecl_sum_data_type * data,  time_t sim_time, int params_index, bool is_rate,
                                        int time_index1 , int time_index2 , double weight1 , double weight2 ) {
 
-  int params_index = ecl_sum_vector_iget_param_index(keylist, key_index);
   double value = 0.0;
-  bool is_rate = ecl_sum_vector_iget_is_rate(keylist, key_index);
   if (is_rate) {
     int time_index = ecl_sum_data_get_index_from_sim_time(data, sim_time);
     // uses step function since it is a rate
@@ -1205,12 +1203,21 @@ void ecl_sum_data_fwrite_interp_csv_line(const ecl_sum_data_type * data, time_t 
   ecl_sum_data_init_interp_from_sim_time(data, sim_time, &time_index1, &time_index2, &weight1, &weight2);
 
   for (int i = 0; i < num_keywords; i++) {
-    double value = ecl_sum_data_vector_iget( data, sim_time, keylist , i , time_index1, time_index2, weight1, weight2);
+    int params_index = ecl_sum_vector_iget_param_index(keylist, i);
+    if (params_index >= 0) {
+      bool is_rate = ecl_sum_vector_iget_is_rate(keylist, i);
+      double value = ecl_sum_data_vector_iget( data, sim_time, params_index , is_rate, time_index1, time_index2, weight1, weight2);
 
-    if (i == 0)
-      fprintf(fp, "%f", value);
-    else
-      fprintf(fp, ",%f", value);
+      if (i == 0)
+        fprintf(fp, "%f", value);
+      else
+        fprintf(fp, ",%f", value);
+    } else {
+      if (i == 0)
+        fputs("", fp);
+      else
+        fputs(",", fp);
+    }
   }
 }
 
@@ -1218,12 +1225,18 @@ void ecl_sum_data_get_interp_vector( const ecl_sum_data_type * data , time_t sim
   int num_keywords = ecl_sum_vector_get_size(keylist);
   double weight1, weight2;
   int    time_index1, time_index2;
+  double missing_value = -1;
 
   ecl_sum_data_init_interp_from_sim_time(data, sim_time, &time_index1, &time_index2, &weight1, &weight2);
   double_vector_reset( results );
   for (int i = 0; i < num_keywords; i++) {
-    double value = ecl_sum_data_vector_iget( data, sim_time, keylist , i , time_index1, time_index2, weight1, weight2);
-    double_vector_iset( results, i , value );
+    int params_index = ecl_sum_vector_iget_param_index(keylist, i);
+    if (params_index >= 0) {
+      bool is_rate = ecl_sum_vector_iget_is_rate(keylist, i);
+      double value = ecl_sum_data_vector_iget( data, sim_time, params_index, is_rate, time_index1, time_index2, weight1, weight2);
+      double_vector_iset( results, i , value );
+    } else
+      double_vector_iset( results, i , missing_value);
   }
 }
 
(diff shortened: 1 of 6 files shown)


berland:
Is this commit also handling the case when the number of parameters from parameters.txt vary between realizations?

joakim-hove:
> Is this commit also handling the case when the number of parameters from parameters.txt vary between realizations?

No - this code is not related to `parameters.txt` at all - but the feature request is noted.

lib/ecl/ecl_sum_data.c
```diff
@@ -1205,25 +1203,40 @@ void ecl_sum_data_fwrite_interp_csv_line(const ecl_sum_data_type * data, time_t
...
   for (int i = 0; i < num_keywords; i++) {
-    double value = ecl_sum_data_vector_iget( data, sim_time, keylist , i , time_index1, time_index2, weight1, weight2);
-    double_vector_iset( results, i , value );
+    int params_index = ecl_sum_vector_iget_param_index(keylist, i);
+    if (params_index >= 0) {
+      bool is_rate = ecl_sum_vector_iget_is_rate(keylist, i);
+      double value = ecl_sum_data_vector_iget( data, sim_time, params_index, is_rate, time_index1, time_index2, weight1, weight2);
+      double_vector_iset( results, i , value );
+    } else
+      double_vector_iset( results, i , missing_value);
```
pgdr: I'm not sure I understand this part (I'm sure I don't understand this part) but is it safe to set the values to -1?  Is it necessary?

joakim-hove: > I'm not sure I understand this part (I'm sure I don't understand this part) but is it safe to set the values to -1? Is it necessary?

Good question - I was slightly uncomfortable doing that. The point of the whole PR is that if the keyword vector asks for keywords: `["KEY1", "KEY2", "KEY3"]` and the summary instance in question is missing `KEY2` we still need to return a result with three elements - to ensure that things are correctly aligned. As matters stand it is actyally the `_fprintf( )` function which is the most important - in that case the printed result should:

```txt
100,,200
```

i.e. an empty element in the csv output. We could:

1. [Maybe] just remove the offending function.
2. Pass in a user supplied value to be used in the case of missing values.
3. Use a different value than `-1` - e.g. `nan`.
4. Return an additional mask-like structure to indicate which values are valid.

lib/ecl/ecl_sum_vector.c
```diff
@@ -58,6 +58,27 @@ ecl_sum_vector_type * ecl_sum_vector_alloc(const ecl_sum_type * ecl_sum){
...
+  for (int i=0; i < stringlist_get_size(vector->key_list); i++) {
+    const char * key = stringlist_iget(vector->key_list, i);
+    if (ecl_sum_has_general_var(ecl_sum, key))
+      ecl_sum_vector_add_key(new_vector, key);
+    else
+      ecl_sum_vector_add_invalid_key(new_vector, key);
+  }
+  return new_vector;
+}
+
```
pgdr: Do I understand correctly if `params_index < 0` then the element is invalid?

If so, should maybe have an `ecl_sum_vector_is_valid(const ecl_sum_vector_type * vec, int i)` that returns true if and only if `vec->node_index[i] < 0`?

lib/ecl/ecl_sum_vector.c
```diff
@@ -58,6 +58,27 @@ ecl_sum_vector_type * ecl_sum_vector_alloc(const ecl_sum_type * ecl_sum){
...
+static void ecl_sum_vector_add_invalid_key(ecl_sum_vector_type * vector, const char * key) {
+  int_vector_append(vector->node_index_list, -1);
+  bool_vector_append(vector->is_rate_list, false);
+  stringlist_append_copy(vector->key_list, key);
+}
+
+
+
+ecl_sum_vector_type * ecl_sum_vector_alloc_layout_copy(const ecl_sum_vector_type * vector, const ecl_sum_type * ecl_sum) {
+  ecl_sum_vector_type * new_vector = ecl_sum_vector_alloc(ecl_sum);
```
pgdr: Shouldn't this use `vector->ecl_sum` and not take `ecl_sum` as an argument?  Or perhaps I don't understand this function.   Can you add a comment?

python/tests/ecl/test_sum.py
```diff
@@ -287,23 +288,32 @@ def test_invalid(self):
...
-        t = case.getDataStartTime( ) + datetime.timedelta( days = 43 );
-        data = case.get_interp_row( kw_list , t )
-        for d1,d2 in zip(data, [ case.get_interp("FOPT", date = t),
-                                 case.get_interp("FOPT", date = t),
-                                 case.get_interp("FOPT", date = t) ]):
+        t = case1.getDataStartTime( ) + datetime.timedelta( days = 43 );
+        data = case1.get_interp_row( kw_list , t )
+        for d1,d2 in zip(data, [ case1.get_interp("FOPT", date = t),
+                                 case1.get_interp("FOPT", date = t),
+                                 case1.get_interp("FOPT", date = t) ]):
```
pgdr: I have a question regarding `get_interp_row`, but it doesn't really belong to this PR.  But that function in `c` seems to link `ecl_sum` back to `ecl_sum_vector`.  Could it not be placed in `ecl_sum_vector` instead?

lib/ecl/ecl_sum_vector.c
joakim-hove: > Do I understand correctly if params_index < 0 then the element is invalid?

Yes

> should maybe have an ecl_sum_vector_is_valid(const ecl_sum_vector_type * vec, int i)

YAGNI??

lib/ecl/ecl_sum_vector.c
joakim-hove: > Can you add a comment?

Yes - I will add a comment.

lib/ecl/ecl_sum_data.c
```diff
@@ -1205,25 +1203,40 @@ void ecl_sum_data_fwrite_interp_csv_line(const ecl_sum_data_type * data, time_t
   ecl_sum_data_init_interp_from_sim_time(data, sim_time, &time_index1, &time_index2, &weight1, &weight2);
 
   for (int i = 0; i < num_keywords; i++) {
-    double value = ecl_sum_data_vector_iget( data, sim_time, keylist , i , time_index1, time_index2, weight1, weight2);
+    int params_index = ecl_sum_vector_iget_param_index(keylist, i);
+    if (params_index >= 0) {
```
pgdr: I need it here?  We don't need to _expose_ the function out of this file, but I am confused as to whether `params_index=0` is valid or invalid.  If we had an "iget valid" function, that would be beyond doubt?

python/tests/ecl/test_sum.py
joakim-hove: I have not looked at this code right now - but I *think* you are mxing two different types of vectors:

`ecl_sum_vector`: is one keyword, all timesteps.
`ecl_sum_keyword_vector`: A vector/collection of keywords, which are typically used to evaluate many keywords at the sime time point [which has a significant startup cost].

lib/ecl/ecl_sum_data.c
joakim-hove: >  If we had an "iget valid" function, that would be beyond doubt?

Sure - I will add it.

joakim-hove:
OK - fixup with:
1. Added function to check if index is valid.
2. Comments 
3. Removed setting of invalid value - user should set that beforehand.

PR #238, opened 2017-11-08
add METBAR unit summary file

ReedOnly:
**Task**
Add SPE1 test case with metbar units, which is closer to the units used by Statoil in Eclipse

**Pre un-WIP checklist**
- [ ] implement the tests in the new SPE1 case


JensGM:
Rename to SPE1.plt

PR #240, opened 2017-11-08
nex2ecl command-line options -f, -h, and -o

JensGM:
The nex2ecl application is extended to handle command-line options for
format `(-f|--format)`, help `(-h|--help)`, and output path `(-o|--output)`.

### DIFF
--- a/applications/nexus/nex2ecl.cpp
+++ b/applications/nexus/nex2ecl.cpp
@@ -16,37 +16,148 @@
 */
 
 #include <cctype>
-
-#include <string>
-
+#include <getopt.h>
 #include <iostream>
+#include <string>
 
 #include <ert/util/util.h>
 
 #include "nexus/util.hpp"
 
+namespace {
 
-void create_ecl_sum(char *file_path) {
-    std::string filename(file_path);
-    auto ending = filename.rfind(".plt");
-    if (ending == std::string::npos) {
-        filename += ".plt";
-    }
+static const constexpr char *usage =
+R"(nex2ecl [options]... <nexus_plt_file>
+Try 'nex2ecl --help' for more information.)";
 
-    nex::NexusPlot plt = nex::load(filename);
+static const constexpr char *help =
+R"(Usage:
+  nex2ecl [options]... <nexus_plt_file>
+  nex2ecl -h | --help
+Options:
+  -f --format                Format the output.
+  -h --help                  Show this screen.
+  -o --output <file>|<path>  Write output to files prefixed by <file> or place
+                             output at <path>.
+)";
 
-    char * basename;
-    util_alloc_file_components( file_path , NULL, &basename , NULL);
-    util_strupr( basename );
+static struct option long_options[] = {
+    {"format", no_argument,       nullptr, 'f'},
+    {"help",   no_argument,       nullptr, 'h'},
+    {"output", required_argument, nullptr, 'o'}
+};
 
-    ecl_sum_type *ecl_sum = nex::ecl_summary(basename, plt);
-    ecl_sum_fwrite(ecl_sum);
-    ecl_sum_free(ecl_sum);
-    free( basename );
 }
 
+void create_ecl_sum( std::string input_path,
+                     std::string output_path,
+                     bool format_output ) {
+    auto ending = input_path.rfind(".plt");
+    if (ending == std::string::npos) {
+        input_path += ".plt";
+    }
 
-int main(int argc, char **argv) {
-    create_ecl_sum(argv[1]);
-    exit(0);
+    std::string ecl_case_name {};
+    {
+        char *plt_basename;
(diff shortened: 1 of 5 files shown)


applications/nexus/nex2ecl.cpp
```diff
@@ -16,37 +16,148 @@
...
+
+            if ( !basename ) {
+                std::cerr << "Failed to parse output path "
+                          << output_path << "." << std::endl;
+                exit(1);
+            }
+
+            if ( path && !util_is_directory(path) ) {
+                std::cerr << path << " does not exists." << std::endl;
+                exit(1);
```
joakim-hove: Could potentially do `util_make_path( path )` here.

applications/nexus/nex2ecl.cpp
```diff
@@ -16,37 +16,148 @@
...
+        util_alloc_file_components( input_path.c_str(),
+                                    NULL,
+                                    &plt_basename,
+                                    NULL );
+        util_strupr( plt_basename );
+
+        if ( output_path.empty() ) {
+            ecl_case_name = std::string( plt_basename );
+        } else if ( util_is_directory(output_path.c_str()) ) {
+            ecl_case_name = output_path + std::string( plt_basename );
```
joakim-hove: Looks to me like a path separator is missing here?

applications/nexus/nex2ecl.cpp
```diff
@@ -16,37 +16,148 @@
...
+Try 'nex2ecl --help' for more information.)";
+
+static const constexpr char *help =
+R"(Usage:
+  nex2ecl [options]... <nexus_plt_file>
+  nex2ecl -h | --help
+Options:
+  -f --format                Format the output.
+  -h --help                  Show this screen.
+  -o --output <file>|<path>  Write output to files prefixed by <file> or place
```
joakim-hove: I think: ` Write output to files with basename <file> or place` will be more familiar to users.

applications/nexus/nex2ecl.cpp
```diff
@@ -16,37 +16,148 @@
...
-    util_alloc_file_components( file_path , NULL, &basename , NULL);
-    util_strupr( basename );
+            if ( path ) {
+                ecl_case_name += std::string( path );
+                ecl_case_name += UTIL_PATH_SEP_CHAR;
+            }
+            ecl_case_name += std::string( basename );
 
-    ecl_sum_type *ecl_sum = nex::ecl_summary(basename, plt);
+            if ( path ) free( path );
```
joakim-hove: As long as `path` and `basename`have been ǸULL initialized you *can* call `free( )` unconditionally.

lib/nexus/nexus_plot.cpp
```diff
@@ -314,7 +315,7 @@ ecl_sum_type* nex::ecl_summary(const std::string& ecl_case, const NexusPlot& plt
     bool time_in_days = true;
 
     ecl_sum_type * ecl_sum = ecl_sum_alloc_writer( ecl_case.c_str(),
-        fmt_output,
+        format_output,
```
joakim-hove: Extremely ugly that the `bool format_output` comes as part of the `ecl_sum_alloc_writer( )` call - that should have been in the `ecl_sum_fwrite( )` call; but that is for another PR.

applications/nexus/nex2ecl.cpp
```diff
@@ -16,37 +16,148 @@
 */
 
 #include <cctype>
-
-#include <string>
-
+#include <getopt.h>
```
joakim-hove: This should be protected in some way. The symbol `ERT_HAVE_GETOPT` does the cmake check to look for `getopt.h` - my suggestion would be to make the complete build of `nex2ecl` conditional on that symbol.

applications/nexus/nex2ecl.cpp
```diff
@@ -16,37 +16,148 @@
...
+
+    bool format_output = false;
+    std::string output_path {};
+    std::string input_path {};
+
+    /*
+     * Parse options
+     */
+
+    int c;
```
joakim-hove: No reason to declare `c`outside the `for` loop?

applications/nexus/nex2ecl.cpp
JensGM: This was forgotten, will add.

applications/nexus/nex2ecl.cpp
JensGM: Agree

applications/nexus/nex2ecl.cpp
JensGM: Makes sense, I'm also adding some changes to the way the basename vs path is handled

applications/nexus/nex2ecl.cpp
JensGM: No, there isn't, fixing

PR #242, opened 2017-11-09
Test area fixup

joakim-hove:
**Task**
Squelch some warnings.


joakim-hove:
Linux works - the OSX nodes seem broken. Merging.

PR #244, opened 2017-11-09
Nexus unit conversions

JensGM:
Member function conversion is added to nexus/UnitSystem. This is to
facilitate conversions from nexus units to ecl units, in particular kRM3
to RM3 and kSM3 to SM3. Currently only the conversions for nexus unit
system METBAR is supported. The conversion table is located in
`lib/nexus/unit.cpp`.

This PR requires PR https://github.com/Statoil/libecl/pull/243

### DIFF
--- a/lib/nexus/include/nexus/unit.hpp
+++ b/lib/nexus/include/nexus/unit.hpp
@@ -12,7 +12,8 @@ namespace nex {
             lab,
             metric_kPa,
             metric_kg_cm2,
-            metric_bars
+            metric_bars,
+            unit_type_count
         };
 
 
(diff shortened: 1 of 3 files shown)


lib/nexus/include/nexus/unit.hpp
```diff
@@ -57,13 +59,19 @@ namespace nex {
             // transmissibility,
             viscosity,
             volume,
-            water_cut
+            water_cut,
+            measure_count
```
joakim-hove: I would guess count and identity - i.e. no dimensionless number - could be the same?

JensGM: This did confuse me somewhat as well actually. That entry wasn't actually meant to be a measure, but the count of measures in the enum.

JensGM: I've renamed it to make this clearer

PR #245, opened 2017-11-11
Added small decorator to log test methods.

joakim-hove:


python/python/ecl/test/extended_testcase.py
```diff
@@ -28,6 +28,21 @@
     sys.stderr.write("Warning: could not import file test_env.py - this might lead to test failures.")
 
 
+# Function wrapper which can be used to add decorator @log_test to test
+# methods. When a test has been decorated with @log_test it will print
+# "starting: <test_name>" when a method is complete and "complete: <test_name>"
```
andreabrambilla: "is complete" should probably be "is starting" or something similar

joakim-hove: Fixed - merging.

PR #248, opened 2017-11-14
Fixed two bugs in the nexus code

JensGM:
One bug caused all variables in WELL instances to be 0, the other was
that nexus to eclipse unit conversion were never applied.


joakim-hove:
When fixing bugs you should strive hard to start with a test that fails.

JensGM:
@joakim-hove In this case there isn't a test that fails, but maybe this PR should also include a test that would fail without the fix?

joakim-hove:
> In this case there isn't a test that fails, 

No - that goes without saying; otherwise this would never have been merged.

> but maybe this PR should also include a test that would fail without the fix?

Yes - that is *exactly* how you should have started the process of fixing the bug.

PR #250, opened 2017-11-20
Write nexus data to correct ecl timesteps

JensGM:
Before this patch, nexus data written to ecl was assumed to start at
timestep 0. In ecl, all keywords and variables are present from timestep
0. In nexus, however, instances of regions, well, etc. are not present
in the dataset before the timestep at which the first appear. This
caused instances that appear at a later timestep to start at timestep 0.
Now the data is written at the timestep it appears.


joakim-hove:
https://github.com/Statoil/libecl/pull/248#issuecomment-344246065

JensGM:
@joakim-hove To test this we'd need a dataset where this occurs. At the moment we don't have any open datasets where this change is relevant.

pgdr:
@JensGM Equally important to the code that you have written that (supposedly [since there are no tests]) fixes the issue, is the test that guarantees that the issue is not reintroduced.

Are you saying that it is infeasible to create a test for this issue?

_Untested functionality is extremely volatile._

JensGM:
@pgdr Creating the test will not be very difficult. It's just that we don't have something to test it on. Getting a nexus version of Norne would be handy. We did, however, verify that the data was correct after the fix by visualizing it in s3graf.

joakim-hove:
> It's just that we don't have something to test it on.

Well - then you have to make something. I suggest you create a copy of the SPE1 case and modify it so that one of the wells is not introduced before some timesteps have passed.

joakim-hove:
I realize it is not part of this PR, but I would very much appreciate if the variable `smspec_nodes` [here](https://github.com/JensGM/libecl/blob/3915dbdada60902879260537ae0f9971c5d537a6/lib/nexus/nexus_plot.cpp#L409) was renamed to e.g. `ecl_values` or something like that - the point is that the name `smspec_nodes` indicates that it is a list of exactly `smspec_nodes` - but it is in addition timesteps *and* actual values, both of which are orthogonal to my expectation when it comes to the content of a variable with name `smspec_nodes`.

JensGM:
@joakim-hove I'll rename the variable and look into making a data set.

ReedOnly:
I modified the SPE1 test case in #251 to introduce a 3rd well starting to produce 4 years after the others. This is clearly not handled by inspecting the result in S3GRAF and can be used to make a test for this PR.

JensGM:
@joakim-hove There is now a test for this behavior and I've renamed `smspec_nodes` to `ecl_values`

PR #251, opened 2017-11-20
Add nexus test case with delayed well

ReedOnly:
Production well 3 is added to SPE1 case which becomes active on 01/01/1984, 4 years after well 1 starts producing. With this we can make a test for fixing the problem of wrong time scale.


joakim-hove:
Very good. I would prefer if the new file was added as e.g. SPE1_extra_well, so that the original SPE1 was retained.

PR #253, opened 2017-11-24
Added requirements.txt for pip install of requirements.

joakim-hove:
**Task**
Use external cwrap + requirements.txt for `pip install`

**Pre un-WIP checklist**

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,10 +65,11 @@ install:
   - export CONDA_HOME="$HOME/miniconda"
   - export PATH="$CONDA_HOME/bin:$PATH"
   - hash -r
+  - pip install -r requirements.txt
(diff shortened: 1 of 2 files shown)


andreabrambilla:
Out of curiosity, why numpy ended up in requirements.txt, but pylint and pandas are "manually" installed?

pgdr:
Pandas should pull numpy by itself in conda, pylint is only test-requirement.

andreabrambilla:
I thought so about pylint, but I still don't get the thing with pandas.

Is pandas required only by the tests as well? Or is it actually required by libecl?

joakim-hove:
> Is pandas required only by the tests as well? Or is it actually required by libecl?

The pandas requirement is not for real - that is stale (I think ...)

PR #254, opened 2017-11-28
Added switch to disable internal cwrap

joakim-hove:
**Task**
Preparations for using external cwrap. Combined with ert#86 the default will be to build with external, i.e. `pip install` cwrap on Travis when this and ert#86 have been merged. 

**Pre un-WIP checklist**

Statoil/ert#86

### DIFF
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -7,12 +7,17 @@ init_python( 2.7 )
 
 find_python_package(numpy 1.7.1 ${PYTHON_INSTALL_PREFIX})
 if (NOT DEFINED PY_numpy)
-   message(WARNING "numpy module not found - Python wrappers not enabled")
+   message(WARNING "numpy module not found - Python wrappers not enabled. Install with: \"pip install numpy\"")
    set( BUILD_PYTHON OFF PARENT_SCOPE )
    return()
 endif()
 
 find_python_package(cwrap 1 ${PYTHON_INSTALL_PREFIX})
+if (NOT DEFINED PY_cwrap)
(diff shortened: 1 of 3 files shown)


python/CMakeLists.txt
```diff
@@ -7,12 +7,17 @@ init_python( 2.7 )
...
-   message(WARNING "numpy module not found - Python wrappers not enabled")
+   message(WARNING "numpy module not found - Python wrappers not enabled. Install with: \"pip install numpy\"")
    set( BUILD_PYTHON OFF PARENT_SCOPE )
    return()
 endif()
 
 find_python_package(cwrap 1 ${PYTHON_INSTALL_PREFIX})
+if (NOT DEFINED PY_cwrap)
+  if (NOT INSTALL_CWRAP)
+    message(WARNING "cwrap module not found Python wrappers not enabled. Install with: \"pip insall cwrap\"")
```
pgdr: typo

pgdr:
Looks okay, but what means are in place to keep `cwrap` here up-to-date?  Should we perhaps remove the `cwrap` folder, add a git submodule, and a symlink to that?

python/CMakeLists.txt
joakim-hove: Fixed

joakim-hove:
> but what means are in place to keep cwrap here up-to-date? 

None whatsoever.

> Should we perhaps remove the cwrap folder, add a git submodule, and a symlink to that?

We should remove the `cwrap` folder here in this project; and then possible add a git submodule - or alternatively just state `cwrap` *is*  dependency which should be pip installed.

PR #257, opened 2017-12-01
Identify Flag for Constant Oil Compressibility

bska:
This commit is based on reverse engineering a set of INIT files.  Item 39 (i.e., index 38 in zero-based indexing) of LOGIHEAD appears to be true if the simulation run uses constant oil compressibility (typically in the case of the PVCDO oil property keyword).

### DIFF
--- a/lib/include/ert/ecl/ecl_kw_magic.h
+++ b/lib/include/ert/ecl/ecl_kw_magic.h
@@ -341,6 +341,7 @@ values (2e20) are denoted with '*'.
 #define LOGIHEAD_REV_ENDPOINT_SCALING_INDEX        18
 #define LOGIHEAD_ALT_ENDPOINT_SCALING_INDEX        19
 #define LOGIHEAD_MISC_DISPLACEMENT_INDEX           35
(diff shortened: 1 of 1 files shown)


bska:
Note: This is very much a work in progress.  The reverse engineering is based on a limited set of `INIT`-file examples and the implementation will probably need adjustment to conform to LibECL standards.

joakim-hove:
This is marked as [WIP] - and you describe it as work in progress, but I am happy to merge this the moment you say so.

bska:
> This is marked as [WIP]

Not for much longer.

> and you describe it as work in progress,

It sort of is.  The main reason for not declaring it "final" is that I don't have access to a comprehensive set of cases that vary **only** the oil property keywords (i.e., between `PVDO`, `PVTO` and `PVCDO`).  I also haven't figured out how (or if) `PVCO` (live oil in compressibility form) fits into the picture.

That said and based on the data available to me, I am nevertheless reasonably convinced that item 39 of the (INIT file's main grid's) LOGIHEAD vector really does signify whether or not the oil compressibility is constant (true for constant compressibility, false otherwise).

I'm therefore going to remove the "[WIP]" tag and solicit review/comments on the naming of the manifest constant and whether to write `#define ... 38` or if the current `#define ... (39-1)` is acceptable.

PR #260, opened 2017-12-06
Will silently ignore ImportError for test env.

joakim-hove:
**Task**
Remove/soften dependency on `test_env` module.

**Pre un-WIP checklist**


joakim-hove:
Getting this fully verified requires running tests - with internal testdata - in several repositories. I will merge now - and hopefully never get back to it.

PR #262, opened 2017-12-12
make option for different fields to output

ReedOnly:
Until now classname NETWORK was assumed to be the main field, but experience from recent plotfiles shows that this is not the case, and the NETWORK class may not even be present. Default is changed to FIELD, and can be specified with the -F <field> option.

### DIFF
--- a/lib/nexus/nexus_plot.cpp
+++ b/lib/nexus/nexus_plot.cpp
@@ -293,7 +293,8 @@ struct eclvar {
     int32_t nexus_timestep;
 };
 
-void field_smspec( std::vector< eclvar >& ecl_values,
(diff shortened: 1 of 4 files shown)


lib/nexus/nexus_plot.cpp
```diff
@@ -390,7 +394,8 @@ void well_smspec( std::vector< eclvar >& ecl_values,
 
 ecl_sum_type* nex::ecl_summary(const std::string& ecl_case,
                                bool format_output,
-                               const NexusPlot& plt) {
+                               const NexusPlot& plt,
+                               const std::string field_name) {
```
joakim-hove: This could be a reference.

PR #270, opened 2018-01-02
fixed: python package dependency

akva2:
**Task**
Debian packaging

**Approach**
Let loose on the world, then let the users bleed.

**Pre un-WIP checklist**


joakim-hove:
> Let loose on the world, then let the users bleed.

So - is this ready?

akva2:
yes. already built new packages using it etc. we found the issue through what you quote ;)

PR #278, opened 2018-01-02
WIP: Extended testcase class attributes

joakim-hove:
**Task**
The test class `ExtendedTestCase` in the `ecl.test` contains a mixture of generic testing functionality, which is also used in libres, ert and ert-statoil, and settings specific to the libecl repository. With this PR these things should be split.

Statoil/libres#197
Statoil/ert#90

### DIFF
--- a/python/ecl/test/extended_testcase.py
+++ b/python/ecl/test/extended_testcase.py
@@ -13,13 +13,6 @@ from .source_enumerator import SourceEnumerator
 from ecl.util import installAbortSignals
 from ecl.util import Version
 
-TESTDATA_ROOT = None
-SHARE_ROOT = None
-SOURCE_ROOT = None
-try:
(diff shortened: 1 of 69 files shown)


pgdr:
Are you planning on also deleting `ecl.test`?

joakim-hove:
> Are you planning on also deleting ecl.test?

The current `ecl.test` contains a mix of reasonably generic test functionality which is used in both this repository and in other repositories, and some not-really-well-working functionality designed to test specifics - mainly from the libres repository. The ambition for this task[1] is just to clean up & split this better, so that everything left in `ecl.test` is generic and stateless. Maybe it should be removed/trimmed down in the future, but that is not part of this task.

[1]: This task again is just about streamlining to facilitate `pip install libecl`

PR #281, opened 2018-01-04
Reorganized Statoil based python testing.

joakim-hove:
**Task**
This is again part of the preparations for `pip install libecl`. Have moved the conditional logic related to Statoil testdata from cmake to Python.

**Approach**
Created decorator `statoil_test` which can be used on tests which require the Statoil testdata, not really happy with this decorator. Improvements welcome!

**Pre un-WIP checklist**


joakim-hove:
> I like the PR, I find it generally good to reduce the cmake code to the minimum. 

:-)

> Out of curiosity, why is this PR required for supporting `pip install`?

Not for `pip install` as such, but for testing with: `setup.py test`

PR #285, opened 2018-01-08
WIP: Implement summary resample

stefoss23:
**Task**
To ensure that the different realisations generated by Nexus can be loaded in ERT they must have the same time axis. To facilitate we should create a ecl_sum function:

ecl_sum_type * ecl_sum_alloc_resample(const char * ecl_case, const ecl_sum_type * ecl_sum, const time_t_vector_type * times), 

which should allocate a new summary case resampled to the times in the time_t_vector_type input vector.

**Approach**
Function added to ecl_sum. 

**Pre un-WIP checklist**

### DIFF
--- a/lib/ecl/ecl_sum.c
+++ b/lib/ecl/ecl_sum.c
@@ -714,9 +714,6 @@ ecl_sum_type * ecl_sum_alloc_resample(const char * ecl_case, const ecl_sum_type 
   if ( !time_t_vector_is_sorted(times, false) )
     return NULL;
   
-  bool   fmt_case        = ecl_sum->fmt_case;
-  bool   unified         = ecl_sum->unified;
-  char * key_join_string = ecl_sum->key_join_string;
   const int * grid_dims  = ecl_smspec_get_grid_dims(ecl_sum->smspec);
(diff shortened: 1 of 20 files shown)


lib/ecl/ecl_sum.c
```diff
@@ -690,6 +702,56 @@ const char * ecl_sum_get_general_var_unit( const ecl_sum_type * ecl_sum , const
...
+  time_t start_time = ecl_sum_data_get_data_start(ecl_sum->data);
+
+  if ( time_t_vector_get_first(times) < start_time )
+    return NULL;
+  if ( time_t_vector_get_last(times) > ecl_sum_data_get_sim_end(ecl_sum->data) )
+    return NULL;
+  if ( !time_t_vector_is_sorted(times, false) )
+    return NULL;
+  
+  bool   fmt_case        = ecl_sum->fmt_case;
```
joakim-hove: Not necessary to create local variables for these - just use `ecl_sum->fmt_case`, `ecl_sum->unified` and so on in the `ecl_sum_alloc_writer( )` call.

PR #288, opened 2018-01-10
WIP: removed log.c with associated files and functionality.

stefoss23:
**Task**
Log.c and associated files will be moved from libecl to libres. It is a wish to mold libecl for pure eclipse functionality.

**Approach**

**Pre un-WIP checklist**

Statoil/libres#203


joakim-hove:
OK - this looks fine; I guess you only need to add the [MessageLevelEnum](https://github.com/Statoil/libecl/blob/master/python/ecl/util/enums/message_level_enum.py)

PR #297, opened 2018-01-22
posix_spawn: create new process group for child process.

joakim-hove:
**Task**
Make sure spawned out processes get their own pgroup.

**Pre un-WIP checklist**

### DIFF
--- a/lib/util/util_spawn.c
+++ b/lib/util/util_spawn.c
@@ -20,13 +20,13 @@ extern char **environ;
 #define STDERR_FILENO 2
 
 
-static void __init_attributes(posix_spawnattr_t * attributes) {
+static void spawn_init_attributes__(posix_spawnattr_t * attributes) {
   posix_spawnattr_init(attributes);
   attributes->__flags |= POSIX_SPAWN_SETPGROUP;
   attributes->__pgrp= 0;
 }
 
-static void __init_redirection(posix_spawn_file_actions_t * file_actions, const char *stdout_file, const char *stderr_file) {
+static void spawn_init_redirection__(posix_spawn_file_actions_t * file_actions, const char *stdout_file, const char *stderr_file) {
   int status;
   status = posix_spawn_file_actions_init(file_actions);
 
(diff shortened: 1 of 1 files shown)


pgdr:
Shouldn't `spawn` be moved to res while we're at it?

pgdr:
And shouldn't we stop using `__varname` in C?  For instance, isn't `__argv` some preserved runtime variable?

joakim-hove:
> Shouldn't spawn be moved to res while we're at it?

OK - will do.

> And shouldn't we stop using __varname in C? 

Good point - I think `__symbol` is reserved in the sense that the standard library might introduce it as a conflicting symbol any day.

joakim-hove:
> Shouldn't spawn be moved to res while we're at it?

I started the process, but the traceback in the `util_abort( )` function is based on calling the `addr2line` program with the `util_spawn( )` function, would therefor at least like to wait with this removal.

pgdr:
Well, `util_abort` doesn't belong in the public API of ecl either, but I agree that we can wait with moving it out.

lib/util/util_spawn.c
```diff
@@ -20,13 +20,13 @@ extern char **environ;
 #define STDERR_FILENO 2
 
 
-static void __init_attributes(posix_spawnattr_t * attributes) {
+static void spawn_init_attributes__(posix_spawnattr_t * attributes) {
```
pgdr: I would guess that if you absolutely need to annotate the function names, then _one_ trailing underscore would suffice?

joakim-hove:
> Well, util_abort doesn't belong in the public API of ecl either,

No - that might be true; but the backtrace capabilities of `util_abort(...)` have in the past proved tremendously valuable when debugging. The bug rate in libecl is not very high - but it still feels a bit brutal to say good bye to the backtrace possibility.

pgdr:
I think this patch should have been reviewed before merging.

joakim-hove:
> I think this patch should have been reviewed before merging.

Well; I thought you had reviewed it?

PR #302, opened 2018-01-26
Fixup in numpy polyfit use

joakim-hove:
**Task**
The PR: #300 was not really tested - relevant data was internal. Fixup here.

**Pre un-WIP checklist**


joakim-hove:
OK - this works on the internal tests and fixes the build again. Merging.

PR #304, opened 2018-01-29
Using member functions for posix_spawn_attributes

stefoss23:
**Task**
There is a compile error concerning util_spawn.c when building with clang. 

**Approach**
Possibly, clang does not comprehend accessing posix_spawnattr_t variables directly the way gcc does. Instead, indirect access with member functions are used. 

**Pre un-WIP checklist**

### DIFF
--- a/lib/util/util_spawn.c
+++ b/lib/util/util_spawn.c
@@ -19,11 +19,15 @@ extern char **environ;
 #define STDOUT_FILENO 1
 #define STDERR_FILENO 2
 
-
 static void spawn_init_attributes__(posix_spawnattr_t * attributes) {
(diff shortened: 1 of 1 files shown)


joakim-hove:
This looks like a correct fix; should probably in addition replace the line:
```C
attributes->__pgrp= 0;
```
with:
```
posix_spawnattr_setpgroup( attributes, 0);
```
See: https://opensource.apple.com/source/xnu/xnu-3789.41.3/libsyscall/wrappers/spawn/posix_spawn.c.auto.html

pgdr:
Is this related to #297 and Statoil/libres#205?

PR #308, opened 2018-01-31
Should correctly infer report step from filenames like: "case.x0078"

joakim-hove:
Should fix: https://github.com/Statoil/libecl/issues/303


lib/include/ert/ecl/ecl_util.h
```diff
@@ -118,8 +118,9 @@ bool            ecl_util_unified_file(const char *filename);
 const char    * ecl_util_file_type_name( ecl_file_enum file_type );
 char          * ecl_util_alloc_base_guess(const char *);
 int             ecl_util_filename_report_nr(const char *);
-ecl_file_enum   ecl_util_get_file_type(const char * , bool * , int * );
 ecl_file_enum   ecl_util_inspect_extension(const char * ext , bool *_fmt_file, int * _report_nr);
+ecl_file_enum   ecl_util_get_file_type__(const char * , bool * , int * , bool * valid_case);
```
pgdr: I thought double underscore meant "private" or "secret"?  Why use it here, and why expose it in the header?

lib/ecl/ecl_util.c
```diff
@@ -120,38 +120,40 @@ char * ecl_util_alloc_base_guess(const char * path) {
   return base;
 }
 
+static bool is_valid_case(const char * filename) {
```
pgdr: perhaps add a comment on what a valid case, or better yet, what constitutes a valid filename?

lib/include/ert/ecl/ecl_util.h
pgdr: I must say that I find this API very peculiar.
```c
ecl_util_get_file_type(const char * , bool * , int * );
```
What are the necessary parameters here?

joakim-hove: > I thought double underscore meant "private" or "secret"? Why use it here, and why expose it in the header?

That is loosely what I imply here - or ~alternatively "the implementation" of the `ecl_util_get_file_type()` function - which is probably a more correct description in this case. It is used by the Python code, which can then do slightly more elaborate error checking based on this function; as seen from the C code it could have been `static`. To be visible for `dlopen()` from Python it can not be `static` - but I agree it could be removed from the header; can of course also change the name - but that is "reasonably consistent" - or at least not wildly out of line - with the rest of the code.

joakim-hove: > I must say that I find this API very peculiar.

Hehe - yes I can understand that; it is ~10 years old. It is documented in the `ecl_util.c` file; will add argument names in the header.

joakim-hove:
OK - added some comments and removed `ecl_util_get_file_type__()` from header file.

joakim-hove:
Further comments?

lib/ecl/ecl_util.c
```diff
@@ -121,37 +121,49 @@ char * ecl_util_alloc_base_guess(const char * path) {
 }
 
 
+/*
+  The eclipse filenames must have either ALL_CAPITAL.EGRID or alternatviely
+  all_lowercase.egrid; mixed case is illegal. This function scans all the
+  characters after the last '/' and checks if they are all lowercase, or all
+  UPPER CASE.
+*/
```
pgdr: Why is mixed case illegal?

joakim-hove: Ask Schlumberger, I guess it is to ensure compliance on case insensitive filsystems

pgdr: Okay, but do _we_ have to follow Schlum here?  Can't we read a file called `JohanSverdrup.EGRID` without issues?

lib/ecl/ecl_util.c
```diff
@@ -121,37 +121,49 @@ char * ecl_util_alloc_base_guess(const char * path) {
...
+
+    if (islower(c))
+      contains_case |= lower_case;
+
+    if (contains_case == mixed_case)
       break;
   }
 
-  return valid;
+  return (contains_case != mixed_case);
```
pgdr: I think it would be easier to read if it was something like
```c
static bool uniform_case(const char * filename) {
  bool lower = true;
  bool upper = true;
  const char *name = strrchr(filename, UTIL_PATH_SEP_CHAR);
  for (int i = 0; i < strlen(name); i++) {
    if (isupper(name[i]))
      upper = true;
    if (islower(name[i]))
      lower = true;
  }
  return !(upper && lower);
}
```

lib/ecl/ecl_util.c
joakim-hove: That I agree - I think[1] we can read a file with any name if it is passed directly to the open function like:
```
grid = EclGrid("my_really_FUNNY_GRID.txt")
file = EclFile("MixedCase.UnRst")
```
However the functions this PR is about is related to create filename from components, infer types and so on - and then I think it makes sense to follow the conventions; but maybe not?


[1]: Will verify with a test!

joakim-hove:
OK - this has been simplifed significantly; the PR now just checks that we can correctly extract information from a filename - whatever case mix there is.

PR #317, opened 2018-02-02
Make libecl compatible with Python 3

JensGM:
Several things need to be changed for libecl to be compatible with python 3. 

This PR depends on https://github.com/Statoil/cwrap/pull/16

### DIFF
--- a/python/tests/ecl_tests/test_grid_generator.py
+++ b/python/tests/ecl_tests/test_grid_generator.py
@@ -16,7 +16,10 @@
 #  for more details.
 
 from itertools import product as prod
-import operator, random, numpy
+import operator
+import random
+import numpy
+import six
 
 from ecl import EclDataType
 from ecl.eclfile import EclKW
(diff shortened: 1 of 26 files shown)


python/tests/ecl_tests/test_grid_generator.py
```diff
@@ -113,7 +116,8 @@ def test_extract_grid_slice_spec(self):
     def assertSubgrid(self, grid, subgrid, ijk_bound):
         sijk_space = prod(*[range(d) for d in subgrid.getDims()[:-1:]])
         for sijk in sijk_space:
-            gijk = tuple([a+b for a, b in zip(sijk, zip(*ijk_bound)[0])])
+            gijk = tuple([a + b for a, b in
+                         zip(sijk, list(zip(*ijk_bound))[0])])
```
pgdr: Here you actually don't need the `[...]`, you could write `tuple(a + b for a, b in ...)` but just leave it as it is.

PR #319, opened 2018-02-02
Summary restart load permission problem

joakim-hove:
**Task**
The problem: A summary case refers to another case which it has been restarted from. With a special configuration of cwd and path to the history case this would fail hard. This is fixed here: #305. As part of fixing that problems where encountered when you do not have file system access to the history case. With this PR we have basic access (read) control.

NB: `-Werror` bug fixed here: #321 

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -216,6 +216,8 @@ include(CheckIncludeFile)
 include(CheckSymbolExists)
 include(CheckTypeSize)
 
+check_function_exists( access HAVE_POSIX_ACCESS)
+check_function_exists( _access HAVE_WINDOWS__ACCESS)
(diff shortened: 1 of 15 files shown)


markusdregi:
@joakim-hove The linux builds are also failing now.. Could you please fix that?

joakim-hove:
> @joakim-hove The linux builds are also failing now.. Could you please fix that?

I think you are on the wrong PR. #305 is the one with the mac problems.

The errors in this PR are due to `-Werror`[1] - they are fixed here: #321 - which should be merged first.


[1]: Why this suddenly appeared now??????

markusdregi:
Ops! I'll shift my focus to that one instead. Thanks!

PR #334, opened 2018-02-16
P2 p3 changes

stefoss23:
**Task**
Enable even more tests to  succeed in python 3. Also enable any python tests to run in python 3 in travis.

**Approach**
A short description of the approach.

**Pre un-WIP checklist**

### DIFF
--- a/python/ecl/__init__.py
+++ b/python/ecl/__init__.py
@@ -80,10 +80,10 @@ __version__ = "0.0.0"
 #    module should contain the variable lib_path pointing to the
 #    directory with shared object files.
 try:
-    import __ecl_lib_info
-    ecl_lib_path = __ecl_lib_info.lib_path
-    ert_so_version = __ecl_lib_info.so_version
-    __version__ = __ecl_lib_info.__version__
+    from .__ecl_lib_info import EclLibInfo
+    ecl_lib_path = EclLibInfo.lib_path
+    ert_so_version = EclLibInfo.so_version
+    __version__ = EclLibInfo.__version__
 except ImportError:
     pass
 except AttributeError:
--- a/python/ecl/grid/ecl_region.py
+++ b/python/ecl/grid/ecl_region.py
@@ -29,6 +29,7 @@ import ctypes
 
 from cwrap import BaseCClass
 
+import ecl
 from ecl.util.util import monkey_the_camel
 from ecl.util.util import IntVector
 
@@ -927,7 +928,11 @@ class EclRegion(BaseCClass):
             else:
                 raise TypeError("Type mismatch")
         else:
-            self.scale_kw( target_kw , 1/other , force_active )
+            if target_kw.data_type.is_int():
+                scale = 1 // other
+            else:
+                scale = 1.0 / other
+            self.scale_kw( target_kw , scale , force_active )
 
 
     def copy_kw( self , target_kw , src_kw , force_active = False):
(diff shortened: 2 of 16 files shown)


JensGM:
Are there any tests that fail?

python/ecl/__init__.py
```diff
@@ -80,15 +81,27 @@
...
     pass
 except AttributeError:
     pass
 
+#This are python2 -> python3 intermediate arithmetic functions:
+
+#Same functionality as for python2 return x/y: 
+#if either x or y is a float the return value is a float
+#otherwise the result i rounded down (5/2 -> 2, -5/2 -> -3)
+def python2_div(x, y):
```
JensGM: This does not seem like a good solution. I would rather import the new division for python2 and change the implementation to use `//` where its needed, if it is *really* needed.

python/ecl/grid/ecl_region.py
```diff
@@ -927,7 +928,7 @@ def idiv_kw( self , target_kw , other , force_active = False):
             else:
                 raise TypeError("Type mismatch")
         else:
-            self.scale_kw( target_kw , 1/other , force_active )
+            self.scale_kw( target_kw , ecl.python2_div(1, other) , force_active )
```
JensGM: This seem like it should really use a true division as it is a scaling function (floats make sense).

python/tests/ecl_tests/test_grdecl.py
```diff
@@ -20,7 +20,7 @@
 from ecl.grid import EclGrid
 from tests import EclTest, statoil_test
 
-
+from cwrap import Prototype, CFILE, load, open as copen
```
JensGM: Do you need `load` and `CFILE` here?

python/ecl/__init__.py
pgdr: I agree with Jens, `from future import division` together with `//` where needed should be enough.

python/tests/ecl_tests/test_grdecl.py
stefoss23: No, not all of it. It was a copy paste.

python/ecl/grid/ecl_region.py
stefoss23: I will try to use the suggested solutions here.

stefoss23:
Answer to question by JensGM: Although TRAVIS processes tests without error in libecl and in Python3, there are a few tests that still fail on my computer. This is my main dilemma here.

python/ecl/grid/ecl_region.py
JensGM: The tests may be wrong if they rely on integer division. I had a similar issue when I first ported libecl to python3. I let the error be and used integer division, but made a comment in the code that the tests are wrong and created an issue on github.

python/ecl/well/well_connection.py
```diff
@@ -65,6 +65,9 @@ def connectionFactor(self):
     def __eq__(self, other):
         return self._equal(other)
 
+    def __hash__(self):
+        return id(self)
```
pgdr: It would be better to return e.g. `hash(repr(self))` than `id`

python/tests/ecl_tests/test_ecl_sum_vector.py
```diff
@@ -47,7 +47,8 @@ def test_basic(self):
         pfx = 'EclSum(name'
         self.assertEqual(pfx, repr(self.ecl_sum)[:len(pfx)])
         it = iter(self.ecl_sum)
-        t = self.ecl_sum[it.next()] # EclSumVector
+        #t = self.ecl_sum[it.next()] # EclSumVector
```
pgdr: uncommented code should be removed

PR #339, opened 2018-02-20
Added simple function to inplace add the square of an ecl_kw instance.

joakim-hove:
For performance

**Pre un-WIP checklist**


joakim-hove:
Continued in #344

PR #340, opened 2018-02-20
Implemented function to create ACTNUM from PORV

joakim-hove:
See: https://github.com/Statoil/libecl/issues/338


joakim-hove:
> It would be cool if you also had a test verifying that the ACTNUM was the same as number of active cells. 

Hmmm - I feel that the final assert:
```Python
self.assertEqual(tuple(actnum), (0,1,1,0))
```
should verify that?

PR #341, opened 2018-02-21
Added function to create scattered EclKW copy

joakim-hove:
**Task**
Function to create "global" EclKW instance from a compact representation and an ACTNUM vector.

**Pre un-WIP checklist**


joakim-hove:
Continued in #344

PR #349, opened 2018-03-11
[WIP] Comment out signal handling

markusdregi:
**Task**
_Enforce no signal handling - if ever to be merged a more elegant solution will come this way.._

**Approach**

**Pre un-WIP checklist**


joakim-hove:
You beat me to it with a coupld of hours; I have just pushed the same changes directly to an upstream branch: https://github.com/Statoil/libecl/tree/default-signals - will try everest with that branch now.

PR #355, opened 2018-03-14
Will look for the INTEHEAD keyword an extract unit system.

joakim-hove:
Internalize an ID for the unit system used when loading summary files.

Feature requested by web portal.

**Pre un-WIP checklist**

### DIFF
--- a/lib/ecl/ecl_smspec.c
+++ b/lib/ecl/ecl_smspec.c
@@ -141,6 +141,7 @@ struct ecl_smspec_struct {
   float_vector_type * params_default;
 
   char              * restart_case;
+  ert_ecl_unit_enum   unit_system;
 };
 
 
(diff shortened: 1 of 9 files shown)


lib/ecl/ecl_smspec.c
```diff
@@ -367,6 +375,15 @@ static ecl_data_type get_wgnames_type(const ecl_smspec_type * smspec) {
 
 static void ecl_smspec_fortio_fwrite( const ecl_smspec_type * smspec , fortio_type * fortio) {
   int num_nodes           = ecl_smspec_num_nodes( smspec );
+  {
```
pgdr: are these necessary?

joakim-hove:
> I would prefer the guards removed. 

I guess you mean the `{ ... }` scoping? I have removed it - extracting the content to a function.

> it would be better with an additional non-statoil-test that has something other than metric.

The reason we did not have a proper such test is that we do not - yet - have proper support for writing anything different than METRIC. I have monkey patched something now.

PR #359, opened 2018-03-19
util_same_file: in WINDOWS there is only check of pathname.

stefoss23:
**Task**
util_same_file does not work properly in WINDOWS,

**Approach**
Check in util_same_file for UNIX. IF not, the pathname is checked by name comparison. 

**Pre un-WIP checklist**

### DIFF
--- a/lib/util/util.c
+++ b/lib/util/util.c
@@ -2702,6 +2702,7 @@ bool util_ftruncate(FILE * stream , long size) {
 */
 
 bool util_same_file(const char * file1 , const char * file2) {
+#ifdef ERT_HAVE_UNISTD
   stat_type buffer1 , buffer2;
   int stat1,stat2;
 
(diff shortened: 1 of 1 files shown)


stefoss23:
@magnesj Can you please test this.

lib/util/util.c
```diff
@@ -2715,6 +2716,20 @@ bool util_same_file(const char * file1 , const char * file2) {
       return false;
   } else
     return false;  // Files which do not exist are no equal!
+#else
+  if (util_file_exists(file1) & util_file_exists(file2)) {
+    char * abs_path1 = util_alloc_abs_path(file1);
+    char * abs_path2 = util_alloc_abs_path(file2);
+    bool check = false;
```
joakim-hove: I suggest instead:
```C
bool same_file = util_string_equal(abs_path1, abs_path2);
...
return same_file;
```

stefoss23: OK

lib/util/util.c
```diff
@@ -2715,6 +2716,20 @@ bool util_same_file(const char * file1 , const char * file2) {
       return false;
   } else
     return false;  // Files which do not exist are no equal!
+#else
+  if (util_file_exists(file1) & util_file_exists(file2)) {
```
joakim-hove: NB: This should be logical and - i.e. `&&`; you are using a bitwise and.

lib/util/util.c
```diff
@@ -2702,6 +2702,7 @@ bool util_ftruncate(FILE * stream , long size) {
 */
 
 bool util_same_file(const char * file1 , const char * file2) {
+#ifdef HAVE_UNISTD_H
```
joakim-hove: This: https://github.com/Statoil/libecl/blob/5748593069787de71e2d2ebbb0832efb0320676d/lib/include/ert/util/util.h#L377 is the symbol you should use.

magnesj:
I will ask @bjornjensen to do a review.

bjornjensen:
Tested on Windows. It works!

PR #369, opened 2018-04-04
Add case name for 'anonymous' restart writer

pgdr:
When generating "restart writers" and "writers" for ecl sum, ecl could crash (`repr`) due to missing case name.  Add simple case name.

Also add an explicit type checking and raise TypeError if wrong type.  This is only to get a better error message than the one ctypes raises.


pgdr:
Second commit adds `repr`:
```
EclSumTStep(sim_days=30.0, sim_time=2015-01-31 00:00:00, report=155, ministep=155) at 0x18c7470
```

PR #374, opened 2018-04-11
Windows transform restart kw

joakim-hove:
Tranform '/' characters to '\' in the embedded restart path when loading summay files on Windows.


bjornjensen:
Tested.
Result: success!

PR #381, opened 2018-04-19
Ecl to cpp

stefoss23:
**Task**
As part of converting libecl to c++, ecl files are set to c++ standard. 

**Approach**
Usually .c is changed to .cpp, but also some syntactical changes. 

**Pre un-WIP checklist**


joakim-hove:
This is cool!

Issue #385, opened 2018-04-24
build fails on 32 bit systems


andlaus:
on my 32-bit ARM system for testing (featuring a debian jessie based distribution with GCC 4.9.2), I get the following compiler error:

```
/home/and/src/libecl/lib/util/buffer.cpp: In function ‘size_t buffer_fwrite_compressed(buffer_type*, const void*, size_t)’:
/home/and/src/libecl/lib/util/buffer.cpp:905:94: error: invalid conversion from ‘size_t* {aka unsigned int*}’ to ‘long unsigned int*’ [-fpermissive]
         util_compress_buffer( ptr , byte_size , &buffer->data[buffer->pos] , &compressed_size);
```

which is correct because `long unsigned int` is only equivalent to `size_t` on 64 bit systems.

joakim-hove:
> on my 32-bit ARM system for testing (featuring a debian jessie based distribution with GCC 4.9.2), I get the following compiler error:

OK - thank you; will look into it.

joakim-hove:
I don't have access to 32bit anywhere, cool if you can give this: a spin: https://github.com/Statoil/libecl/pull/386

Removing the zlib stuff has been on the agenda for some time; I guess that should be ramped up.

andlaus:
> I don't have access to 32bit anywhere, cool if you can give this: a spin: #386

On a x86-64 system you can e.g. use an x86-32 version of ubuntu 16.04 in a virtual machine.

andlaus:
#387 does the trick. thanks for fixing this quickly.

PR #387, opened 2018-04-25
Use unsgined long instead of size_t

joakim-hove:


pgdr:
Can you specify in the description what the problem was and the reasoning behind the fix?

joakim-hove:
> Can you specify in the description what the problem was and the reasoning behind the fix?

After switching the files under `util/` to C++ there was a compilation error on 32 bit: #385. The problem has probably been there for many years, but the C compiler has not seen it / complained.

PR #391, opened 2018-04-26
Ecl sum numpy vector

joakim-hove:
**Task**
Export numpy vector from ecl_sum

### DIFF
--- a/lib/include/ert/ecl/ecl_sum_data.h
+++ b/lib/include/ert/ecl/ecl_sum_data.h
@@ -24,6 +24,7 @@
 extern "C" {
 #endif
 #include <stdlib.h>
+#include <stdint.h>
 #include <time.h>
 
 #include <ert/util/time_t_vector.h>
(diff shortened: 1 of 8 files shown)


python/CMakeLists.txt
```diff
@@ -19,6 +19,13 @@ if (NOT DEFINED PY_cwrap)
    return()
 endif()
 
+find_python_package(pandas 1 ${PYTHON_INSTALL_PREFIX})
+if (NOT DEFINED PY_cwrap)
+   message(WARNING "Pandas module not found Python wrappers not enabled. Install with: \"pip install pandas\"")
+   set( ENABLE_PYTHON OFF PARENT_SCOPE )
+   return()
+endif()
+
```
pgdr: PY_cwrap → PY_pandas

lib/include/ert/ecl/ecl_sum_data.h
```diff
@@ -87,9 +88,22 @@ typedef struct ecl_sum_data_struct ecl_sum_data_type ;
...
+  double_vector_type     * ecl_sum_data_alloc_seconds_solution( const ecl_sum_data_type * data , const smspec_node_type * node , double value, bool rates_clamp_lower);
+  void                     ecl_sum_data_init_double_frame_interp(const ecl_sum_data_type * data,
+                                                                 const ecl_sum_vector_type * keywords,
+                                                                 const time_t_vector_type * time_points,
+                                                                 double * output_data);
+
+  void ecl_sum_data_init_double_vector_interp(const ecl_sum_data_type * data,
+                                              const smspec_node_type * smspec_node,
+                                              const time_t_vector_type * time_points,
+                                              double * output_data);
```
pgdr: That's _a lot_ of new symbols ... 15 new header symbols, are they necessary?

pgdr:
There is no description, and no comments nor documentation, so it's somewhat difficult to get a grasp of what is really the new feature here.

Why not do it completely from Python to begin with?  Is that too slow?

python/CMakeLists.txt
joakim-hove: Good catch.

lib/include/ert/ecl/ecl_sum_data.h
joakim-hove: Ehhh - 16 actually; well - to be able to initialize the numpy/pandas datastructures; yes they are necessary. Indeed some of them could have been in an "internal header" - but there is no such system yet.

joakim-hove:
> There is no description, and no comments nor documentation, 

The `EclSum.numpy_vector()` function is documented, but I agree also the `EclSum.pandas_frame() function should be documented.

> Why not do it completely from Python to begin with? Is that too slow?

Well; we need a way to get data from C to Python; and I am quite enthusiastic about this PR because I think the functions from this PR should become *the way* to do it, and that several other functions can be removed/deprecated. Among other things I think adding a `time_index` argument is very beneficial, then lots the report_step related functionality can be removed.

This was also the reason I was so keen on deprecation.

Issue #395, opened 2018-05-01
Crash when loading well information


joakim-hove:
From Resinsight Contact person Håkon Høgstøl in Statoil:

```
Hei
 
Det er et problem med krasj i ResInsight (well_conn_alloc_from_kw) ved import av en del  Peregrino 
modeller.  Kan du se på feilen? En mulig relevant observasjon er at krasj kommer fra Eclise 2017.1 
kjøringer, mens noen nesten like modeller fra Eclipse 2017.2 er ok.
 
Jeg har kopiert over et case som krasjer her:
Unix_tr:  /project/multiscale/OPM/Ceetron/simu/BUG_cases/2018.01.1_peregrino_wellcon/ R6_HM2016B_FFP_BASE.EGRID
 
 
Kopi av feilmelding:
 
rio-linrgs003:/project/peregrino/ressim/sector/main/2017b/ver001/e100/simu/amap/BNEUC64:$ ResInsight(13388)/ findLibraryInternal: plugins should not have a 'lib' prefix: "libkfilemodule.so"
ResInsight(13388)/ KPluginLoader::load: The plugin "libkfilemodule" doesn't contain a kde_plugin_verification_data structure                                                                
 
Error message: well_conn_alloc_from_kw: icon direction value:1369277249 not recognized
 
See file: /tmp/ert_abort_dump.hhgs.20180426-092917.log for more details of the crash.
Setting the environment variable "ERT_SHOW_BACKTRACE" will show the backtrace on stderr.
 
rio-linrgs003:/project/peregrino/ressim/sector/main/2017b/ver001/e100/simu/amap/BNEUC64:$ more /tmp/ert_abort_dump.hhgs.20180426-092917.log
 
 
Abort called from: well_conn_alloc_from_kw (/home/jenkinstr/workspace/ResInsight-Master-RH6-Octave3.8.1-odb/ThirdParty/Ert/lib/ecl/well_conn.c:216)
 
Error message: well_conn_alloc_from_kw: icon direction value:1369277249 not recognized
 
 
 
--------------------------------------------------------------------------------
#00 util_abort__(..)                                                              in ??:0
#01 well_conn_alloc_from_kw(..)                                                   in ??:0
#02 well_conn_collection_load_from_kw(..)                                         in ??:0
#03 ???(..)                                                                       in well_state.c:0
#04 well_state_add_connections2(..)                                               in ??:0
#05 well_state_alloc_from_file2(..)                                               in ??:0
[...]

PR #396, opened 2018-05-02
Icon invalid value

joakim-hove:
See: #395

### DIFF
--- a/lib/ecl/well_conn.cpp
+++ b/lib/ecl/well_conn.cpp
@@ -142,9 +142,6 @@ well_conn_type * well_conn_alloc_fracture_MSW( int i , int j , int k , double co
   return well_conn_alloc__(i , j , k , connection_factor , dir , open , segment_id , false);
 }
 
-
-
-
 /*
   Observe that the (ijk) and branch values are shifted to zero offset to be
   aligned with the rest of the ert libraries.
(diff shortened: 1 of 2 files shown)


lib/ecl/well_conn.cpp
```diff
@@ -157,73 +154,85 @@ well_conn_type * well_conn_alloc_from_kw( const ecl_kw_type * icon_kw ,
...
+  well_conn_dir_enum dir = well_conn_fracX;
+
+  /* Set the status */
+  {
+    int int_status = ecl_kw_iget_int( icon_kw , icon_offset + ICON_STATUS_INDEX );
+    if (int_status > 0)
+      open = true;
+    else
+      open = false;
+  }
```
pgdr: These lines could perhaps be replaced with
```c
bool open = ecl_kw_iget_int(icon_kw, icon_offset + ICON_STATUS_INDEX) > 0
```

lib/ecl/well_conn.cpp
```diff
@@ -157,73 +154,85 @@ well_conn_type * well_conn_alloc_from_kw( const ecl_kw_type * icon_kw ,
...
+    break;
+  case(ICON_FRACX):
+    dir = well_conn_fracX;
+    break;
+  case(ICON_FRACY):
+    dir = well_conn_fracY;
+    break;
+  default:
+    util_abort("%s: icon direction value:%d not recognized\n",__func__ , int_direction);
+  }
```
pgdr: You can remove `default` now if we have the outside range covered above?

lib/ecl/well_conn.cpp
pgdr: Perhaps we should rename `open` to something else

joakim-hove: Have replaced the `open` assignment with your suggestion. The name `open` is quite descriptive in my opinion - so I have left that.

lib/ecl/well_conn.cpp
joakim-hove: Done.

joakim-hove:
Thank you for the review

lib/ecl/well_conn.cpp
pgdr: The point wasn't that it wasn't descriptive, but I thought maybe it wasn't considered in good taste to reuse standard library names.

joakim-hove: OK - I agree; have changed to `is_open`

PR #403, opened 2018-05-13
Removing C++ warnings

stefoss23:
**Task**
Removing warnings due to c->c++ conversions.

**Approach**

**Pre un-WIP checklist**


joakim-hove:
The cmake commit has probably sneeked in from here: #399  - and should be removed/fixed; otherwise this is ready to be merged.

stefoss23:
No, there are more commits.

PR #408, opened 2018-05-16
Build shall fail if CXX warnings.

stefoss23:
**Task**
Build shall fail if CXX warnings.

**Approach**
Changed libecl code to remove warnings.
Different cmake setup defined in build_total:
- -DERT_USE_OPENMP=ON
- CXX flag: -Werror -Wno-unused-result


joakim-hove:
The same change was merged here: #409

PR #412, opened 2018-05-30
Visual Studio support for Python api

Galfodo:
**Task**
Implement support for building Python wrappers for Visual Studio 2015 and newer

**Approach**
Check for MSVC and MSVC_VERSION in cmake script, rename ecl.dll -> libecl.dll, make dll export all symbols.


joakim-hove:
Cool! This is a very welcome addition, and will certainly be merged. Some minor things:

1. The symbol export is based on: ` set_target_properties(ecl PROPERTIES WINDOWS_EXPORT_ALL_SYMBOLS ON)` - that is a feature from cmake version 3.4, the project as a whole has a minimun cmake version requirement of 2.8.12 - would lke to keep it at that. Fortunately the `WINDOWS_EXPORT_ALL_SYMBOLS` does not seem to create any problems on a linux build with cmake 2.8.12.

Would be good if you just added a comment at the top of the top-level `CMakeLists.txt` saying that: "Iff you are creating Python wrappers on Windows the actual version requirement is actually 3.4".

2. I guess you have tested this? The main developers and users are on Linux, so Windows support is unfortunately the neglected step-son; in particular that implies that the automatic testing on Appveyor does not run tests on Windows ;-(  Fixing that is a large task, but I guess we should strive to include some - testing also on Windows.

For now it would be good if you could pass the `-DENABLE_PYTHON=ON` switch to appveyor build script.

Then it is merge time!

lib/util/TestArea.cpp
```diff
@@ -15,6 +15,9 @@
    See the GNU General Public License at <http://www.gnu.org/licenses/gpl.html>
    for more details.
 */
+
+#ifndef _MSC_VER // Currently not in Visual Studio build
```
joakim-hove: What is the purpose of this? Does the `TestArea.cpp` not compile with MSVC?

Galfodo: I got linker errors when building TestArea.cpp ( undefined symbol test_area_* ), but these went away when turning off ERT_BUILD_CXX, so I will revert these changes.

Galfodo:
1. I will add the comment as per your suggestion. The "better" way of exporting symbols is to decorate each function with an export macro, but maybe that ship has sailed.

2. I have not tested it systematically, but I have managed to import some grids with property time series (.egrid, .init, .unrst) and visualize them. I get crashes when importing .grdecl files, but I will look into that.

lib/util/TestArea.cpp
joakim-hove: Good!

joakim-hove:
> The "better" way of exporting symbols is to decorate each function with an export macro, but maybe that ship has sailed.

Ehh - yes; that ship has left the harbour, but yes I agree that the API is way to large and sprinkling explicit export symbols throughout the code might have been healthy. Will not be now though!

> have not tested it systematically, but I have managed to import some grids with property time series (.egrid, .init, .unrst) and visualize them. 

That is good enough - then it essentially holds together.

> I get crashes when importing .grdecl files, but I will look into that. 

OK - good, the `.grdecl` files are extremely weakly structured - so that parsing is a bit guess and run.

> When turning on BUILD_TESTS, I get lots of linker errors ( undefined symbol test_work_area_*) + missing header file unistd.h, which MSVC does not have.

Yes - as I already mentioned Windows is untested territory; unfortunately.

joakim-hove:
Now - we *really* must get pip going!

PR #418, opened 2018-06-08
Grdecl double

joakim-hove:
Add overload using double* all the way for grdecl grid creation


atgeirr:
I believe this is soon ready to be merged from my point of view (just waiting for Jenkins in the opm-common PR). What more is required on this side?

joakim-hove:
> What more is required on this side?

It is ready - I'll merge it.

PR #420, opened 2018-06-12
Created unsmry_loader for LARGE UNSMRY files.

stefoss23:
**Task**
Create a loader to store and return data in std::vector<float> format for very large UNSMRY files. 

**Approach**
A class unsmry_loader has been made. It reads files using: ecl_file_view_index_fload_kw, again using fortio_data_fseek. The entire data file is not stored in memory as used in default reading. 

**Pre un-WIP checklist**


joakim-hove:
Continued here #438

PR #421, opened 2018-06-14
ecl3 fortio

jokva:
A proposal for a new approach to fortio, and base all the old fortio functionality on this new implementation.

The e3 (amalgamation of ecl3 and ecl, i18n style) directory serves as the for-now private, but to-be-public headers.

This work *should* not intefere with any other ecl component.

### DIFF
--- a/lib/e3/ecl/fortio.h
+++ b/lib/e3/ecl/fortio.h
@@ -1,0 +1,183 @@
+#ifndef ECL_FORTIO_H
+#define ECL_FORTIO_H
+
+#include <stdio.h>
+#include <stdint.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/*
+ * As per the gnu fortran manual, int32 is sufficient for the record byte
+ * marker. optionally we could support 8-byte markers with either compile-time
+ * configuration or a run-time switch
+ *
+ * http://gcc.gnu.org/onlinedocs/gfortran/File-format-of-unformatted-sequential-files.html
+ *
+ * By default, all functions assume strict fortran compatibility (i.e. with
+ * trailing record size) and network (big-endian) byte order.
+ *
+ *
+ * A Fortran program writes unformatted data to file in a statemente like:
+ *
+ *    integer array(100)
+ *    write(unit) array
+ *
+ * it actually writes a head and tail in addition to the actual
+ * data. The header and tail is a 4 byte integer, which value is the
+ * number of bytes in the immediately following record. I.e. what is
+ * actually found on disk after the Fortran code above is:
+ *
+ *   | 400 | array ...... | 400 |
+ *
+ */
+
+/*
+ * The ecl_fio functions are exception safe, that is, if a function fails, the
+ * file pointer is rewinded to before the function was called, and output
+ * parameters are not modified, as if the function was never called.
+ *
+ * This comes with a few exceptions:
+ * 1. if ECL_ERR_SEEK is returned, the roll-back of the file pointer itself
+ *    failed and NOTHING IS GUARANTEED. The file stream is left in an unspecified
+ *    state, and must be recovered accordingly.
+ * 2. in eclfio_get, the output record buffer must always be considered dirty
+ *    and incomplete unless the function suceeds, or ECL_EINVAL is returned.
+ *
+ *
+ * ECL_ERR_SEEK should be rather rare, but to provide strong guarantees, this
+ * error must be handled carefully.
+ */
+
+/*
+ * every function takes a const char* opts parameter. This is a tiny
+ * configuration language inspired by printf and fopen. every character not in
+ * the set of keys is ignored. the opts parameter must be null terminated.
+ *
+ * if two options setting the same parameter (e.g. i and f, or e and E), the
+ * last one in the option string takes effect.
+ *
+ * options
+ * -------
+ *  record data types:
+ *      c - characters, sizeof(char)
+ *      i - (signed)integers, sizeof(int32_t), default
+ *      f - single-precision float, sizeof(float)
+ *      d - double-precision float, sizeof(double)
+ *
+ * behaviour:
+ *      E - assume big-endian record data (default)
+ *      e - assume little-endian record data
+ *      t - transform/byteswap data according to data type (default)
+ *      T - don't transform/byteswap data (does not affect heads/tails)
+ *
+ * endianness parameter applies to both head, tail, and data, but head/tail can
+ * be interepreted with endianness byteswapping data by disabling transform
+ *
+ * fault tolerance:
+ *      # - ignore size hint
+ *      ~ - force no-tail (assume only head)
+ *      $ - allow no-tail (don't fail on missing tail)
+ */
+
+/*
+ * Get the size (number of elements) of the current record. The file position
+ * is approperiately rewinded afterwards, as if the function was never called.
+ *
+ * If this function fails, out is not modified.
+ *
+ * If the read fails, ECL_ERR_READ is returned.
+ *
+ * This function is largely intended for peeking the size of the next record,
+ * to approperiately allocate a large enough buffer, which is useful when
+ * dealing with unknown files. If it is know in advance how large the records
+ * are, it is not necessary to call this function before reading a record.
+ */
+int eclfio_sizeof( FILE*, const char* opts, int32_t* out );
+
+/*
+ * Advance the file position n records. The file position is reset if the
+ * function fails, as if the function was never called.
+ *
+ * Returns ECL_OK if all records were skipped. If it fails, either
+ * ECL_INVALID_RECORD or ECL_ERR_READ is returned, depending on the source of
+ * the error, same rules as that of eclfio_get.
+ *
+ * This function does not distinguish seek errors for any n not +-1, so to
+ * figure out which record fails, one record at a time must be skipped.
+ */
+int eclfio_skip( FILE*, const char* opts, int n );
+
+/*
+ * Get the next record, and its number of elements.
+ *
+ * The record buffer is generally assumed to be of approperiate size, which can
+ * be queried with eclfio_sizeof.
+ *
+ * On success, the value of recordsize denotes the number of elements read,
+ * whose size is determined by the "cifd" options. It is generally assumed that
+ * recordsize upon calling this function contains the size of the record
+ * buffer, as a failsafe mechanism - if a record is larger than this value, the
+ * read will be aborted and the file position rolled back. To opt out of this
+ * check, add # to opts.
+ *
+ * Both recordsize and record can be NULL, in which case the number of elements
+ * read is not returned, and no data is returned respectively. This allows
+ * precise reporting on how many elements each skipped records contains.
+ *
+ * If the elementsize is larger than 1, and transformation has not been
+ * explicitly disabled, endianness will be converted appropriately.
+ *
+ * It is assumed that all record has an appropriate head and tail. If it is
+ * know that no record has a tail, force this by passing ~ in opts. However, if
+ * it is uncertain if all records has tails, or it's alternating between tail
+ * and no-tail, the $ option tries to recover from missing tails by assuming
+ * the current position is the start of the next record.
+ *
+ * The contents of record* is unspecified in case of read failures, and may not
+ * be relied upon. If the function returns ECL_EINVAL, the output record is
+ * untouched.
+ *
+ * This function returns ECL_OK upon success, ECL_ERR_READ in case of read- or
+ * seek errors, ECL_INVALID_RECORD if either the record tail is broken and
+ * options is set accordingly. The list of error codes is not exhaustive, and
+ * robust code should have fallthrough error handling cases.
+ */
+int eclfio_get( FILE*, const char* opts, int32_t* recordsize, void* record );
+
+/*
+ * Put a record of nmemb elements
+ *
+ * This function will write both head and tail, unless tail writing is
+ * explicitly disabled with ~. If (nmemb * elemsize) overflows int32, the write
+ * is aborted and ECL_EINVAL is returned.
+ *
+ * put largely follows the same rules as get, including those of endianness.
+ * The file pointer is rolled back if any part of the function should fail, as
+ * if the function was never called.
+ *
+ * If a write fails after partial writes, no attempts are made to roll back
+ * written changes.
+ *
+ * Returns ECL_OK on success, or ECL_ERR_WRITE on failure. If ECL_ERR_SEEK is
+ * returned, the integrity of the file stream can not be guaranteed, and its
+ * state is considered unspecified.
+ */
+int eclfio_put( FILE*, const char* opts, int nmemb, const void* );
+
+enum ecl_errno {
+    ECL_OK = 0,
+    ECL_ERR_UNKNOWN,
+    ECL_ERR_SEEK,
+    ECL_ERR_READ,
+    ECL_ERR_WRITE,
+    ECL_INVALID_RECORD,
+    ECL_EINVAL,
+};
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif //ECL_FORTIO_H
--- a/lib/ecl/fortio.c
+++ b/lib/ecl/fortio.c
@@ -26,6 +26,7 @@
 #include <ert/util/type_macros.h>
 #include <ert/ecl/fortio.h>
 
+#include <ecl/fortio.h>
 
 #define FORTIO_ID  345116
 
--- a/lib/ecl/fortio.c
+++ b/lib/ecl/fortio.c
@@ -434,17 +434,11 @@ bool fortio_is_fortio_file(fortio_type * fortio) {
 */
 
 int fortio_init_read(fortio_type *fortio) {
-  int elm_read;
-  int record_size;
-
-  elm_read = fread(&record_size , sizeof(record_size) , 1 , fortio->stream);
-  if (elm_read == 1) {
-    if (fortio->endian_flip_header)
-      util_endian_flip_vector(&record_size , sizeof record_size , 1);
-
-    return record_size;
-  } else
-    return -1;
+    int32_t record_size;
+    int err = eclfio_sizeof( fortio->stream, fortio->opts, &record_size );
+    // this function exposes successful reads as advanced file pointers
+    if( err == 0 ) fseek( fortio->stream, 4, SEEK_CUR );
+    return err ? -1 : record_size;
 }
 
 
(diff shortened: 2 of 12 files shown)


jokva:
I hope (and think) the vision of this gets clearer once I've written some python extension code, or maybe some new code that uses fortio (instead of just calling into the old fortio with the new backend).

lib/src/fortio.cpp
```diff
@@ -0,0 +1,337 @@
...
+    fguard guard( fp );
+
+    std::int32_t head;
+    auto read = std::fread( &head, sizeof( head ), 1, fp );
+    if( read != 1 ) return ECL_ERR_READ;
+
+    if( o.bigendian )
+        head = (std::int32_t)ntohl( (std::uint32_t)head );
+
+    if( head < 0 ) return ECL_INVALID_RECORD;
```
jokva: This needs to check for int32 overflow too.

joakim-hove:
This is beautiful!

lib/src/fortio.cpp
joakim-hove: > This needs to check for int32 overflow too.

OK; is the head and tail unsigned in the Fortran standard?

joakim-hove:
> I hope (and think) the vision of this gets clearer once I've written some python extension code, or maybe some new code that uses fortio (instead of just calling into the old fortio with the new backend).

I think this looks very nice - and I am looking forward to getting it in. Python extensions, and also code using the new fortio functions seems reasonably clear to me. What I do not manage to follow/see entirely right now is how to handle the data types like `ecl_sum` and `ecl_grid`which hold on to quite a lot of state; but I am confident that will also find it's solution.

It is inspiring to see the plain C interface with a C++ implementation fully in an anonymous namespace.

jokva:
> What I do not manage to follow/see entirely right now is how to handle the data types like ecl_sum and ecl_gridwhich hold on to quite a lot of state; but I am confident that will also find it's solution.

Honestly, I have no clue yet either, but I'm sure we'll figure it out. What's cool is that if it ends up being hidden behind a pointer type, we can use a proper class, no problem.

lib/src/fortio.cpp
```diff
@@ -0,0 +1,337 @@
...
+
+    return ECL_ERR_UNKNOWN;
+} catch( std::runtime_error& ) {
+    return ECL_ERR_SEEK;
+}
+
+int eclfio_put( std::FILE* fp,
+                const char* opts,
+                int nmemb,
+                const void* record ) try {
```
joakim-hove: What is the runtime cost of this `try ... catch`?

lib/src/fortio.cpp
```diff
@@ -0,0 +1,337 @@
...
+    if(   std::numeric_limits< std::int32_t >::max()
+        < std::int64_t(nmemb) * std::int64_t(o.elemsize) ) {
+        /*
+         * a huge record (>2G) is being requested, but the 4-byte headers don't
+         * allow that. consider this an invalid argument and bail
+         */
+        return ECL_EINVAL;
+    }
+
+    fguard guard( fp );
```
joakim-hove: Could we scope the guard variable:
```C++
{
      fguard guard(fp);
      ....
      ....
}
return ECL_OK;
```
then we can let the destructor play out normally - without the explcit `guard = nullptr` assignment?

lib/src/fortio.cpp
jokva: (virtually) zero, unless the rewind fails, in which case you pay for stack unwinding. But at that point we're probably in such a dire mess that any runtime cost is irrelevant.

lib/src/fortio.cpp
jokva: Well, you need to signal the guard that we're exiting on success, otherwise it'll attempt to roll back.

What *could* be done is a trivial method that reads return value:

```
int exit( int errc ) { if( errc == ECL_OK ) this->fp = nullptr; return errc; }

[...]

if( head == tail ) {
    return guard.exit( ECL_OK );
}
```

But that's only slightly more ergonomic.

joakim-hove: Fair enough - I realized after I wrote that I misunderstood the mechanism slightly.

lib/src/fortio.cpp
```diff
@@ -0,0 +1,337 @@
...
+            throw std::runtime_error("");
+        }
+    }
+
+    ~fguard() noexcept( false ) {
+        if( !this->fp ) return;
+
+        const auto err = std::fsetpos( fp, &this->pos );
+        if( err ) throw std::runtime_error("");
+    }
```
andreabrambilla: I know this is a controversial topic, but I wouldn't `throw` from a destructor 
Specifically in this case, if you are unable to seek, it means something went awfully wrong before you tried to seek. And doing it like this will actually mask the real error.

Take for instance `eclfio_sizeof`. What can fail there is the `std::fread`, and if that fails, the method is supposed to return `ECL_ERR_READ`. But if `std::fread` failed so badly that the following seek fails as well, then the method returns `ECL_ERR_SEEK`. So there is a discrepancy between the error reported and what error actually occurred

jokva: I disagree - if you're in a situation where the *rollback* fails, literally nothing can be guaranteed anymore, and it doesn't matter if it was the read that failed, or an argument was invalid, because at that point callers must take *extra* care to restore a well-known state.

Maybe the return type of ECL_ERR_SEEK is a red herring, because it's not simply a seek that's failed probably, it's the exception guarantee to boot. It's pretty disastrous (and rare, I would imagine). But I would still argue that I don't if the read fail if that also means that my seeking is busted.

lib/ecl/fortio.c
```diff
@@ -451,41 +432,24 @@ bool fortio_is_fortio_file(fortio_type * fortio) {
...
     }
-    {
-      int block_index = data_element / block_size;
-      int headers = (block_index + 1) * 4;
-      int trailers = block_index * 4;
-      offset_type bytes_to_skip = data_offset + headers + trailers + (data_element * element_size);
 
-      fortio_fseek(fortio, bytes_to_skip, SEEK_SET);
-    }
+    eclfio_skip( fortio->stream, fortio->opts, data_element / block_size );
```
joakim-hove: A chunk of data is generally written in a set of record, where each record has a header and trailer. The blocksizes are quite small: https://github.com/Statoil/libecl/blob/master/lib/ecl/ecl_kw.cpp#L63 - so a large chunk of data is typically split over many blocks. The current master version of the code takes these blocks, and the headers and trailers into account, the current `eclfio_skip()` call is too simple.

Open question: should the low level `eclfio` layer provide functionality to completely hide the fact that data is (typically) split over several blocks? I would vote carefully yes.

jokva: Hm, so eclipse effectively maintains a maximum size of a record (in the fortran sense) of 4kb, and that longer arrays are just concatenated records?

I would argue that eclfio should in that case not even care, that's it's just a simple I/O layer, and it makes sense for it to be agnostic and tolerant when it comes to these sizes. The fact that records can or will be longer is easily maintained in a higher layer.

joakim-hove: Eclipse maintains a maximum record size - which is quite small, exactly how small depends on the datatype in question.

jokva: I would argue that gives us even more reason to make fortio ignorant of it, and let other layers deal with it.

For reading it doesn't matter, because a shorter record and a block would be indistinguishable. Which at that point makes the interface asymmetric (`put` knows about a block size, but `get` does not). There's not too much harm in having another parameter that says block size, I guess, but it does seem a tad clumsy.

I'll think a bit about it, but thanks for the input on the bug, I'll fix it soon.

joakim-hove: > I would argue that gives us even more reason to make fortio ignorant of it, and let other layers deal with it.

It is not 100% clear to me what you mean here, personally I see conflicting arguments pointing in different directons. I guess it comes to whether this is *fortran* centric functionality/api *eclipse centric*.
 
**record foxus:** Letting `fortio` focus purelly on records makes the api clean and simple to write and maintain. The disadvantage is that as an eclipse centric consumer of this functionality you are not interested in the records - at all; you are interested in IO of full arrays - that is also the semantic level where you are interested in guarantees of atomicity. The records is a detail of eclipse implementation detail which is really not interesting.

A typical array size is O(10000) elements, so in the current implementation that will amount to 100 rounds of setup and teardown in the `put` and `get` methods; I fear that might incur runtime cost in a very inner loop.

2. Letting `fortio` work with full arrays, and also maintain the eclipse expectations of record sizes makes the api/implementation slightly less clean - and as you say potentially asymmetric. But it will be more focused on the actual use case - as I see it.

So - all in all I would vote for this api:
```C
int eclfio_gut(FILE*, const char*, int32 size, void * data);
int eclfio_put(FILE*, const char*, int32 nmemb, int32 record_size, void * data)
```
Where it is assumed that the record functiaonlity is handled fully as an implementation detail inside the `eclfio`implementation.


Finally it should be noted that "random-access" functionality to just read a value here and there is relevant, and becoming more so, the necessary offset calculations require awareness of the blocks.

lib/e3/ecl/fortio.h
```diff
@@ -0,0 +1,183 @@
...
+ * if the function was never called.
+ *
+ * If a write fails after partial writes, no attempts are made to roll back
+ * written changes.
+ *
+ * Returns ECL_OK on success, or ECL_ERR_WRITE on failure. If ECL_ERR_SEEK is
+ * returned, the integrity of the file stream can not be guaranteed, and its
+ * state is considered unspecified.
+ */
+int eclfio_put( FILE*, const char* opts, int nmemb, const void* );
```
joakim-hove: I agree that the `nmemb` is the natural parameter here, but just something to be aware of: Strings in eclipse can be stored in two different ways; the most common is 8 character arrays (with space padding):

```
WGNAMES = ["WELL1   ", "WELL_123", "GRPABC  "]
```
When passing this to current `eclfio_put` function the `nmemb` variable must be 24. I.e. there is a difference between `opts="d"` where the internal layer knows the size of the `double`, and the strings where outer layer must convert to a char representation - sizewise - not that I think there is an alternative though.   

In addition to the `'CHAR'` datatype which amounts to fixed 8 character strings there is also support for `Cnnn` strings which are fixed size, space padded, strings of length `nnn`.

joakim-hove:
The keyword test failing is this:
```python
str_len = 410
kw = EclKW("TEST_KW", 10, EclDataType.ECL_STRING(str_len))
for i in range(10):
     kw[i] = str(i)

file_name = "ecl_kw_test"
with openFortIO(file_name, mode=FortIO.WRITE_MODE) as fortio:
       kw.fwrite(fortio)
'
with openFortIO(file_name) as fortio:
       loaded_kw = EclKW.fread(fortio)

self.assertEqual(kw, loaded_kw)
```
i.e. essentially writing a keyword consisting of ten `Cnnn` strings to file and reading it back in again. The passes for `str_len <= 409` and fails for `str_len >= 410`; when reading back the first 4096 bytes are correct - to me this smells like problems with an IO buffer of size 4096 bytes. Problem might have been there all the time - just unearthed now?

jokva:
> i.e. essentially writing a keyword consisting of ten Cnnn strings to file and reading it back in again. The test passes for str_len <= 409 and fails for str_len >= 410; when reading back the first 4096 bytes are correct - to me this smells like problems with an IO buffer of size 4096 bytes. Problem might have been there all the time - just unearthed now?

No, a 4096 buffer error makes sense, it's wildly untested.

lib/ecl/fortio.c
jokva: > A typical array size is O(10000) elements, so in the current implementation that will amount to 100 rounds of setup and teardown in the put and get methods; I fear that might incur runtime cost in a very inner loop.

I wouldn't worry about that, as we're about to do IO (which will dominate everything), and it's no more expensive than doing fgetpos/fsetpos, which essentially is just a struct member access. It does add some potential for robustness on writing multiple records in writing an array, because you can retry when it fails on that particular write.

jokva: Anywho, I think the best approach would be to have one very fortran centric layer, that deals with endianness, and the fstream api directly, and build eclipse-centric functionality on top of that, which now has no need to deal with the C fstream api (at least no so much).

jokva: Yep, that's one immediate benefit. Maybe fortio will never be useful outside of `ecl`, and I'm honestly ok with that.

lib/src/fortio.cpp
```diff
@@ -0,0 +1,384 @@
...
+ * to/from host carefully swaps endianness for a large array.
+ *
+ * The switch guarantees that byte swap happens properly, even on platforms
+ * where sizeof(char) == sizeof(int32), without accidently terminating before,
+ * relying on htons being well-defined on such platforms. Should any non-char,
+ * int16, and int32 sizes be passed through, nothing should happen, but
+ * non-zero is returned.
+ */
+
+int to_host( void* dst, int size, std::int32_t nmemb ) {
```
joakim-hove: When adding these functions can you at the same the remove the endian swap functions from `util.c`

lib/src/fortio.cpp
```diff
@@ -0,0 +1,384 @@
...
+#include <ciso646>
+#include <cstdint>
+#include <cstdio>
+#include <cstring>
+#include <limits>
+#include <stdexcept>
+
+#include <ecl/fortio.h>
+
+#ifdef HAVE_NETINET_IN_H
```
joakim-hove: Depending a bit on how things are done; but I think it will be necessary to do an endian swapping of the integer element from the header block; that would require access to the `htonl()` function also in another compilation unit. Would be good if this `#ifdef ... #endif` block was assembled in a private header.

lib/e3/ecl/fortio.h
```diff
@@ -0,0 +1,183 @@
...
+#define ECL_FORTIO_H
+
+#include <stdio.h>
+#include <stdint.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/*
```
joakim-hove: It does not matter for some time - but I think it should be an open question whether the `fortio.h` should be publically exposed at all. I am leaning to no.

lib/src/fortio.cpp
jokva: I think we can afford to wait a bit longer, it's they're still (somewhat) in use.

```
ecl/ecl_kw.cpp:276:    util_endian_flip_vector(buffer, sizeof_iotype, ecl_kw->size);
ecl/ecl_kw.cpp:296:  util_endian_flip_vector(buffer, sizeof_iotype, ecl_kw->size);
ecl/ecl_kw.cpp:315:      util_endian_flip_vector(buffer, sizeof_iotype, ecl_kw->size);
ecl/ecl_kw.cpp:1277:        util_endian_flip_vector(io_buffer, sizeof_iotype, int_vector_size(index_map));
ecl/ecl_kw.cpp:1388:      util_endian_flip_vector(&size , sizeof size , 1);
ecl/ecl_kw.cpp:1728:      util_endian_flip_vector(&size , sizeof size , 1);

lib/rms/rms_tagkey.cpp:458:      util_endian_flip_vector(tagkey->data , tagkey->sizeof_ctype , tagkey->size);
```

lib/src/fortio.cpp
jokva: I think you're right, but it's a bit premature to re-introduce a config.h (or similar) with only one module implemented. Mind if we wait a bit?

lib/src/fortio.cpp
```diff
@@ -0,0 +1,384 @@
...
+    int elemsize    = sizeof( std::int32_t );
+
+    // transform by default unless record is nullptr and this record should be
+    // skipped
+    bool transform    = true;
+    bool force_notail = false;
+    bool allow_notail = false;
+};
+
+options parse_opts( const char* opts ) noexcept {
```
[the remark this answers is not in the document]
jokva: I feel it's a bit premature, I'd like to play a bit with the keyword design before locking into the same configuration style, but it's reasonable if fortio ends up being a proper implementation detail.

lib/src/fortio.cpp
```diff
@@ -0,0 +1,385 @@
...
+        case sizeof( std::int32_t ):
+            for( int i = 0; i < nmemb; ++i ) {
+                std::uint32_t x;
+                std::memcpy( &x, ptr, sizeof( x ) );
+                x = ntohl( x );
+                std::memcpy( ptr, &x, sizeof( x ) );
+                ptr += sizeof( x );
+            }
+            return 0;
+
```
joakim-hove: This needs an 8 byte entry as well.

lib/src/fortio.cpp
joakim-hove: OK - but as soon as possible I think we should avoid having two implementations; but that goes without saying I guess.

lib/src/fortio.cpp
joakim-hove: > Mind if we wait a bit?

Sure.

lib/src/fortio.cpp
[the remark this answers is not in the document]
joakim-hove: OK - let's see how it pans out; I am eagerly waiting.

lib/src/fortio.cpp
jokva: For sure, but flip-vector is already public and relied upon by res  (and other bits of ecl), and for starters it does no harm. I think we can afford to wait before we purge util, as long as we don't rely on it in e3.

lib/ecl/fortio.c
```diff
@@ -119,7 +119,7 @@ static fortio_type * fortio_alloc__(const char *filename , bool fmt_file , bool
 */
 
 static bool __read_int(FILE * stream , int * value, bool endian_flip) {
-    return !eclfio_sizeof( stream, endian_flip ? "" : "e", value );
+    return !eclfio_sizeof( stream, endian_flip ? "c" : "ce", value );
```
joakim-hove: I assume this should have the `"i"` type specifier, otherwise the endian swapping of the data will not be correct?

jokva: I thought so too (which is why the tests failed), turns out __read_int expects the size in bytes of the current record, not the number-of-integers, so it always needs `c`. `eclfio_sizeof` doesn't swap bytes except for the header, which is always assumed to be `int32`

joakim-hove:
Is this getting ready for merge?

lars-petter-hauge:
Yes.

jokva:
Actually hang on a minute, this is slightly out-of-sync.

jokva:
It should be ready this afternoon (at maybe 18-19)

jokva:
I think we're good now.

PR #423, opened 2018-06-21
Add CIT and CPT to total_vars in smspec_node

lars-petter-hauge:
**Task**
Added `CIT` and `CPT` to `smspec_node.cpp:334 total_vars`, otherwise a call with keyword e.g. CCIT to smspec_node_is_total would return false.

**Pre un-WIP checklist**
- [ x] Statoil tests pass locally


lars-petter-hauge:
I didn't add any more tests to this commit. The only place I found that any of these keys were tested is in `libecl/python/tests/ecl_tests/test_ecl_file_statoil.py`, but that relies on a specifically made SMSPEC file.

PR #424, opened 2018-06-21
Expose ecl_file_view_iget_distinct_kw in Python API

Galfodo:
**Task**
Expose ecl_file_view_iget_distinct_kw in Python API to be able to quickly enumerate distinct keywords in file.

**Approach**
Create ECLPrototype for the C API function


joakim-hove:
Thank you - I will merge this, but I would prefer if the python Api was something like:

```def unique_kw(self)
    # Assemble list of unique keywords or alternatively a generator.
```

PR #425, opened 2018-06-27
Refactor ecl_sum_data

stefoss23:
**Task**
Refactor ecl_sum_data to be compatible to use the new unsmry_loader. 

**Approach**
The data contained in ecl_sum_data is split in severl ecl_sum_file_data_type instances. 

**Pre un-WIP checklist**

### DIFF
--- a/lib/ecl/ecl_sum_data.cpp
+++ b/lib/ecl/ecl_sum_data.cpp
@@ -199,7 +199,7 @@
    method to access the time dimension of the data, i.e. all the
    functions like ecl_sum_get_general_var() expect the time direction
    to be given as a ministep; however it is also possible to get the
-   data by giving an invector_typeternal (not that internal ...) index. In
+   data by giving an internal (not that internal ...) index. In
    ecl_sum_data.c the latter functions have _iget():
 
 
@@ -860,8 +860,6 @@ double ecl_sum_data_iget( const ecl_sum_data_type * data , int time_index , int 
   ecl::ecl_sum_file_data_type * file_data = data->data_list[list_index];
   int * param_map = data->param_map[list_index];
   return file_data->iget( time_index - data->data_list_first_index[list_index] , param_map[params_index] );
-
-  return 0;
 }
 
 
--- a/lib/ecl/ecl_sum_data.cpp
+++ b/lib/ecl/ecl_sum_data.cpp
@@ -370,70 +370,6 @@ ecl_sum_data_type * ecl_sum_data_alloc_writer( ecl_smspec_type * smspec ) {
 }
 
 
-static void ecl_sum_data_fwrite_report__( const ecl_sum_data_type * data , int report_step , fortio_type * fortio) {
-  {
-    ecl_kw_type * seqhdr_kw = ecl_kw_alloc( SEQHDR_KW , SEQHDR_SIZE , ECL_INT );
-    ecl_kw_iset_int( seqhdr_kw , 0 , 0 );
-    ecl_kw_fwrite( seqhdr_kw , fortio );
-    ecl_kw_free( seqhdr_kw );
-  }
-
-  {
-    int index , index1 , index2;
-
-    ecl_sum_data_report2internal_range( data , report_step , &index1 , &index2);
-    for (index = index1; index <= index2; index++) {
-      const ecl_sum_tstep_type * tstep = ecl_sum_data_iget_ministep( data , index );
-      ecl_sum_tstep_fwrite( tstep , ecl_smspec_get_index_map( data->smspec ) , fortio );
-    }
-  }
-}
-
-
-
-static void ecl_sum_data_fwrite_multiple_step( const ecl_sum_data_type * data , const char * ecl_case , bool fmt_case , int report_step) {
-  char * filename = ecl_util_alloc_filename( NULL , ecl_case , ECL_UNIFIED_SUMMARY_FILE , fmt_case , 0 );
-  fortio_type * fortio = fortio_open_readwrite( filename , fmt_case , ECL_ENDIAN_FLIP );
-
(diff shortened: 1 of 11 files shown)


joakim-hove:
This is a large PR on a complex piece of code; it is not entirely trivial to review it either ... this will take some time. Some initial comments:

1. I am *extremely* concerned that the new implementation ecl_sum_file_data never becomes public; I would therefor prefer if the header file  ecl_sum_file_data.hpp was located in a new directory - e.g. "private/ecl_sum_file_data.hpp - and included as #include "private/ecl_sum_file_data.hpp".

2. I think the index_remapping should be part of the ecl_sum_data_file class; it should be possible to construct that mapping - possibly the trivial mapping - as part of the constructor.

3. I *hope* - but I am not certain, that it should be possible to also store the data_list_first_index and data_list_first_index information in the ecl_sum_file_data structure.

lib/include/ert/ecl/ecl_sum_file_data.hpp
```diff
@@ -0,0 +1,67 @@
...
+#include <ert/ecl/ecl_smspec.hpp>
+#include <ert/ecl/ecl_sum_tstep.hpp>
+#include <ert/ecl/ecl_file.hpp>
+
+namespace ecl {
+
+#define INVALID_MINISTEP_NR -1
+#define INVALID_TIME_T 0
+
+class ecl_sum_file_data_type {
```
joakim-hove: Remove the _type suffix - that is C code.

lib/include/ert/ecl/ecl_sum_file_data.hpp
```diff
@@ -0,0 +1,67 @@
...
+
+  double  days_start;
+  double  sim_length;
+  int first_report_step;
+  int last_report_step;
+  ecl_smspec_type * smspec;
```
joakim-hove: This pointer should be const

lib/include/ert/ecl/ecl_sum_file_data.hpp
```diff
@@ -0,0 +1,67 @@
...
+  void                 append_tstep__(ecl_sum_tstep_type * tstep);
+
+  void                 build_index();
+  bool                 has_report_step(int report_step ) const;
+  void                 report2internal_range(int report_step , int * index1 , int * index2 ) const;
+  void                 fwrite_report__( int report_step , fortio_type * fortio) const;
+  bool                 check_file( ecl_file_type * ecl_file );
+  void                 add_ecl_file(int report_step, const ecl_file_view_type * summary_view, const ecl_smspec_type * smspec);
+
+  public:
```
joakim-hove: Can you please reorder the content of the class so the public stuff comes first, and the private at the bottom.

joakim-hove:
My suggestion is that massage the ecl_sum_file_data class quite a lot first; possibly only merging that in the first merge?

lib/include/ert/ecl/ecl_sum_file_data.hpp
```diff
@@ -0,0 +1,67 @@
...
+class ecl_sum_file_data_type {
+
+  double  days_start;
+  double  sim_length;
+  int first_report_step;
+  int last_report_step;
+  ecl_smspec_type * smspec;
+  vector_type * data;
+  int_vector_type * report_first_index;
+  int_vector_type * report_last_index;
```
joakim-hove: Add a params_mapping as a private member here; I think that can be implemented as std::vector<int> instead of int *

lib/include/ert/ecl/ecl_sum_file_data.hpp
```diff
@@ -0,0 +1,67 @@
...
+
+class ecl_sum_file_data_type {
+
+  double  days_start;
+  double  sim_length;
+  int first_report_step;
+  int last_report_step;
+  ecl_smspec_type * smspec;
+  vector_type * data;
+  int_vector_type * report_first_index;
```
joakim-hove: Can these be  std::vector<int>?

lib/include/ert/ecl/ecl_sum_file_data.hpp
```diff
@@ -0,0 +1,67 @@
...
+  void                 report2internal_range(int report_step , int * index1 , int * index2 ) const;
+  void                 fwrite_report__( int report_step , fortio_type * fortio) const;
+  bool                 check_file( ecl_file_type * ecl_file );
+  void                 add_ecl_file(int report_step, const ecl_file_view_type * summary_view, const ecl_smspec_type * smspec);
+
+  public:
+    ecl_sum_file_data_type(ecl_smspec_type * smspec);
+    ~ecl_sum_file_data_type();
+
+    int                  get_length() const;
```
joakim-hove: Trying to be a bit less verbose - can you remove the leadeing get_?

joakim-hove:
Observe that the following fundamental assumption holds: "For each ecl_sum_file_data instance the start_date and the ecl_sum_file_data as a whole should be ordered in the same way." The advantage with this is that it should be possible to just add ecl_sum_file_data instances in one flat list, and then sort them afterwards - I think that will simplify the implementation somewhat.

lib/ecl/ecl_sum_data.cpp
```diff
@@ -205,12 +210,8 @@
 
 
 
-#define INVALID_MINISTEP_NR -1
-#define INVALID_TIME_T 0
-
 struct ecl_sum_data_struct {
   ecl_smspec_type        * smspec;                 /* A shared reference - only used for providing good error messages. */
-  vector_type            * data;                   /* Vector of ecl_sum_tstep_type instances. */
   double                   days_start;
```
joakim-hove: It seems several of these fields are now duplicated between this struct and the new  ecl_sum_file_data class?

joakim-hove:
Merged in #432

PR #432, opened 2018-07-04
Ecl sum data refactor

joakim-hove:
Eclipse summary data is split over two files: The `CASE.SMSPEC` file contains header information, and the actual values  along with all *time* related information is in the `CASE.UNSMRY`[1]. In libecl the data information goes to the `ecl_sum_data` structure.

When an eclipse simulation has been restarted the `CASE.SMSPEC` file can contain a pointer to the case we have restarted from, and this can continue in a long chain. Libecl will by default load all of these linked summary cases, and transparently present the complete history. Previously the implementation was based on "flattening" the timesteps from the different cases in the `ecl_sum_data` structure. With this PR the different cases retain their structure as individual cases, and each case is stored in a new `ecl_sum_file_data` structure. 

The purpose this refactoring is to facilitate on-demand loading of the results from summary results. That will come in a subsequent PR.

Continuation of: https://github.com/Statoil/libecl/pull/425

[1]: The `CASE.UNSMRY` file can alternatively be split in mulitple files `CASE.Snnnn` - that is a digression in the current context.


lib/ecl/ecl_sum_file_data.cpp
```diff
@@ -0,0 +1,736 @@
...
+  return this->time_range.second;
+}
+
+time_t ecl_sum_file_data::iget_sim_time(int time_index) const {
+  const ecl_sum_tstep_type * ministep = iget_ministep( time_index  );
+  return ecl_sum_tstep_get_sim_time(ministep);
+}
+
+double ecl_sum_file_data::iget_seconds(int time_index) const {
+  throw;
```
pgdr: Is this a mistake, or should it actually be like that?  If it's intended, please add a comment.

joakim-hove: Hmm - clearly mistake.

lib/ecl/ecl_sum_file_data.cpp
```diff
@@ -0,0 +1,736 @@
...
+          int internal_index = vector_get_size( data ) - 1;
+          this->report_map[report_step].second = internal_index;
+
+          sim_length = ecl_sum_tstep_get_sim_days( tstep );
+          this->time_range.second = ecl_sum_tstep_get_sim_time(tstep);
+
+          rebuild_index = false;
+        }
+      }
+    }
```
pgdr: I won't take your guard away from you if you really want it there, but I really think that the following is easier (fewer nested `if`s).

```c++
bool need_rebuild =
     prev_tstep == NULL
  || ecl_sum_tstep_get_report(prev_tstep) != ecl_sum_tstep_get_report(tstep))
  || ecl_sum_tstep_get_sim_days(prev_tstep) >= ecl_sum_tstep_get_sim_days(tstep);


if (need_rebuild) {
  int internal_index = vector_get_size( data ) - 1;
  this->report_map[report_step].second = internal_index;

  sim_length = ecl_sum_tstep_get_sim_days( tstep );
  this->time_range.second = ecl_sum_tstep_get_sim_time(tstep);

  rebuild_index = false;
}
```

lib/ecl/ecl_sum_file_data.cpp
```diff
@@ -0,0 +1,736 @@
...
+    */
+    days_start      = ecl_sum_tstep_get_sim_days( first_ministep );
+    sim_length      = ecl_sum_tstep_get_sim_days( last_ministep );
+    this->time_range = std::make_pair<time_t, time_t>(ecl_sum_tstep_get_sim_time( first_ministep ),
+                                                      ecl_sum_tstep_get_sim_time( last_ministep ));
+  }
+
+  /* Build up the report -> ministep mapping. */
+  {
+    int internal_index;
```
pgdr: This guard is not necessary if you put `internal_index` into the `for` loop.

lib/ecl/ecl_sum_file_data.cpp
```diff
@@ -0,0 +1,736 @@
...
+
+// **************************** Start Reading ************************************
+
+bool ecl_sum_file_data::check_file( ecl_file_type * ecl_file ) {
+  if (ecl_file_has_kw( ecl_file , PARAMS_KW ) &&
+      (ecl_file_get_num_named_kw( ecl_file , PARAMS_KW ) == ecl_file_get_num_named_kw( ecl_file , MINISTEP_KW)))
+    return true;
+  else
+    return false;
+}
```
pgdr: replace `it` with `return` and delete the rest.

lib/ecl/ecl_sum_file_data.cpp
```diff
@@ -0,0 +1,736 @@
...
+  }
+}
+
+
+bool ecl_sum_file_data::fread(const stringlist_type * filelist) {
+  if (stringlist_get_size( filelist ) == 0)
+    return false;
+
+  {
+    ecl_file_enum file_type = ecl_util_get_file_type( stringlist_iget( filelist , 0 ) , NULL , NULL);
```
pgdr: This guard is not necessary.

lib/ecl/ecl_sum_file_data.cpp
```diff
@@ -0,0 +1,736 @@
...
+    if ((stringlist_get_size( filelist ) > 1) && (file_type != ECL_SUMMARY_FILE))
+      util_abort("%s: internal error - when calling with more than one file - you can not supply a unified file - come on?! \n",__func__);
+
+    {
```
pgdr: This is one of these guards that immediately follows a brace-less `if` sentence which is really dangerous in my opinion.  It seems the scope below belongs to the `if`.

Furthermore, if you put the preceding `int filenr` into the `for` loop, there is nothing that leaks out anyway, so the guard can safely be deleted.

lib/ecl/ecl_sum_file_data.cpp
joakim-hove: I agree - that code was moved unchanged.

lib/ecl/ecl_sum_file_data.cpp
joakim-hove: Again - very old code.

lib/ecl/ecl_sum_file_data.cpp
```diff
@@ -0,0 +1,736 @@
...
+          this->time_range.second = ecl_sum_tstep_get_sim_time(tstep);
+
+          rebuild_index = false;
+        }
+      }
+    }
+    if (rebuild_index)
+      build_index();
+  }
+  //ecl_smspec_lock( ecl_smspec );
```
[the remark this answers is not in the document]
joakim-hove: Hmm - this was a bit more WIP than I thought.

joakim-hove:
OK - thank you for review; I have addressed most of the comments. Will have to look a bit more carefully on the lock issue.

PR #435, opened 2018-07-05
WIP Add eclfio_array_get/put

jokva:
These functions provide an abstraction for the eclipse detail of large
arrays being written across multiple smaller physical blocks.

In effect, it enables reading an eclipse record (the header
keyword-size-type, followed by the array) as:

    header = eclfio_get()
    nmemb = parse-header(header)
    return eclfio_array_get(nmemb)

Remaining work:

- tests!
- some considerations on return types
- port ecl2 to use this

I also propose we add a new data type in opts, `C` or `s`, for the fixed len-8 string. That means the `len` argument to the array function really only makes sense for var-strings (which are special) funky cool functionality where you emulate 64-bit ints by merging two consecutive int32. 

**Task**

**Approach**

**Pre un-WIP checklist**

### DIFF
--- a/lib/e3/ecl/fortio.h
+++ b/lib/e3/ecl/fortio.h
@@ -34,9 +34,10 @@ extern "C" {
  */
 
 /*
- * The ecl_fio functions are exception safe, that is, if a function fails, the
- * file pointer is rewinded to before the function was called, and output
- * parameters are not modified, as if the function was never called.
+ * The eclfio functions get, put, seek, and sizeof are exception safe, that is,
+ * if a function fails, the file pointer is rewinded to before the function was
+ * called, and output parameters are not modified, as if the function was never
+ * called.
  *
  * This comes with a few exceptions:
  * 1. if ECL_INCONSISTENT_STATE is returned, the roll-back of the file pointer
@@ -48,8 +49,11 @@ extern "C" {
  * ECL_INCONSISTENT_STATE should be rather rare, but to provide strong
  * guarantees, this error must be handled carefully.
  *
- * ECL_ERR_SEEK should be rather rare, but to provide strong guarantees, this
- * error must be handled carefully.
+ * The list of error codes is not exhaustive, and ecl reserves the right to
+ * both add new error codes, and redefine certain errors if more specific codes
+ * are added later. robust cost should always have fallthrough error handling.
+ *
+ * If a call succeeds, ECL_OK is returned and is always 0.
  */
 
 /*
@@ -93,12 +97,20 @@ extern "C" {
  *
  * If this function fails, out is not modified.
  *
- * If the read fails, ECL_ERR_READ is returned.
- *
  * This function is largely intended for peeking the size of the next record,
  * to approperiately allocate a large enough buffer, which is useful when
  * dealing with unknown files. If it is know in advance how large the records
  * are, it is not necessary to call this function before reading a record.
+ *
+ * Returns:
+ *
+ * ECL_EOF            if fp was currently at EOF
+ * ECL_UNEXPECTED_EOF if EOF was encountered mid-head
+ * ECL_ERR_READ       if reading failed because of an fstream error
+ * ECL_ERR_UNKNOWN    some other, uncaught error. this signals a bug in ecl
+ *
+ * Note that the file pointer *is still rolled back* after ECL_EOF, so foef(fp)
+ * will return false, even if the file *is* at the real EOF.
  */
 int eclfio_sizeof( FILE*, const char* opts, int32_t* out );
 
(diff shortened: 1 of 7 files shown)


joakim-hove:
> also propose we add a new data type in opts, C or s, for the fixed len-8 string.

We could even parse the type Cnnn - then we do not have to distingush between the 8 charatcter type and variable length type.

lib/src/fortio.cpp
```diff
@@ -409,3 +409,67 @@ int eclfio_put( std::FILE* fp,
...
+
+int eclfio_array_get( std::FILE* fp,
+                      const char* opts,
+                      int len,
+                      int nmemb,
+                      void* array ) {
+
+    if( nmemb < 0 ) return ECL_EINVAL;
+
+    if( len < 1 ) len = 1;
```
joakim-hove: What is the logic here?

jokva:
> We could even parse the type Cnnn - then we do not have to distingush between the 8 charatcter type and variable length type.

Surely, but on the flipside that *greatly* complicates the parseopts function.

lib/src/fortio.cpp
jokva: Just robustness, allowing zero or negative values as "don't worry about len at all), to avoid nasty divide-by-zeros.

joakim-hove:
> Surely, but on the flipside that greatly complicates the parseopts function.

That is obviously true - but *iff* that leads to a unified treatment of the CHAR and Cnnn types it might be worth it?

jokva:
> That is obviously true - but iff that leads to a unified treatment of the CHAR and Cnnn types it might be worth it?

Maybe, but having a concatenation feature has some pretty cool implications too, because then you can `put` your record headers as:

`eclfio_array_put(fp, "c", 16, 1, 1000, header)`

Which serves nicely as documentaiton.

jokva:
> `eclfio_array_put(fp, "c", 16, 1, 1000, header)`

In this context, adding `b` as an alias (`byte`) for `c` could also be useful.

lib/e3/ecl/fortio.h
```diff
@@ -166,6 +166,70 @@ int eclfio_get( FILE*, const char* opts, int32_t* recordsize, void* record );
...
+ */
+#define ECL_DEFAULT_BLOCKSIZE_NUMERIC 1000
+#define ECL_DEFAULT_BLOCKSIZE_STRING  105
+int ecl_default_blocksize( int );
+
+/*
+ * eclfio_array_get/put behave differently w.r.t. errors than their simpler
+ * relatives eclfio_get/put.
+ *
+ * When an operation cannot be successfully completed, eclfio_array does *not*
```
joakim-hove: I think these functions will be the main atomic units; so personally I think argument parsing, roll-back guarantees and so on should be at this level. The fortio level is not interesting for the user,

lib/e3/ecl/fortio.h
```diff
@@ -166,6 +166,70 @@ int eclfio_get( FILE*, const char* opts, int32_t* recordsize, void* record );
...
+
+int eclfio_array_get( FILE*,
+                      const char* opts,
+                      int len,
+                      int nmemb,
+                      void* array );
+int eclfio_array_put( FILE*,
+                      const char* opts,
+                      int len,
+                      int size,
```
joakim-hove: Extra size in header.

joakim-hove:
> In this context, adding b as an alias (byte) for c could also be useful.

Agree

joakim-hove:
This look fine to me; two things:

1. This is the time to get the  CHAR / Cnnnn implementation right. Maybe also the bool?
2. Personally I think these functions should be fundamental user exposed functions, .i.e. options parsing and rollback guard should be handled from here - the one level deeper functions from the previous PR should get a already parsed options object - and rollback should be handled here,

lib/e3/ecl/fortio.h
jokva: > I think these functions will be the main atomic units; so personally I think argument parsing, roll-back guarantees and so on should be at this level. The fortio level is not interesting for the user,

Who's _the user_?

I have mixed feelings about this. If rollback is done on array, then we'd either drop or have to communicate _which_ block failed, but if it only rolls back to the point that failed, then there's immediate recoverability.

Also, that means reading record headers can read with `eclfio_get` directly, with rollback semantics, which is nice.

That being said, `array_get` *will* roll back if nmemb < size.

joakim-hove: > Who's the user?

The "user" in this case is just code - probably same library, slighter higher up in the stack.

> If rollback is done on array, then we'd either drop or have to communicate which block failed, but if it only rolls back to the point that failed, then there's immediate recoverability.

Yes - my point is that array is the relevant atomic unit - in all(?) contexts; i.e. I do not see it as interesting functionality to roll back to middle of an array - i.e. a record.

> Also, that means reading record headers can read with eclfio_get directly, with rollback semantics, which is nice.

I agree that reading headers with rollback semantics is nice - and important, but they can - and should(?) - be implemented as array_get of 16 characters - i.e. again array smenatics; although the array in this case always only consists of one block.

jokva: > The "user" in this case is just code - probably same library, slighter higher up in the stack.

Sure, which is why I personally like this. It allows to check that specific block with sizeof, and quickly determine that the fortran block was of the wrong size, and maybe even pad the output array accordingly for robustness.

> Yes - my point is that array is the relevant atomic unit - in all(?) contexts; i.e. I do not see it as interesting functionality to roll back to middle of an array - i.e. a record.

The main use case is error handling, and I think the cost in interface is worth it personally.

markusdregi:
@jokva I don't see this work moving anytime soon, so please make a note as you see fit (mental, GitHub issue or as a Post-it note on your keyboard) of the potential gold in this work. I'm closing the PR until this is continued. Please reopen if you disagree.

PR #440, opened 2018-07-12
Add optional report_only argument to EclSum.numpy_vector

joakim-hove:


joakim-hove:
Ahhh yes - wait and see the test fail...

PR #445, opened 2018-08-02
WIP: ecl_grid can be allocated with int * instead of actnum_kw

stefoss23:
**Task**
Create a function for allocation of ecl_grid using int * instead of actnum_kw.

**Approach**
Added a new function:
ecl_grid_type * ecl_grid_alloc_ext_actnums(const char * , const int * * ext_actnums);

Currently only tested on main grid. 

**Pre un-WIP checklist**

### DIFF
--- a/lib/ecl/ecl_grid.cpp
+++ b/lib/ecl/ecl_grid.cpp
@@ -2733,8 +2733,8 @@ static ecl_grid_type * ecl_grid_alloc_GRDECL_kw__(ecl_grid_type * global_grid ,
 
 
 /**
-   If you create/load ecl_kw instances for the various fields, these
-   functions can be used to create a GRID instance, without going
+   If you create/load ecl_kw instances for the various fields, this
+   function can be used to create a GRID instance, without going
    through a GRID/EGRID file. Does not support LGR or coarsening
    hierarchies.
 */
@@ -3617,9 +3617,9 @@ ecl_grid_type * ecl_grid_alloc(const char * grid_file ) {
 
 
 // This function is used to override use of the keyword ACTNUM from the EGRID file.
-// ext_actnums must have size equal to the number of grids.
-// from ext_actnums[i] = NULL, actnum is taken from file, otherwise
-// ext_actnums[i] is used instead of the ACTNUM keyword.
+// ext_actnum must have size equal to the number of cells in the main grid
+// if ext_actnum = NULL, actnum is taken from file, otherwise ext_actnums
+// determines which cells are active.
 ecl_grid_type * ecl_grid_alloc_ext_actnum(const char * grid_file, const int * ext_actnum) {
   ecl_file_enum file_type = ecl_util_get_file_type(grid_file , NULL ,  NULL);
   if (file_type == ECL_EGRID_FILE)
--- a/lib/ecl/ecl_grid.cpp
+++ b/lib/ecl/ecl_grid.cpp
@@ -3049,7 +3049,7 @@ static ecl_grid_type * ecl_grid_alloc_EGRID__( ecl_grid_type * main_grid , const
 
 static ecl_grid_type * ecl_grid_alloc_GRID_all_grids(const char * filename) {
   util_abort("%s .GRID files - %s - not supported \n", __func__ , filename);
-  
+  return NULL;
 }
 
 
--- a/lib/ecl/ecl_grid.cpp
+++ b/lib/ecl/ecl_grid.cpp
@@ -2749,16 +2749,6 @@ ecl_grid_type * ecl_grid_alloc_GRDECL_kw( int nx, int ny , int nz ,
   if (actnum_kw)
     actnum_data = ecl_kw_get_int_ptr(actnum_kw);
 
-  return ecl_grid_alloc_GRDECL_kw_PORV( nx, ny , nz , zcorn_kw , coord_kw , actnum_data , mapaxes_kw );
-}
-
-
-ecl_grid_type * ecl_grid_alloc_GRDECL_kw_PORV( int nx, int ny , int nz ,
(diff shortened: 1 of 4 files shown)


joakim-hove:
Some changes:

1. Call the external actnum e.g. actnum_ext
2. The function should not be CALLED _PORV, but something more neutral.
3. It should be possible to use both an external actnum, and the actnum in the file. In that case the product should be the final property. That is because the actnum in file can also have value 2 for dual porosity models, that must be retained. The external actnum can only have value 0 or 1 and can only turn off cells.

lib/ecl/tests/ecl_grid_ext_actnum.cpp
```diff
@@ -0,0 +1,45 @@
...
+  //verify correct correct cells
+  test_work_area_type * work_area = test_work_area_alloc("ext_actnum_main_grid");
+  {
+    const char * filename = "FILE.EGRID";
+
+    ecl_grid_type * grid_write = ecl_grid_alloc_rectangular(2,2,1,   1,1,1,NULL);
+    ecl_grid_fwrite_EGRID( grid_write , filename, true);
+    ecl_grid_free(grid_write);
+
+    int * ext_actnum = new int[4];
```
joakim-hove: This can be: 
```C++
std::vector<int> ext_actnum = {0,1,0,1};
```
and then use `ext_actnum.data()` in the allocator.

lib/ecl/tests/ecl_grid_ext_actnum.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+#include <ert/ecl/ecl_endian_flip.hpp>
+#include <ert/ecl/ecl_kw.hpp>
+#include <ert/ecl/ecl_kw_magic.hpp>
+#include <ert/ecl/ecl_file.hpp>
+#include <ert/ecl/ecl_grid.hpp>
+
+
+void test_1() {
+  //make a main grid (4 cells)
```
joakim-hove: 6 cells.

stefoss23: The comments will be removed.

lib/ecl/ecl_grid.cpp
```diff
@@ -2685,9 +2686,9 @@ static ecl_grid_type * ecl_grid_alloc_GRDECL_kw__(ecl_grid_type * global_grid ,
                                                   const ecl_kw_type * gridhead_kw ,
                                                   const ecl_kw_type * zcorn_kw ,
                                                   const ecl_kw_type * coord_kw ,
-                                                  const ecl_kw_type * actnum_kw ,    /* Can be NULL */
+                                                  const int * actnum_data ,          /* Can be NULL */
```
joakim-hove: Please reorder the argument list here so that the `const int * actnum_data` comes either before all the `const ecl_kw_type *` arguments or after them - looks a bit funny when it comes in the middle.

lib/ecl/ecl_grid.cpp
```diff
@@ -3000,10 +3002,25 @@ static ecl_grid_type * ecl_grid_alloc_EGRID__( ecl_grid_type * main_grid , const
...
 
-
-  /** If ACTNUM is not present - that is is interpreted as - all active. */
-  if (ecl_file_get_num_named_kw(ecl_file , ACTNUM_KW) > grid_nr)
+  // If ACTNUM and ext_actnum are not present - that is is interpreted as - all active. 
+  const int * actnum_data = NULL;
+  if (ecl_file_get_num_named_kw(ecl_file , ACTNUM_KW) > grid_nr) {
     actnum_kw = ecl_file_iget_named_kw( ecl_file , ACTNUM_KW    , grid_nr);
+    actnum_data = ecl_kw_get_int_ptr(actnum_kw);
+  }
```
joakim-hove: I would continue the product block inside the `if (ecl_file_get_num_named_kw(ecl_file, ACTNUM_KW) > grid_nr)` block.

lib/ecl/ecl_grid.cpp
```diff
@@ -3000,10 +3002,25 @@ static ecl_grid_type * ecl_grid_alloc_EGRID__( ecl_grid_type * main_grid , const
...
+  
+  std::vector<int> actnum_product;
+  if (ext_actnum) {
+    if (actnum_kw) {
+      int size = ecl_kw_get_size(actnum_kw);
+      actnum_product = std::vector<int>(size);
```
joakim-hove: Here you create a new vector - nothing wrong with that as such; but I would rather do `actnum_product.resize()`

lib/ecl/ecl_grid.cpp
```diff
@@ -3603,6 +3624,21 @@ ecl_grid_type * ecl_grid_alloc(const char * grid_file ) {
...
+// This function is used to override use of the keyword ACTNUM from the EGRID file.
+// ext_actnum must have size equal to the number of cells in the main grid
+// if ext_actnum = NULL, actnum is taken from file, otherwise ext_actnums
+// determines which cells are active.
+ecl_grid_type * ecl_grid_alloc_ext_actnum(const char * grid_file, const int * ext_actnum) {
+  ecl_file_enum file_type = ecl_util_get_file_type(grid_file , NULL ,  NULL);
+  if (file_type == ECL_EGRID_FILE)
+    return ecl_grid_alloc_EGRID_all_grids(grid_file, true, ext_actnum);
+  else
+    util_abort("%s must have .EGRID file - %s not recognized \n", __func__ , grid_file);
```
joakim-hove: Would prefer if your wrote a (empty ?) `ecl_grid_alloc_GRID_all_grids()` which takes the `ext_actnum` - the body of that function can just be `util_abort()`

PR #447, opened 2018-08-09
WIP: Get unit system of ecl_grid

stefoss23:
**Task**
Access info about the unit system of the ecl grid. 

**Approach**
Added function: ecl_grid_get_unit_system. Output is of type ert_ecl_unit_enum (ECL_METRIC_UNITS etc.).

**Pre un-WIP checklist**

### DIFF
--- a/lib/ecl/ecl_grid.cpp
+++ b/lib/ecl/ecl_grid.cpp
@@ -763,6 +763,10 @@ struct ecl_grid_struct {
   int                   eclipse_version;
 };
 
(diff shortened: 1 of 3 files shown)


joakim-hove:
@magnesj - can you take a look on this?

bjornjensen:
The new function always return ECL_METRIC_UNITS. It seems that the grid->unit_system is never updated with the correct value.

joakim-hove:
Ok - will fix

PR #456, opened 2018-08-21
Ensure that rng_std_normal() is valid

joakim-hove:
The algorithm creating a normally distributed value from a [0,1) random value will fail if input is 0.

### DIFF
--- a/lib/util/rng.cpp
+++ b/lib/util/rng.cpp
@@ -241,6 +241,19 @@ double rng_get_double( rng_type * rng ) {
   return rng->forward( rng->state ) * rng->inv_max;
 }
 
+static double rng_get_positive_double(rng_type * rng) {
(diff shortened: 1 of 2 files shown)


lib/util/rng.cpp
```diff
@@ -285,9 +298,9 @@ void rng_shuffle_int( rng_type * rng , int * data , size_t num_elements) {
 /*****************************************************************/
 
 double rng_std_normal( rng_type * rng ) {
-  const double pi = 3.141592653589;
-  double R1 = rng_get_double( rng );
-  double R2 = rng_get_double( rng );
+  const double pi = 3.1415926535897932384626433832795028841971693993751058209749445923;
```
stefoss23: pi should have a more general place of definition.

PR #458, opened 2018-08-24
WIP: Split statoil grid test

stefoss23:
**Task**
The test statoil_grid_test takes annoyingly long time to run. 

**Approach**
The test is split in four. The shortests test functions are still in the original test file, 3 test functions (with long run times) are moved into separate test files.

**Pre un-WIP checklist**

### DIFF
--- a/python/tests/ecl_tests/test_grid_statoil_coarse.py
+++ b/python/tests/ecl_tests/test_grid_statoil_coarse.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 #  Copyright (C) 2018  Statoil ASA, Norway.
 #
-#  The file 'test_grid_statoil_3.py' is part of ERT - Ensemble based Reservoir Tool.
+#  The file 'test_grid_statoil_1.py' is part of ERT - Ensemble based Reservoir Tool.
 #
 #  ERT is free software: you can redistribute it and/or modify
 #  it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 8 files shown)


python/tests/ecl_tests/test_grid_statoil_coarse.py
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env python
```
joakim-hove: These test should not have the `#!/usr/bin/env python` sequence at the top.

python/tests/ecl_tests/test_grid_statoil_coarse.py
```diff
@@ -0,0 +1,56 @@
...
+from ecl.util.util import DoubleVector, IntVector
+from ecl.util.test import TestAreaContext
+from tests import EclTest, statoil_test
+
+
+@statoil_test()
+class GridCoarceTest(EclTest):
+
+
+    @skipIf(EclTest.slowTestShouldNotRun(), "Slow test of coarse grid skipped!")
```
joakim-hove: Just remove the `@skipIf` functionality - it is not maintained.

python/tests/ecl_tests/test_grid_statoil_coarse.py
```diff
@@ -0,0 +1,56 @@
...
+from cwrap import Prototype
+from cwrap import open as copen
+
+import time
+from ecl import EclDataType, EclUnitTypeEnum
+from ecl.eclfile import EclKW, EclFile, openEclFile
+from ecl.grid import EclGrid
+from ecl.util.util import DoubleVector, IntVector
+from ecl.util.test import TestAreaContext
+from tests import EclTest, statoil_test
```
joakim-hove: For all of the new tests the `import` block can be significantly reduced/simplified. I *think* this should be enough:

```
from ecl.util.test import TestAreaContext
from ecl.grid import EclGrid

from tests impoty EclTest, statoil_test
```

PR #462, opened 2018-09-01
Use std::vector<std::string> instead of stringlist

joakim-hove:
A small step ....

### DIFF
--- a/lib/ecl/ecl_file_view.cpp
+++ b/lib/ecl/ecl_file_view.cpp
@@ -17,9 +17,11 @@
 */
 
 
+#include <vector>
+#include <string>
+
 #include <ert/util/vector.hpp>
 #include <ert/util/hash.hpp>
-#include <ert/util/stringlist.hpp>
 
 #include <ert/ecl/fortio.h>
 #include <ert/ecl/ecl_kw.hpp>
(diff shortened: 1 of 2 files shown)


lib/ecl/ecl_file_view.cpp
```diff
@@ -33,7 +35,7 @@
 struct ecl_file_view_struct {
   vector_type       * kw_list;      /* This is a vector of ecl_file_kw instances corresponding to the content of the file. */
   hash_type         * kw_index;     /* A hash table with integer vectors of indices - see comment below. */
-  stringlist_type   * distinct_kw;  /* A stringlist of the keywords occuring in the file - each string occurs ONLY ONCE. */
+  std::vector<std::string> * distinct_kw;  /* A list of the keywords occuring in the file - each string occurs ONLY ONCE. */
```
andreabrambilla: I was wondering, do we really need to make this dynamically allocated? This can easily become a class member and there will be no need for `new` and `delete`

lib/ecl/ecl_file_view.cpp
```diff
@@ -200,11 +202,12 @@ int ecl_file_view_find_kw_value( const ecl_file_view_type * ecl_file_view , cons
 }
 
 const char * ecl_file_view_iget_distinct_kw( const ecl_file_view_type * ecl_file_view , int index) {
-  return stringlist_iget( ecl_file_view->distinct_kw , index);
+  const std::string& string = ecl_file_view->distinct_kw->operator[](index);
```
andreabrambilla: This is perfectly fine, but I think `ecl_file_view->distinct_kw[index]` looks a bit better

lib/ecl/ecl_file_view.cpp
joakim-hove: Thank you for the review. Yes I would also have preferred the `distinct_kw` to be a class member, but that would require that the parent - i.e. `ecl_file_view` was handled as a proper class, and not just a C struct as is yet the case - and that would be more invasive change.

lib/ecl/ecl_file_view.cpp
joakim-hove: I fully agree  - it is the same argument as above.

joakim-hove:
> up to you to address them or not

Would have liked to - but ...

joakim-hove:
@andreabrambilla : My C++ skills are limited - but growing; I got away by implementing the `xxx_alloc()` function with the default constructor - then the parent "class" could be implemented with `new` and `delete` and then the `std::vector<std::string>` member could have value semantics - that was cool!

PR #463, opened 2018-09-02
path_stack_pop returns void

joakim-hove:
Small simplification in the step towards C++

### DIFF
--- a/lib/include/ert/util/path_stack.hpp
+++ b/lib/include/ert/util/path_stack.hpp
@@ -27,7 +27,7 @@ extern "C" {
   typedef struct path_stack_struct path_stack_type;
 
   path_stack_type * path_stack_alloc();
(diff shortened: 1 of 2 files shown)


lib/include/ert/util/path_stack.hpp
```diff
@@ -27,7 +27,7 @@ extern "C" {
   typedef struct path_stack_struct path_stack_type;
 
   path_stack_type * path_stack_alloc();
-  const char      * path_stack_pop( path_stack_type * path_stack );
+  void              path_stack_pop( path_stack_type * path_stack );
```
andreabrambilla: I have the feeling the goal here is to reproduce the API of c++ `std::stack` (https://en.cppreference.com/w/cpp/container/stack). In that case, it may be also worth renaming `path_stack_peek` to `path_stack_top`

PR #465, opened 2018-09-05
Path stack simplification

joakim-hove:
Using `std::stack` in the `path_stack` implementation.

### DIFF
--- a/lib/util/path_stack.cpp
+++ b/lib/util/path_stack.cpp
@@ -42,7 +42,6 @@
 
 struct path_stack_struct {
   stringlist_type * stack;
-  stringlist_type * storage;
 };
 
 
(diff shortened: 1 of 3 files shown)


lib/util/path_stack.cpp
```diff
@@ -41,8 +44,7 @@
 
 
 struct path_stack_struct {
-  stringlist_type * stack;
-  stringlist_type * storage;
+  std::stack<std::string> stack;
```
andreabrambilla: A very minor comment here. I guess a path stack would hardly have more than 10 or 20 elements. In that case, it is better (performance-wise) to use `std::stack< std::string, std::vector<std::string> >`, or even just `std::vector<std::string>`

joakim-hove: OK; `std::vector<std::string>` it is.

PR #469, opened 2018-09-13
Ecl grid processed copy double

joakim-hove:


joakim-hove:
@totto82: Will you give this a spin

totto82:
This patch does not solve my issue with strange grids. But avoiding converting to float makes it easier to debug my real problem. I will leave it up to you whether you want to merge this or not.

joakim-hove:
Ok - I'll merge it.

PR #485, opened 2018-09-25
Made function for exporting ecl_grid index data to pandas dataframe.

stefoss23:
**Task**
Make function for exporting ecl_grid index data to pandas dataframe.

**Approach**
Similar to that for exporting ecl_sum data to pandas dataframe.

**Pre un-WIP checklist**

### DIFF
--- a/python/ecl/grid/ecl_grid.py
+++ b/python/ecl/grid/ecl_grid.py
@@ -1256,7 +1256,7 @@ class EclGrid(BaseCClass):
 
         return self._create_volume_keyword(active_size)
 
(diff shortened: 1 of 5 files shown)


python/ecl/grid/ecl_grid.py
```diff
@@ -1254,6 +1256,18 @@ def create_volume_keyword(self, active_size=True):
 
         return self._create_volume_keyword(active_size)
 
+    def export_index_frame(self, active_only = False):
```
joakim-hove: Just call it `export_index`

PR #486, opened 2018-09-26
WIP: Export kw as numpy with ecl_grid pandas frame as input

stefoss23:
**Task**
Make a function in EclGrid: export_data(index_frame, kw) -> numpy.array. 
index_frame is created by EclGrid.export_index(bool active_only) and is a pandas.DataFrame
len(numpy.array) = len(index_frame.index) and can be either global size or total_active size.

**Pre un-WIP checklist**

### DIFF
--- a/lib/ecl/ecl_grid.cpp
+++ b/lib/ecl/ecl_grid.cpp
@@ -502,6 +502,7 @@ which splits the lower face along the 0-3 diagnoal and the upper face
 along the 5-6 diagonal.
 */
 
+#define INDEX_DATA_FRAME_SIZE 4
 
 static const int tetrahedron_permutations[2][12][3] = {{
                                                         // K-
@@ -7093,7 +7094,7 @@ ecl_kw_type * ecl_grid_alloc_volume_kw( const ecl_grid_type * grid , bool active
 
 
 //Note: global_index must be allocated w/ ecl_grid->total_active or ecl_grid->global_size int32 data points
-//      index_data must be allocated w/ (4 * ecl_grid->total_active or ecl_grid->global_size) int32 data points.
+//      index_data must be allocated w/ (INDEX_DATA_FRAME_SIZE * ecl_grid->total_active or ecl_grid->global_size) int32 data points.
 void ecl_grid_export_index(const ecl_grid_type * grid, int * global_index, int * index_data , bool active_only) {
   int pos_indx = 0;
   int pos_data = 0;
@@ -7109,6 +7110,11 @@ void ecl_grid_export_index(const ecl_grid_type * grid, int * global_index, int *
           index_data[pos_data++] = grid->cells[g].active_index[0];
         }          
       }
+}
+
+bool ecl_grid_export_data_as_int( ecl_grid_type * grid, int index_size, int * global_index, ecl_kw_type * kw, int * output) {
+  
+  return true;
 }
 
 //
--- a/lib/ecl/ecl_grid.cpp
+++ b/lib/ecl/ecl_grid.cpp
@@ -502,7 +502,6 @@ which splits the lower face along the 0-3 diagnoal and the upper face
 along the 5-6 diagonal.
 */
 
-#define INDEX_DATA_FRAME_SIZE 4
 
 static const int tetrahedron_permutations[2][12][3] = {{
                                                         // K-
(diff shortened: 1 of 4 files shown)


lib/ecl/ecl_grid.cpp
```diff
@@ -7111,4 +7112,53 @@ void ecl_grid_export_index(const ecl_grid_type * grid, int * global_index, int *
       }
 }
 
+//Note: index_size must equal allocated size of output
+void ecl_grid_export_data_as_int( ecl_grid_type * grid, int index_size, int * global_index, ecl_kw_type * kw, int * output) {
```
joakim-hove: `const` "everywhere":
```C++
void ecl_grid_export_data_as_int(const ecl_grid_type * grid, int index_size, const int * global_index, const ecl_kw_type * kw, ...)
```

lib/ecl/ecl_grid.cpp
```diff
@@ -7111,4 +7112,53 @@ void ecl_grid_export_index(const ecl_grid_type * grid, int * global_index, int *
...
+  int * input = ecl_kw_get_int_ptr(kw);
+  if (index_size == ecl_kw_get_size(kw)) {
+    for (int g = 0; g < index_size; g++)
```
joakim-hove: Here you can just use `memcpy(output, input, index_size * sizeof * output)`

lib/ecl/ecl_grid.cpp
```diff
@@ -7111,4 +7112,53 @@ void ecl_grid_export_index(const ecl_grid_type * grid, int * global_index, int *
       }
 }
 
+//Note: index_size must equal allocated size of output
+void ecl_grid_export_data_as_int( ecl_grid_type * grid, int index_size, int * global_index, ecl_kw_type * kw, int * output) {
+  int * input = ecl_kw_get_int_ptr(kw);
```
joakim-hove: The size of the index should not be relevant, the important question to if on is the size of keyword. Before calling this function we must check the size of the keyword, and then send either active_index of global_index; something like:
```python
if len(kw) == grid.get_global_size():
   index = index_frame.get_global
elif len(kw) == grid.get_num_actiev():
   index = index_frame.get_active()
else:
   raise ValueError("The keyword must have a 3D compatible length")
```
and then the C++ code could look like:
```C++
ecl_grid_export_data_as_xxx(, ... int index_size, const int * data_index, .... )
...
for (int i=0; i < index_size; i++) {
     int di = data_index[i];
     if (di >= 0)
         output[i] = input[di]
}
```

lib/ecl/ecl_grid.cpp
```diff
@@ -7111,4 +7112,53 @@ void ecl_grid_export_index(const ecl_grid_type * grid, int * global_index, int *
...
+
+
+//Note: index_size must equal allocated size of output
+void ecl_grid_export_data_as_double( ecl_grid_type * grid, int index_size, int * global_index, ecl_kw_type * kw, double * output) {
+
+  if (index_size == ecl_kw_get_size(kw)) {
+    for (int g = 0; g < index_size; g++)
+      output[g] = ecl_kw_iget_as_double(kw, g);
+  }
+  else if (index_size == grid->size) {
```
joakim-hove: Same as for the int type.

python/ecl/grid/ecl_grid.py
```diff
@@ -1267,7 +1269,33 @@ def export_index(self, active_only = False):
         df = pandas.DataFrame(data=data, index=indx)
         return df
         
- 
+    def export_data(self, index_frame, kw):
+        if not isinstance(index_frame, pandas.DataFrame):
+            raise TypeError("index_frame must be pandas.DataFrame")
+        global_index = numpy.array( index_frame.index )
+        if not(len(global_index) == self.get_global_size() or len(global_index) == self.get_num_active()):
```
joakim-hove:
```python
active_index = index_frame["active_index"]
```

python/ecl/grid/ecl_grid.py
```diff
@@ -1267,7 +1269,33 @@ def export_index(self, active_only = False):
         df = pandas.DataFrame(data=data, index=indx)
         return df
         
- 
+    def export_data(self, index_frame, kw):
+        if not isinstance(index_frame, pandas.DataFrame):
+            raise TypeError("index_frame must be pandas.DataFrame")
+        global_index = numpy.array( index_frame.index )
```
joakim-hove: This makes a new numpy array object with the global index - that should not be necessary.

python/ecl/grid/ecl_grid.py
joakim-hove: This test is wrong - the index can be arbitrarily long or short; it is the length of the kw which must agree with the grid,

PR #491, opened 2018-10-01
Readme updates

joakim-hove:
Statoil/libres#392
Statoil/ert#153

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,55 +8,75 @@ non-unified and formatted and unformatted files are supported.
 
 *libecl* is mainly developed on *Linux* and *OS X*, in addition there
 is a portability layer which ensures that most of the functionality is
-available on *Windows*. The main functionality is written in C, and
+available on *Windows*. The main functionality is written in C/C++, and
 should typically be linked in in other compiled programs. *libecl* was
 initially developed as part of the [Ensemble Reservoir
 Tool](http://github.com/Statoil/ert), other applications using
 *libecl* are the reservoir simulator flow and Resinsight from the [OPM
 project](http://github.com/OPM/).
 
-In addition to the C code there are Python wrappers which make most of
-the *libecl* functionality available from Python. For small interactive
-scripts, forward models e.t.c. this is recommended way to use *libecl*
-functionality.
+In addition to the compiled C/C++ code there are Python wrappers which make most
+of the *libecl* functionality available from Python. For small interactive
+scripts, forward models e.t.c. this is the recommended way to use *libecl*
+functionality. You decide wether to build include the Python wrappers when
+configuring the `cmake` build - pass the option `-DENABLE_PYTHON=ON` to enable
+Python wrappers, by default the Python wrappers are *not* included.
 
+By default `cmake` is configured to install to a system location like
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -49,14 +49,8 @@ will work:
 
     sudo pip install -r requirements.txt
 
-The Python + cmake interaction is handled in a separate project called
-[pycmake](https://github.com/Statoil/pycmake); you can either install
```
markusdregi: Should we not still mention `pycmake` in the README?

README.md
```diff
@@ -49,14 +49,8 @@ will work:
...
-correct version of `pycmake` into your *libecl* code:
+The Python + cmake interaction is handled with two cmake modules in
+`python/cmake/Modules`. These modules will be installed under
+`$prefix/share/cmake/Modules` when you install `libecl`, and they will also be
+used by the downstream modules `libres` and `ert`.
 
-    git submodule update --init pycmake
-
-
-### Installing the Python code ###
```
markusdregi: How do you intend people to install the Python code? With `pip` or `setup.py` or the `CMAKE - enable python` flag?

joakim-hove: > How do you intend people to install the Python code? 

For now the `cmake --enable-python` is the by far best tested. At some point we should probably combine the `cmake + setup.py` based approaches. `pip install` will always only be a Python only solution - so if you are also installing `libres, ...` which utilizes the underlying C code directly, you can for the foreseeable future not use `pip`.

README.md
joakim-hove: > Should we not still mention pycmake in the README? 

My suggestion is no; the problem is that `libres` and `ert` will use the `cmake` modules carried and installed by `libecl`; i.e. the dependency to these files is stronger than what is communicated through the use of a submodule.

The best thing would in my opinion be if we could deploy it to `pypi` - then we could add `pycmake` in the `requirements.txt` independently for `libecl, libres` and `ert` - but that requires some effort; so my suggestion is the current PR. No strong feelings though.

joakim-hove: There is no changes to the pycmake code, we could even duplicate it across the three repos. That way we simplify the build somewhat - I would actually vote for that.

andreabrambilla: minor comment: there is a quote charachter `"` after `build system`that is not matched anywhere

As was discussed earlier today, replicating the pycmake code can be a good option for the time being

README.md
```diff
@@ -8,55 +8,53 @@ non-unified and formatted and unformatted files are supported.
...
-shared libraries built from the C code.
-
-You need to install some Python requirements before the Python code
-will work:
+### Alternative 2: Building with Python ###
+To build *libecl* with Python wrappers you just pass the option
+`-DENABLE_PYTHON=ON` when running `cmake`. In addition you need to install some
+dependencies first:
 
     sudo pip install -r requirements.txt
```
andreabrambilla: if we leave this line here, it's unclear when it has to be run (I know it's obvious if you know what you are doing, but let's make it as easy as possible).
I'd suggest to include it in the script below after the `cd libecl`

joakim-hove: OK - updated.

PR #507, opened 2018-10-23
Smspec member functions converted to new c++ standard

stefoss23:
**Task**
Converted rest of hast_type and char * to std::map and std::string.

**Pre un-WIP checklist**

### DIFF
--- a/lib/ecl/ecl_smspec.cpp
+++ b/lib/ecl/ecl_smspec.cpp
@@ -114,10 +114,10 @@ struct ecl_smspec_struct {
 
   std::map<std::string, node_map> well_var_index; /* Indexes for all well variables:
                                                      {well1: {var1: index1 , var2: index2} , well2: {var1: index1 , var2: index2}} */
-  std::map<std::string, node_map> group_var_index; /* Indexes for group variables.*/
-  std::map<std::string, std::map<int, smspec_node_type*>> region_var_index;  /* The stored index is an offset. */
-  std::map<std::string, std::map<int, smspec_node_type*>>  block_var_index;            /* Block variables like BPR */
-  std::map<std::string, std::map<int, node_map>>           well_completion_var_index; /* Indexes for completion indexes .*/
+  std::map<std::string, node_map> group_var_index;   /* Indexes for group variables.*/
+  std::map<int,         node_map> region_var_index;  /* The stored index is an offset. */
+  std::map<int,         node_map> block_var_index;   /* Block variables like BPR */
+  std::map<std::string, std::map<int, node_map>> well_completion_var_index; /* Indexes for completion indexes .*/
 
 
   std::vector<smspec_node_type*> smspec_nodes;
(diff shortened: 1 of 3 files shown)


lib/ecl/ecl_smspec.cpp
```diff
@@ -108,15 +108,18 @@ struct ecl_smspec_struct {
...
-  hash_type          * group_var_index;            /* Indexes for group variables.*/
-  hash_type          * field_var_index;
-  hash_type          * region_var_index;           /* The stored index is an offset. */
-  hash_type          * misc_var_index;             /* Variables like 'TCPU' and 'NEWTON'. */
-  hash_type          * block_var_index;            /* Block variables like BPR */
-  node_map             gen_var_index              /* This is "everything" - things can either be found as gen_var("WWCT:OP_X") or as well_var("WWCT" , "OP_X") */;
+  std::map<std::string, node_map> group_var_index; /* Indexes for group variables.*/
+  std::map<std::string, std::map<int, smspec_node_type*>> region_var_index;  /* The stored index is an offset. */
+  std::map<std::string, std::map<int, smspec_node_type*>>  block_var_index;            /* Block variables like BPR */
+  std::map<std::string, std::map<int, node_map>>           well_completion_var_index; /* Indexes for completion indexes .*/
```
joakim-hove: The `region_var_index`, `block_var_index` and `well_completion_var_index` members should be changed to:
```
std::map<int, node_map>
```
that way the final lookup can be based on `node_map` for all variables.

lib/ecl/ecl_smspec.cpp
```diff
@@ -837,47 +833,24 @@ static void ecl_smspec_install_special_keys( ecl_smspec_type * ecl_smspec , smsp
...
   case(ECL_SMSPEC_REGION_VAR):
-    if (!hash_has_key(ecl_smspec->region_var_index , keyword))
-      hash_insert_hash_owned_ref( ecl_smspec->region_var_index , keyword , hash_alloc() , hash_free__);
-    {
-      hash_type * var_hash = (hash_type*)hash_get(ecl_smspec->region_var_index , keyword);
-      char num_str[16];
-      sprintf( num_str , "%d" , num);
-      hash_insert_ref(var_hash , num_str , smspec_node);
-    }
+    ecl_smspec->region_var_index[keyword][num] = smspec_node;
```
joakim-hove: This should be: `region_var_index[num][keyword] = smspec_node;`

lib/ecl/ecl_smspec.cpp
```diff
@@ -1042,14 +1015,16 @@ static void ecl_smspec_load_restart( ecl_smspec_type * ecl_smspec , const ecl_fi
       }
 #endif
 
-      util_alloc_file_components( ecl_smspec->header_file , &path , NULL , NULL );
+      util_alloc_file_components( ecl_smspec->header_file.c_str() , &path , NULL , NULL );
```
joakim-hove: Here you can use:
```C++
#include "detail/util/path.hpp"
...
std::string path = ecl::util::path::dirname( ecl_smspec->header_file );
```

lib/ecl/ecl_smspec.cpp
```diff
@@ -1250,6 +1226,36 @@ static bool ecl_smspec_fread_header(ecl_smspec_type * ecl_smspec, const char * h
 }
 
 
+/******************************************************************/
+
+static const smspec_node_type * ecl_smspec_get_var_node( const node_map& mp, const char * var) {
```
joakim-hove: Instead of using `static` you should use an anonymus namespace - that is the C++ way:
```
namespace {
    
    const smspec_node_type * ecl_smspec_get_var_node( ) {  }

}  
```

joakim-hove: Or - maybe not; the `static` keyword seems to have back in favor also for C++: https://stackoverflow.com/questions/154469/unnamed-anonymous-namespaces-vs-static-functions

PR #509, opened 2018-10-24
Resample fixup

joakim-hove:
This is a frustarting f...ing hack to get nexus through


stefoss23:
And, it fails in the ecl_sum_alloc_resampled_test...

PR #511, opened 2018-10-25
WIP: ecl_sum_tstep: parameter data is std::Vector<float>.

stefoss23:
**Task**
ecl_sum_tstep.cpp: data is std::Vector<float>.

**Pre un-WIP checklist**


lib/ecl/ecl_sum_tstep.cpp
```diff
@@ -297,10 +302,10 @@ void ecl_sum_tstep_fwrite( const ecl_sum_tstep_type * ministep , const int * ind
 /*****************************************************************/
 
 void ecl_sum_tstep_iset( ecl_sum_tstep_type * tstep , int index , float value) {
-  if ((index < tstep->data_size) && (index >= 0))
+  if ((index < (int)tstep->data.size()) && (index >= 0))
```
joakim-hove: C++ way: `static_cast<int>( )`

PR #512, opened 2018-10-25
WIP: Smspec std::vector in api

stefoss23:
**Task**
A suggestion for using std::vector in the API. 

**Pre un-WIP checklist**

### DIFF
--- a/lib/ecl/ecl_smspec.cpp
+++ b/lib/ecl/ecl_smspec.cpp
@@ -1648,12 +1648,9 @@ const char * ecl_smspec_get_restart_case( const ecl_smspec_type * ecl_smspec) {
     return NULL;
 }
 
-
-const float * ecl_smspec_get_params_default( const ecl_smspec_type * ecl_smspec ) {
-  return ecl_smspec->params_default.data();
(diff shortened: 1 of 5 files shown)


lib/ecl/ecl_smspec.cpp
```diff
@@ -1648,13 +1648,10 @@ const char * ecl_smspec_get_restart_case( const ecl_smspec_type * ecl_smspec) {
     return NULL;
 }
 
-
-const float * ecl_smspec_get_params_default( const ecl_smspec_type * ecl_smspec ) {
-  return ecl_smspec->params_default.data();
+const std::vector<float>& ecl_smspec_get_params_default( const ecl_smspec_type * ecl_smspec ) {
```
joakim-hove: If you implement this as:
```C++
const float * ecl_smspec_get_params_default( ) {
     return ecl_smspec->params_default.data();
}
```

you can get away without API changes.

stefoss23: True, and this is the solution used in master. But, if a std::vector is returned, use of ecl_smspec_get_params_default is simpler and one does not have to resort to raw pointer operations.

PR #515, opened 2018-10-29
WIP: vector_resize takes extra arg default_value.

stefoss23:
**Approach**
As a part of converting libecl/libres/ert code to c++, the xxx_vector_type objects are under removal, and will be replaced by std::vector. To facilitate the removal, the API of xxx_vector_type is made similar to std::vector. One obstacle is the use of the default_value parameter. This is memorized and used for resize operations, whereas for std::vector, this parameter has to be specified for each resize operation (or 0 is used). 

This PR disregards the setting of default_value for the function xxx_vector_resize.

**Pre un-WIP checklist**

Statoil/libres#414

### DIFF
--- a/lib/util/tests/ert_util_type_vector_test.cpp
+++ b/lib/util/tests/ert_util_type_vector_test.cpp
@@ -263,8 +263,8 @@ void test_resize() {
   int i;
   int def = 77;
   int_vector_type * vec = int_vector_alloc(0,def);
-  int_vector_resize( vec , 10 );
-  test_assert_int_equal( int_vector_size( vec ) , 10 );
+  int_vector_resize( vec , 10 , int_vector_get_default(vec) );
+  test_assert_int_equal( int_vector_size( vec ) , 10  );
   for (i=0; i < 10; i++)
     test_assert_int_equal( int_vector_iget( vec , i ) , def );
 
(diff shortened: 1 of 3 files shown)


lib/util/tests/ert_util_type_vector_test.cpp
```diff
@@ -263,21 +263,21 @@ void test_resize() {
   int i;
   int def = 77;
   int_vector_type * vec = int_vector_alloc(0,def);
-  int_vector_resize( vec , 10 );
-  test_assert_int_equal( int_vector_size( vec ) , 10 );
+  int_vector_resize( vec , 10 , int_vector_get_default(vec) );
```
joakim-hove: I guess the `xxx_get_default()` function will have to go as well, so I would just use the `def` variable here, instead of invoking the `int_vector_get_default()` function.

lib/util/tests/ert_util_type_vector_test.cpp
```diff
@@ -373,11 +373,21 @@ void test_equal_index() {
...
 
-
+void test_misc() {
+  int_vector_type * v = int_vector_alloc(5, 123);
+  test_assert_int_equal( int_vector_iget(v, 2), 123 );
+  int_vector_resize(v, 20, 0);
+  test_assert_int_equal( int_vector_iget(v, 4), 123);
+  test_assert_int_equal( int_vector_iget(v, 9), 0 );
+  test_assert_int_equal( int_vector_iget(v, 19), 0);
+  //int_vector_iset(v, 10, 66);
```
joakim-hove: Remove the commented code

lib/util/tests/ert_util_type_vector_test.cpp
```diff
@@ -373,11 +373,21 @@ void test_equal_index() {
...
+  //test_assert_int_equal( int_vector_iget(v, 10), 66);
+}
 
 int main(int argc , char ** argv) {
+  test_misc();
 
-  int_vector_type * int_vector = int_vector_alloc( 0 , 99);
+  /*int_vector_type * int_vector = int_vector_alloc( 0 , 99);
```
joakim-hove: Looks like a large block of code has been commented out?

stefoss23: A mistake, uncommented.

PR #518, opened 2018-11-01
WIP: nnc_vector.cpp: grid_index_list and nnc_index_list as std::vector.

stefoss23:
**Task**
As part of converting the ert-code to c++, the nnc_vector objects will contain std::vector instead of int_vector_type.

**Pre un-WIP checklist**

### DIFF
--- a/lib/ecl/ecl_nnc_export.cpp
+++ b/lib/ecl/ecl_nnc_export.cpp
@@ -17,6 +17,8 @@
 */
 #include <stdlib.h>
 
+#include <vector>
+
 #include <ert/util/int_vector.hpp>
 
 #include <ert/ecl/ecl_file.hpp>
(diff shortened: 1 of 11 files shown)


lib/ecl/ecl_nnc_export.cpp
```diff
@@ -17,6 +17,8 @@
 */
 #include <stdlib.h>
 
+#include <vector>
+
 #include <ert/util/int_vector.hpp>
```
joakim-hove: It should be possible to remove the `#include <ert/util/int_vector.hpp>` line?

lib/ecl/nnc_info.cpp
```diff
@@ -151,27 +156,28 @@ void nnc_info_add_nnc(nnc_info_type * nnc_info, int lgr_nr, int global_cell_numb
...
 
-const int_vector_type * nnc_info_get_grid_index_list(const nnc_info_type * nnc_info, int lgr_nr) {
+const std::vector<int>& nnc_info_get_grid_index_list(const nnc_info_type * nnc_info, int lgr_nr) {
   nnc_vector_type * nnc_vector = nnc_info_get_vector( nnc_info , lgr_nr );
-  if (nnc_vector)
-    return nnc_vector_get_grid_index_list( nnc_vector );
-  else
-    return NULL;
+  if (!nnc_vector)
+    throw std::invalid_argument(std::string(__func__));
```
joakim-hove: This is completely new semantics; it looks like return `NULL` was completely OK previously; we can then *not* suddenly throw an exception - without really checking the callsites!

lib/include/ert/util/vector_util.hpp
```diff
@@ -0,0 +1,59 @@
...
+   See the GNU General Public License at <http://www.gnu.org/licenses/gpl.html>
+   for more details.
+*/
+
+#include <stdio.h>
+
+#include <vector>
+
+template <class T>
+int vector_util_index(const std::vector<T>& vec, T value) {
```
joakim-hove: A bit skeptical to add this - should be possible to implement equivalent functionality with `std::find`
```
auto iter = std::find(vec.begin(), vec.end, value);
if (iter == vec.end())
    index = -1;
else
    index = iter - vec.begin();
```

lib/ecl/nnc_info.cpp
stefoss23: I can only find the function in test cases outside of nnc_info.cpp.

lib/include/ert/util/vector_util.hpp
stefoss23: It is refactoring . This is a replacement of a function from @type@_vector_index. If the function is not replaced, we will have to use 5 lines of code every time we want to find an index to a value.

abrautomate:

PR #530, opened 2018-11-12
Compute geertsma based on dynamic porevolume

markusdregi:
Computing time shifts based on dynamic porevolumes

### DIFF
--- a/lib/ecl/ecl_subsidence.cpp
+++ b/lib/ecl/ecl_subsidence.cpp
@@ -75,6 +75,7 @@ struct ecl_subsidence_survey_struct {
   char                      * name;           /* Name of the survey - arbitrary string. */
   double                    * porv;           /* Reference pore volume */
   double                    * pressure;              /* Pressure in each grid cell at survey time */
+  double                    * dynamic_porevolume; /* Porevolume in each grid cell at survey time */
 };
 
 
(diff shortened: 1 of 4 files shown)


lib/ecl/ecl_subsidence.cpp
```diff
@@ -193,6 +205,43 @@ static double ecl_subsidence_survey_eval_geertsma( const ecl_subsidence_survey_t
...
 
+static double ecl_subsidence_survey_eval_geertsma_rporv( const ecl_subsidence_survey_type * base_survey ,
+                                                   const ecl_subsidence_survey_type * monitor_survey,
+                                                   ecl_region_type * region ,
+                                                   double utm_x , double utm_y , double depth,
+                                                   double youngs_modulus, double poisson_ratio, double seabed) {
+
+  const ecl::ecl_grid_cache& grid_cache = *(base_survey->grid_cache);
+  const int size  = grid_cache.size();
+  double * weight = (double*)util_calloc( size , sizeof * weight );
```
joakim-hove: This could be:
```
std::vector<double> weight(size);
```

and then in the call to: `ecl_grav_common_eval_geertsma` you can use `weight.data()`.

joakim-hove:
This is good - I had a minor `C++` comment; adress it if you like - and then wham-bang-merge!

PR #531, opened 2018-11-12
WIP: Changed ecl_sum_vector to use numpy_dates instead of mpl_dates

wouterjdb:
**Task**
ecl_sum_vector.py makes use of get_mpl_dates() which leads to a depreciation warning. ecl_sum_vector should use the new numpy_dates()

**Approach**
Replaced all occurances of "mpl_dates" with "numpy_dates". Maintainted __mpl_dates property for backward compatibility. Usage of the mpl_dates property will trigger a Depreciation Warning.

**Pre un-WIP checklist**

Did not perform tests. Need some assistance to get the tests going.

### DIFF
--- a/python/ecl/summary/ecl_sum_vector.py
+++ b/python/ecl/summary/ecl_sum_vector.py
@@ -106,7 +106,7 @@ class EclSumVector(object):
         warnings.warn("The mpl_dates property has been deprecated - use numpy_dates instead",
                      DeprecationWarning)
 
-        return parent.get_mpl_dates(report_only)
+        return parent.get_mpl_dates(self.report_only)
 
     @property
     def numpy_dates(self):
--- a/python/ecl/summary/ecl_sum_vector.py
+++ b/python/ecl/summary/ecl_sum_vector.py
@@ -46,7 +46,8 @@ class EclSumVector(object):
 
         self.__dates = parent.get_dates(report_only)
         self.__days = parent.get_days(report_only)
-        self.__mpl_dates = parent.get_mpl_dates(report_only)
+        self.__mpl_dates = parent.get_numpy_dates(report_only)
+        self.__numpy_dates = parent.get_numpy_dates(report_only)
         self.__report_step = parent.get_report_step(report_only)
         self.__values = None
 
@@ -99,8 +100,20 @@ class EclSumVector(object):
     def mpl_dates(self):
         """
         All the dates as numpy vector of dates in matplotlib format.
+        This property will be replaced by numpy_dates, but is kept for
+        backwards-compatibility for the time-being. Usage will trigger
+        a depreciation warning.
         """
-        return self.__mpl_dates
+        warning.warn("The mpl_dates property has been deprecated - use numpy_dates instead",
+                     DeprecationWarning)
+        return self.__numpy_dates
+
+    @property
+    def numpy_dates(self):
+        """
+        All the dates as numpy vector of dates in matplotlib format.
+        """
+        return self.__numpy_dates
 
     @property
     def report_step(self):
@@ -119,7 +132,7 @@ class EclSumVector(object):
         return EclSumNode(self.__report_step[index],
                           self.__days[index],
                           self.__dates[index],
-                          self.__mpl_dates[index],
+                          self.__numpy_dates[index],
                           self.__values[index])
 
 
--- a/python/ecl/summary/ecl_sum_vector.py
+++ b/python/ecl/summary/ecl_sum_vector.py
@@ -112,7 +112,7 @@ class EclSumVector(object):
     @property
     def numpy_dates(self):
         """
-        All the dates as numpy vector of dates in matplotlib format.
+        All the dates as numpy vector of dates in numpy format.
         """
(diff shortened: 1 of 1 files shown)


abrautomate:

python/ecl/summary/ecl_sum_vector.py
```diff
@@ -99,8 +100,20 @@ def days(self):
...
         """
-        return self.__mpl_dates
+        warning.warn("The mpl_dates property has been deprecated - use numpy_dates instead",
+                     DeprecationWarning)
+        return self.__numpy_dates
+
+    @property
+    def numpy_dates(self):
+        """
+        All the dates as numpy vector of dates in matplotlib format.
```
markusdregi: Are you sure that this is matplotlib format? When looking at the parent implementation it really seems to be `numpy` format..

python/ecl/summary/ecl_sum_vector.py
```diff
@@ -99,8 +100,20 @@ def days(self):
...
         """
         All the dates as numpy vector of dates in matplotlib format.
+        This property will be replaced by numpy_dates, but is kept for
+        backwards-compatibility for the time-being. Usage will trigger
+        a depreciation warning.
         """
-        return self.__mpl_dates
+        warning.warn("The mpl_dates property has been deprecated - use numpy_dates instead",
+                     DeprecationWarning)
+        return self.__numpy_dates
```
markusdregi: You are now changing the behaviour of the function you are deprecating. Should continue to return `self.__mpl_dates` until removed.

python/ecl/summary/ecl_sum_vector.py
```diff
@@ -46,7 +46,8 @@ def __init__(self, parent, key, report_only = False):
 
         self.__dates = parent.get_dates(report_only)
         self.__days = parent.get_days(report_only)
-        self.__mpl_dates = parent.get_mpl_dates(report_only)
+        self.__mpl_dates = parent.get_numpy_dates(report_only)
```
markusdregi: Should not be changed..

wouterjdb: This I did because I don't want to __init__ __mpl_dates with get_mpl_dates() as this will trigger the Depreciation warning.

Looking at it again I actually suggest to remove self.__mpl_dates completely (as far as I can see it is merely used as a placeholder for the @propery mpl_dates.

So instead I suggest that the @property mlp_dates calls 
`EclSum.get_mpl_dates(report_only)`.

This way the depreciation warning will not be called upon __init__ of the sum vector but only when someone requests the mpl_dates property.

python/ecl/summary/ecl_sum_vector.py
wouterjdb: Yes, I see. I did this to avoid having to init __mpl_dates. See next comment; that might be a better solution.

python/ecl/summary/ecl_sum_vector.py
wouterjdb: I see that EclSumNode also contains the mpl_date (I didn't spot this because I already changed this to numpy_dates). Within the EclSumNode libecl/python/ecl/summary/ecl_sum_node.py the mpl_date doesn't seem to be used.

Any idea if we can simply change this to numpy date as well or would that break other parts?[libecl/python/ecl/summary/ecl_sum_node.py](https://github.com/Statoil/libecl/blob/5a47e7406c944d0667d435e4f489094ba7c15732/python/ecl/summary/ecl_sum_node.py)

python/ecl/summary/ecl_sum_vector.py
```diff
@@ -99,8 +100,21 @@ def days(self):
...
         All the dates as numpy vector of dates in matplotlib format.
+        This property will be replaced by numpy_dates, but is kept for
+        backwards-compatibility for the time-being. Usage will trigger
+        a depreciation warning.
         """
-        return self.__mpl_dates
+        warning.warn("The mpl_dates property has been deprecated - use numpy_dates instead",
+                     DeprecationWarning)
+
+        return EclSum.get_mpl_dates(report_only)
```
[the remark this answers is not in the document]
wouterjdb: I saw the parent being used in the code yes. However, parent is not an input argument in the mpl_dates(), so guess doing that would fail?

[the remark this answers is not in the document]
markusdregi: No we keep it around as self.parent, so that one you can use freely.

joakim-hove:
Markus - should we merge the future work we discussed yesterday before this?

markusdregi:
I don’t think the order matters that much, so I’ll just make a PR when I’ve picked it up and then we’ll see what hits the finish line and gets merged first?

markusdregi:
Thank you for your contribution @wouterjdb!

markusdregi:
Argh, should of course have squashed those commits when I merged it. Anyway, great work on your side!

wouterjdb:
No problem - now it looks like I have done a lot of work. Good for my stats :laughing: Merry Christmas and a happy new year to all of you. :christmas_tree: :santa:

PR #533, opened 2018-11-13
NNC export support Intersect format

andreabrambilla:
The Intersect simulator stores NNC information differently wrt Eclipse. With this PR, libecl uses the correct NNC data when running the NNC export.

Note that this affects exclusively for the nnc export. `EclGrid` would still expose the wrong data if loaded from an Intersect file

See https://github.com/Statoil/libecl/issues/505

**Pre un-WIP checklist**

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -350,6 +350,7 @@ foreach (name   ecl_alloc_cpgrid
                 ecl_grid_init_fwrite
                 ecl_grid_reset_actnum
                 ecl_grid_ext_actnum
+                ecl_nnc_export_intersect
                 ecl_sum_data_intermediate_test
                 ecl_grid_cell_contains
                 ecl_unsmry_loader_test
@@ -446,7 +447,6 @@ foreach(name ecl_coarse_test
              ecl_grid_layer_contains
              ecl_restart_test
              ecl_nnc_export
-             ecl_nnc_export_intersect
              ecl_nnc_export_get_tran
              ecl_nnc_data_statoil_root
              ecl_sum_case_exists
@@ -520,8 +520,6 @@ add_test(NAME ecl_nnc_export4 COMMAND ecl_nnc_export ${_eclpath}/DualPoro/DUAL_D
 add_test(NAME ecl_nnc_export5 COMMAND ecl_nnc_export ${_eclpath}/DualPoro/DUALPORO TRUE)
 add_test(NAME ecl_nnc_export6 COMMAND ecl_nnc_export ${_eclpath}/nestedLGRcase/TESTCASE_NESTEDLGR TRUE)
 add_test(NAME ecl_nnc_export7 COMMAND ecl_nnc_export ${_eclpath}/TYRIHANS/BASE20150218_MULTFLT FALSE)
-
-add_test(NAME ecl_nnc_export_intersect COMMAND ecl_nnc_export_intersect)
 
 add_test(NAME ecl_nnc_export_get_tran COMMAND ecl_nnc_export_get_tran
     ${_eclpath}/Troll/MSW_LGR/2BRANCHES-CCEWELLPATH-NEW-SCH-TUNED-AR3)
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -446,6 +446,7 @@ foreach(name ecl_coarse_test
              ecl_grid_layer_contains
              ecl_restart_test
              ecl_nnc_export
+             ecl_nnc_export_intersect
              ecl_nnc_export_get_tran
              ecl_nnc_data_statoil_root
              ecl_sum_case_exists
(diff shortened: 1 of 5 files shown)


lib/CMakeLists.txt
```diff
@@ -520,6 +521,8 @@ add_test(NAME ecl_nnc_export5 COMMAND ecl_nnc_export ${_eclpath}/DualPoro/DUALPO
 add_test(NAME ecl_nnc_export6 COMMAND ecl_nnc_export ${_eclpath}/nestedLGRcase/TESTCASE_NESTEDLGR TRUE)
 add_test(NAME ecl_nnc_export7 COMMAND ecl_nnc_export ${_eclpath}/TYRIHANS/BASE20150218_MULTFLT FALSE)
 
+add_test(NAME ecl_nnc_export_intersect COMMAND ecl_nnc_export_intersect)
```
joakim-hove: The test is add as part of the loop further up in the file - the `add_test()` commands listed here are for the tests which require (unique) commandline arguments.

lib/ecl/tests/ecl_nnc_export_intersect.cpp
```diff
@@ -0,0 +1,121 @@
...
+   auto tran_kw = ecl_kw_alloc(TRANNNC_KW, INIT_NNC_NUM, ECL_DOUBLE);
+   for(auto i = 0; i < INIT_NNC_NUM; ++i) {
+      ecl_kw_iset_int(nnc1_kw, i, 1);
+      ecl_kw_iset_int(nnc2_kw, i, 2);
+      ecl_kw_iset_double(tran_kw, i, 2. * i);
+   }
+
+   // write to file directly using fortio
+   auto init_filename = util_alloc_tmp_file("/tmp", "ecl_nnc_export_intersect_init_file", false);
+   auto fortio = fortio_open_writer(init_filename, false, true);
```
joakim-hove: Instead of the final `true` you can use:
```
#include <ert/ecl/ecl_endian_flip.h>
...
...
auto fortio = fortio_open_writer(init_filename, false, ECL_ENDIAN_FLIP);
```

lib/ecl/tests/ecl_nnc_export_intersect.cpp
```diff
@@ -0,0 +1,121 @@
...
+   ecl_kw_fwrite(tran_kw, fortio);
+   fortio_fclose(fortio);
+
+   // reopen the file as an ecl file
+   auto out = ERT::ert_unique_ptr<ecl_file_type, ecl_file_close>(
+         ecl_file_open(init_filename, 0)
+         );
+
+   free(init_filename);
+   return out;
```
joakim-hove: The `nnc1_kw, nnc2_kw` and `tran_kw` variables are not disposed of - not that it matters very much; but my ocd would like to see:
```
ecl_kw_free( nnc1_kw );
ecl_kw_free( nnc2_kw );
ecl_kw_free( tran_kw );
```

joakim-hove:
Some minor comments - this can be merged when the internal tests pass.

lib/ecl/ecl_nnc_export.cpp
```diff
@@ -86,15 +122,22 @@ static int  ecl_nnc_export__( const ecl_grid_type * grid , int lgr_index1 , cons
...
+  if(ecl_nnc_intersect_format(grid, init_file)) {
+     // Intersect format
+     total_valid_trans = ecl_nnc_export_intersect__(init_file, nnc_data, &nnc_index);
+  }
+  else {
+    // Eclipse format
+    total_valid_trans = ecl_nnc_export__( grid , 0 , init_file , nnc_data , &nnc_index );
+    {
+      int lgr_index;
+      for (lgr_index = 0; lgr_index < ecl_grid_get_num_lgr(grid); lgr_index++) {
```
joakim-hove: The `int lgr_index`can be declared inside the loop.

lib/ecl/tests/ecl_nnc_export_intersect.cpp
```diff
@@ -0,0 +1,121 @@
...
+                                             Killing with SIGKILL (-9) will not give a backtrace.*/
+}
+
+
+ERT::ert_unique_ptr<ecl_grid_type, ecl_grid_free>
+make_intersect_grid() {
+   auto out = ERT::ert_unique_ptr<ecl_grid_type, ecl_grid_free>(
+         ecl_grid_alloc_rectangular(GRIDX, GRIDY, GRIDZ, 1., 1., 1., nullptr)
+         );
+   for(auto i = 0; i < GRID_NNC_NUM; ++i)
```
joakim-hove: This is at least invalid from a domain point of view; although the test might actually work. I would suggest something like:
```
for (auto i=0; i < GRID_NNC_NUM; i++)
     ecl_grid_add_self_nnc(out.get(), 2*i, 2*i + 1, i);
```

lib/ecl/tests/ecl_nnc_export.cpp
```diff
@@ -74,16 +74,18 @@ int count_kw_data( const ecl_file_type * file , ecl_grid_type * grid , const cha
 
 void test_count(const char * name) {
   char * grid_file_name = ecl_util_alloc_filename(NULL , name , ECL_EGRID_FILE , false  , -1);
+  char * init_file_name = ecl_util_alloc_filename(NULL , name , ECL_INIT_FILE , false  , -1);
   ecl_grid_type * grid = ecl_grid_alloc( grid_file_name );
   ecl_file_type * grid_file = ecl_file_open( grid_file_name , 0 );
+  ecl_file_type * init_file = ecl_file_open( init_file_name , 0);
```
joakim-hove: The internal equionor test fails because the `ecl_file_open(init_file_name)` returns `NULL` - because there is no such file. I have updated the testdata internally, so it should work again quite soon.

andreabrambilla: great, thanks!

joakim-hove: OK - then we merge!

andreabrambilla:
@joakim-hove is it ok to merge this now?

Issue #521, opened 2018-11-04
How to handle interpolated time outside range


joakim-hove:
This issue is about the `ecl_sum_get_interp()` family of functions. Assuming an eclipse summary case spans the time interval `[t0, t1]` if you call `xxx_get_interp()` with `t` &notin; `[t0,t1]` you will currently get an exception. 

As part of a desperate nexus fix here: https://github.com/Statoil/libecl/pull/506 and https://github.com/Statoil/libecl/pull/510 the `ecl_sum_resample_..` function was allowed to resample (i.e. interpolate) time values outside the interval - so right now the different interpolation functions are not consistent in this regard. We can:

1. Essentially revert https://github.com/Statoil/libecl/pull/506 and https://github.com/Statoil/libecl/pull/510 to recover the previous throw for &notin; behaviour.
2. We can let all interpolation accept &notin; - in that case rate values should be extrapolated with zero, all other values can be meaningfully extrapolated with first and last value for `t < t0` and `t > t1` respectively. 

I would vote for 2.

markusdregi:
In 2, would we then interpolate outside the `[t0, t1]` interval as a step function, as in for `t1 + eps` any rate would be 0? Or with a function that extremely quickly decreases towards 0?

In general I guess the very best would be if we could support both and then it would be up to the user to argue whether one would like something like this or not?

joakim-hove:
> In 2, would we then interpolate outside the [t0, t1] interval as a step function, as in for t1 + eps any rate would be 0? Or with a function that extremely quickly decreases towards 0?

Yes - step function.

> In general I guess the very best would be if we could support both and then it would be up to the user to argue whether one would like something like this or not?

Well - in a way we already do; the user can query for `t0` and `t1` and if the default behaviour of trivial extrapolation is not acceptable any alternative can be implemented. The point is that I *think* alternative 2 will have the highest satisfaction rate.

markusdregi:
How about changing the API of `resample` as follows:

```python
def resample(self,
        new_case_name,
        time_points,
        lower_interpolation=False,
        upper_interpolation=False,
        ):
```

And then, if you do not toggle the `interpolation` arguments, you get the same old behaviour with failure. If you set them as `True`, you get the value of `start_time` if `time < start_time` and the value of `end_time` if `end_time < time` (as implemented in the above PR's). Finally, if you set `lower_interpolation` (or `upper_interpolation`) as a numeric value, i.e. `1`, you get `1` back for every `time < start_time`.

[...]

PR #534, opened 2018-11-13
Support for extrapolation in the resample function.

xjules:
Handles #521 and relates to #506 and #510 

Currently, there are two additional bool parameters for ecl_sum_alloc_resample (lower_extrapolation and upper_extrapolation)
TODO: Need to make proper tests in cpp (ecl_sum_alloc_resampled_test.cpp) . The first test in python sketched (test_resample_extrapolate). 
- lower_extrapolation=True -> clamping to the first value for t < start_time or if isRate than derivative is 0
- upper_extrapolation=True -> clamping to the last value for t > end_time or if isRate than derivative is 0

**Pre un-WIP checklist**


xjules:
Need to finish tests also in cpp so not for merging yet!

lib/ecl/ecl_sum.cpp
```diff
@@ -779,39 +779,21 @@ const char * ecl_sum_get_general_var_unit( const ecl_sum_type * ecl_sum , const
 /*****************************************************************/
 
 
-ecl_sum_type * ecl_sum_alloc_resample(const ecl_sum_type * ecl_sum, const char * ecl_case, const time_t_vector_type * _times) {
+ecl_sum_type * ecl_sum_alloc_resample(const ecl_sum_type * ecl_sum, const char * ecl_case, const time_t_vector_type * _times, const bool lower_extrapolation, const bool upper_extrapolation) {
```
joakim-hove: No point in adding `const` to the input scalar parameters - in my opinion.

joakim-hove: Please rename `_times` &rightarrow `times` - the leading underscore was just a f... hack.

joakim-hove:
Looks good

python/ecl/summary/ecl_sum.py
```diff
@@ -1481,8 +1481,8 @@ def export_csv(self, filename, keys=None, date_format="%Y-%m-%d", sep=";"):
 
 
 
-    def resample(self, new_case_name, time_points):
-        new_case = self._resample(new_case_name, time_points)
+    def resample(self, new_case_name, time_points, lower_extrapolation=False, upper_extrapolation=False):
```
markusdregi: I this one burns as heavily as I believe in the C-code if you try to extrapolate with all extrapolation deactivated, I think it would be good to detect this case and raise a Python exception instead. That is also testable ;)

xjules: Currently testing only via jenkins (any suggestions how to execute these tests locally?). We can have a chat about this tomorrow.

joakim-hove: I don't really understand the question; but basically:
```
cmake .. -DBUILD_TESTS=ON -DENABLE_PYTHON=ON ...
make -j 8
ctest -j 8 ...
```

xjules: It crashes on this function:  
```ecl_sum_vector_type * ecl_sum_vector = ecl_sum_vector_alloc(ecl_sum, true);```
The testing question related to this decorator: @statoil_test() (i.e., skipping all the tests)

Comment (xjules) [reply python/ecl/summary/ecl_sum.py]: Update: interpolation works, just the starting time for the sampling interval was given by the input summary instead of the given time points.  Need to finish the test tomorrow and it'll be good to go.

**Comment by markusdregi** [comment]
Apparently the Python bindings does not work yet.. Will wait for that.

**Comment by xjules** [comment]
Update: python test fails when doing resampling on the provide `testcase - local/ECLIPSE/cp_simple3/SIMPLE_SUMMARY3`
In particular, it crashes on this function `ecl_sum_add_var` when copying node Nr. 205 `WWGR` from `cp_simple3` using `ecl_sum_add_smspec_node`.

**Comment by xjules** [comment]
PS: Which means that the problem is not related to python binding, but to actual testcase. Essentially, it is the following copying the fails to copy that node. Shall, we update the testcase?
```
for (int i = 0; i < ecl_smspec_num_nodes(ecl_sum->smspec); i++) {
    const smspec_node_type * node = ecl_smspec_iget_node_w_node_index(ecl_sum->smspec, i);
    ecl_sum_add_smspec_node(  ecl_sum_resampled, node );
  }
```

**Comment by joakim-hove** [comment]
> Essentially, it is the following copying code which fails to copy that node.

What do you mean with: "...*fails to copy*.."?

**Comment by xjules** [comment]
> > Essentially, it is the following copying code which fails to copy that node.
> 
> What do you mean with: "..._fails to copy_.."?
I'm running the following code to get the nodes that have successfully passed.
```
   const smspec_node_type * node = ecl_smspec_iget_node_w_node_index(ecl_sum->smspec, i);
    if (util_string_equal(smspec_node_get_keyword(node), "TIME"))
        continue;
    printf("copying node %d - %s\n",i,smspec_node_get_keyword(node));
    ecl_sum_add_smspec_node(  ecl_sum_resampled, node );
```
in `ecl_sum_add_smspec_node` this fails: 
```
ecl_sum_add_var(ecl_sum,
                         smspec_node_get_keyword(node),
                         smspec_node_get_wgname(node),
                         smspec_node_get_num(node),
                         smspec_node_get_unit(node),
                         smspec_node_get_default(node));
```

**Comment by xjules** [comment]
after some debugging, it fails in `ecl_smspec.cpp`
```
void ecl_smspec_insert_node(ecl_smspec_type * ecl_smspec, smspec_node_type * smspec_node)
```
Still, need to debug.

File: lib/ecl/ecl_sum.cpp [hunk 3]
```diff
@@ -821,35 +804,52 @@ ecl_sum_type * ecl_sum_alloc_resample(const ecl_sum_type * ecl_sum, const char *
...
-    /* Look up interpolated data in the original case. */
-    ecl_sum_get_interp_vector( ecl_sum, t, ecl_sum_vector, data);
+    time_t input_t = time_t_vector_iget(times, report_step);
+    if (input_t < start_time) {
+      //clamping to the first value for t < start_time or if it is a rate than derivative is 0
+      for (int i=1; i < ecl_smspec_num_nodes(ecl_sum->smspec); i++) {
+        double value = 0;
+        const smspec_node_type * node = ecl_smspec_iget_node_w_node_index(ecl_sum->smspec, i);
+        if (!smspec_node_is_rate(node))
+            value = ecl_sum_iget_first_value(ecl_sum,i-1);
```
joakim-hove: Instead of `i-1` here the code should use `smspec_node_get_params_index(node)` - they can be different.

PR #535, opened 2018-11-15
WIP: Well state to std

stefoss23:
**Task**
Well state paramerers converted to std.

**Pre un-WIP checklist**


lib/ecl/well_state.cpp
```diff
@@ -105,7 +109,7 @@ links are as follows:
   Segment1: C2
   Segment2: C3
   Segment3: C4, C5
-  Segment4: C6
+  Segment4: C6state_free(
```
joakim-hove: Some typo?

PR #526, opened 2018-11-09
smspec_node into class

stefoss23:
**Task**
Smspec node made into a class w/ member functions. The old Smspec wrapper class is obsolete and therefore removed. 

**Pre un-WIP checklist**
- [ ] Statoil tests pass locally

* Statoil/libres#438
* Statoil/ert#167

PR #538, opened 2018-11-18
Smspec c++

joakim-hove:
Continuation of #526

Statoil/libres#458
Statoil/ert#167

**Note:** An important lesson learned while working with this task was the following:

In the old C code a quite common pattern was the following:
```C
my_type * ptr = my_alloc( );

// Insert pointer in vector container, which will enable index based lookup. The 
// vector will alse take ownership of the object and discard it when going out of scope.
vector_insert_owned_ref( storage, ptr, my_free__);

// Insert pointer in hash table, to enable string based lookup. The hash table will not manage
// the underlying object.
hash_insert_ref(hash_table, "KEY", ptr);
```
This has worked reasonably well - when switching to C++ it was intially tempting to install value objects in a `std::vector<my_class>` instance, and then access pointers to those from a `std::map<my_class *>` - *that does not work*. The point is that when the `std::vector<my_class>` manages the instances they will not be in stable memory, and things fall apart. This pattern is quite common in librecl/libres!


lib/ecl/tests/ecl_smspec.cpp
```diff
@@ -23,28 +23,13 @@
 #include <ert/ecl/ecl_sum.hpp>
 #include <ert/ecl/ecl_smspec.hpp>
 
-void test_sort( ecl_smspec_type * smspec )
```
joakim-hove: Note: this test is removed because the whole `ecl_smspec_sort()` functionality is removed.

PR #539, opened 2018-11-19
Add Windows Install Instructions MSVS2017 and setup batch file

trhallam:
**Task**
Update the build/install instructions/guide for windows.

**Approach**
Remove old instructions, add new markdown instructions. 
Add example/usable batch file for install.

**Pre un-WIP checklist**
No other tests.


Windows_Setup.bat
```diff
@@ -0,0 +1,40 @@
...
+cmake --build . --config Release
+
+:: copy builds to install path
+XCOPY /Y lib\Release %INSTALLPATH%\lib\Release
+XCOPY /Y bin\Release %INSTALLPATH%\bin\Release
+cd lib
+
+FOR /d %%a in (python*) DO XCOPY %%a %INSTALLPATH%\lib\%%a /s /e /h /y
+
+cd ..\..
```
andreabrambilla: This last part confuses me a bit. I would expect users to run `nmake install` (or `ninja install` if they want things a bit faster). If that doesn't work, I'd say we have a bug and we should fix it, copying files directly seems to be a workaround

WINDOWS.md
```diff
@@ -0,0 +1,34 @@
+# Compiling and Installing **_libecl_** on Windows
+
+## Prerequisits:
+1. Python 2.7 or 3.x installation
+   - Download a python instalation https://www.python.org/ or a python environment solution such as Anaconda https://anaconda.org/
+   - Use the python package manager **pip** to install **_libecl_** requirements: 
+	  * numpy
+	  * pandas
+      * cwrap
```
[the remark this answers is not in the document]
andreabrambilla: One more minor thing. Try to use only spaces and no tabs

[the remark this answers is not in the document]
trhallam: Got it, I wasn't sure where to start with this PR. I just basically documented what I did and hopefully we can improve it to be more generic for the everyday user.  I'll make some changes as you suggest and perhaps we can go from there. See further comments below.

WINDOWS.md
```diff
@@ -0,0 +1,34 @@
...
+7. Open a python session and run `import ecl` to check that the install and paths are now set.
+
+## Notes:
+If you encounter build errors try to run the commands in Windows_Setup.bat sequentially. Objects may be missing form the path
+or you may encounter build errors.
+
+**_libecl_** build currently warns about compatability with Python3+ but functionality appears to be working at this time (Nov-2018).
+
+## FAQ:
+The dlopen() error in python can be fixed by running `pip uninstall cwrap` and then `pip install cwrap` in the shell.
```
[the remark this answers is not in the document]
trhallam: No, it is quite a weird one. Sometimes when I completed the build I would get this error when importing ecl. The uninstall install trick appears to fix it. If you follow the trace back it is to do with the load function in cwrap prototype class. I will try to recreate and report back.

WINDOWS.md
```diff
@@ -0,0 +1,34 @@
...
+
+3. Download and install the community version of Microsoft Visual Studio https://visualstudio.microsoft.com/downloads/ . 
+Testing has been successful with Visual Studio 2017. At a minimum **_libecl_** requires the VS Studio packages for cmake, msbuild, c and c++
+compilers (CLI.exe).
+
+4. Edit Windows_Setup.bat in the repository top directory.
+   - Edit the path definitions to match your install of MSVS and Python
+   - Edit the install location to your prefered location. 
+
+5. Run Windows_Setup.bat
```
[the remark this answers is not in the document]
trhallam: Understood, I did this for a few reasons.
Mainly, setting up all the paths is a pain. You could rely on the VS prompt to have the paths right but if your python environment is separate (as mine is) it is not easy to modify VS prompt paths to point at the correct Python env. It also captures the command line calls the user would have to make anyway. This comment applies to the post below as well. Perhaps this is better addressed through text commentary and left to the user to decide.

An alternative, is you could include a Cmake json file example for VS. I couldn't get libecl to build from within VS though.

Windows_Setup.bat
```diff
@@ -0,0 +1,40 @@
...
+set INSTALLPATH=C:\dev\ERT
+
+:: make build directory
+mkdir temp-build
+cd temp-build
+
+:: Codes for building ERT
+:: The cmake -G option made need to change if you have a newer/older version of Visual Studio
+cmake -G "Visual Studio 15 2017 Win64" -DBUILD_SHARED_LIBS=ON -DCMAKE_BUILD_TYPE=Release -DENABLE_PYTHON=ON -DERT_BUILD_CXX=OFF -DBUILD_APPLICATIONS=ON ..
+cmake --build . --config Release
```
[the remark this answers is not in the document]
trhallam: Totally agree but wasn't sure how to capture this. The build options are common with the .appveoyer task. Do we know which build options do not work for Windows?

Windows_Setup.bat
trhallam: You can set the install path for the binary builds but the ?PYTHONPREFIX? option does not work under Windows so the ecl site-packages folder ends up in an obscure part of the build folder structure. I didn't experiment with _nmake_ or _ninja_, would these resolve that issue?

We could also resolve most of these comments by removing the bat file. It sounds like I wasn't using the best approach to building **libecl** under windows but I struggled to get other methods to work. Sorry for my ignorance with c compilers.

Windows_Setup.bat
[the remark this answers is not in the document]
jokva: > The paths and the cmake flags used here are arbitrary and specific to your installation, I don't think we should include them in the repo

Kinda, but not really. These are generic options to build with a visual studio generator, and not overly specific to trhallam's system.

Windows_Setup.bat
andreabrambilla: I may be wrong `nmake` and `ninja` would perform pretty much the same operations as building the `INSTALL` target in visual studio, so I guess this might actually be a bug in libecl. I'll try to reproduce it and open a dedicated issue if needed

WINDOWS.md
```diff
@@ -0,0 +1,30 @@
+# Compiling and Installing **_libecl_** on Windows
+
+## Prerequisits:
+1. Download or clone the **_libecl_** repository to your local disk.
+
+2. Python 2.7 or 3.x installation
+   - Download a python instalation https://www.python.org/ or a python environment solution such as Anaconda https://anaconda.org/
+
+3. In command prompt, navigate to the **_libecl_** source directory you created in step 1. Use the python package manager **pip** to install **_libecl_** requirements via `pip install -r requirements.txt`
```
andreabrambilla: Instead of _command prompt_ here we can suggest the user to open a suitable MSVC command prompt. For instance, with VS17, the entry _x64 Native Tools Command Prompt fo VS 2017_ can be found somewhere in the Start.
If we do this here, we can remove point `5` further down

Also, we could write something like "Make sure `python` can be executed and that your python environment is set up correctly (in case you are using Anaconda or any other environment manager)

WINDOWS.md
```diff
@@ -0,0 +1,30 @@
...
+    
+4. Download and install the community version of Microsoft Visual Studio https://visualstudio.microsoft.com/downloads/ .
```
andreabrambilla: I would remove `the community version of`

WINDOWS.md
```diff
@@ -0,0 +1,30 @@
...
+   - Download a python instalation https://www.python.org/ or a python environment solution such as Anaconda https://anaconda.org/
+
+3. In command prompt, navigate to the **_libecl_** source directory you created in step 1. Use the python package manager **pip** to install **_libecl_** requirements via `pip install -r requirements.txt`
+    
+4. Download and install the community version of Microsoft Visual Studio https://visualstudio.microsoft.com/downloads/ . 
+Testing has been successful with Visual Studio 2017. At a minimum **_libecl_** requires the VS Studio packages for cmake, msbuild, c and c++ compilers (CL.exe).
+
+5. From the command prompt in step 3 one can activate the _Win64_ MSVC environment by running **vcvars64.bat**. This can be found in the MSVC install directory {PATH TO INSTALL}\Microsoft Visual Studio\2017\Community\VC\Auxiliary\Build\
+
+6. Execute the build commands with the desired CMAKE parametres from `README.md`. The cmake generator should be Ninja.
```
andreabrambilla: I did not mean that we should push people to use `ninja` (well I'd say we should, but that's only my personal opinion).
An example is definitely good to have, and for that it may be worth using the generator for VS, as that's what most people will probably use. If anyone wants to play with cmake options, he/she probably can extrapolate enough info from that example

trhallam: Understood, I can put it back to VS generator but that example does not work for me. I cannot get it to acknowledge the `CMAKE_INSTALL_PREFIX` where as `ninja` works fine.

Windows_Setup.bat
trhallam: Apologies, I must have been doing something wrong. I got the VS generator to build and install correctly. I've updated the example in the markdown file to reflect a VS example. Hopefully this rounds us out.

WINDOWS.md
```diff
@@ -0,0 +1,33 @@
...
+4. Open a MSVC command prompt such as _x64 Native Tools Command Prompt for VS 2017_ from your start menu. In the open prompt, navigate to the **_libecl_** source directory you created in step 1. Use the python package manager **pip** to install **_libecl_** requirements via `pip install -r requirements.txt`. If Python is not accessible from the prompt it may be necessary to add the Python environment location to your system path variable `PATH`. 
+    
+5. Execute the build commands with the desired CMAKE parameters from `README.md`. The cmake generator can be _Ninja_ or and appropriate version of VSMC. For the availble options type `cmake -G` in the MSVC command prompt.  
+
+   An example build and install is provided below where %VARIABLE% are user defined directory paths:
+~~~~
+    cmake -G "Visual Studio 15 2017 Win64" -DCMAKE_INSTALL_PREFIX=%INSTALLPATH% -DBUILD_SHARED_LIBS="ON"  -DENABLE_PYTHON="ON"    -DCMAKE_BUILD_TYPE="Release" %SOURCEPATH%
+    cmake --build %BUILDPATH% --config Release --target install
+~~~~
+6. For **_libecl_** to be accessable in Python the `%INSTALLPATH%` and Python subdirectories must be added to the `PATH` and `PYTHONPATH` variables.
```
andreabrambilla: `accessable` -> `accessible`

Also, let's specify that `%INSTALLDIR%\lib\pythonX.Y\site-package`must be in the `PYTHONPATH`

trhallam:
@andreabrambilla Thanks for your help Andrea. I have pushed through that last round of changes. I played around with `cmake -G "Visual Studio 15 2017 Win64"` a bit more last night and manged to get it to work right. Not sure what I was  doing wrong before. Anyway, hopefully this helps the next person who doesn't know what they are doing either.

Are you happy to leave in the comment about `dlopen()` ? I haven't had time to try and recreate that issue and it may be due to something else I was doing while I experimented with different build commands.

trhallam:
@andreabrambilla I was working on my master so the only way I could work out how to do it was with `rebase`. Hope that's OK. Should be good to roll in now.

andreabrambilla:
That's perfectly fine, thanks once again :+1:

joakim-hove:
Thank you both for the effort!

PR #540, opened 2018-11-20
Well ts cpp

stefoss23:
**Task**
Well_ts converted to c++ std.


lib/ecl/well_ts.cpp
```diff
@@ -100,41 +104,23 @@ static well_node_type * well_node_alloc( well_state_type * well_state) {
...
-  const well_node_type * node2 = well_node_safe_cast_const( arg2 );
-
-  if (node1->sim_time < node2->sim_time)
-    return -1;
-  else if (node1->sim_time == node2->sim_time)
-    return 0;
-  else
-    return 1;
 
+static bool well_node_time_cmp( const well_node_type * node1 , const well_node_type * node2) {
```
joakim-hove: Since this now implements the `<` test it should be renamed `well_node_time_lt`.

PR #551, opened 2018-12-19
Update the implementation to identify region-to-region variables

joakim-hove:
The type of the summary variables is inferred by looking at the string, e.g. all variables like `WWCT` and `WOPT` which start with `W` are well variables.

Variables starting with 'R' are region properties, but for region properties there is a sub-category of variables which describe region-to-region properties. This PR updates the code to identify region-to-region properties. Previously the algorithm to detect a region to region variable was just to check the third character - and all variables with `var[2] == 'F'`were classified as region to region variables.

This PR will change the classification in two ways:

1. It is stricter - in addition to third character we *also check the fourth character* - and require that `var[3] == 'T' || var[3] == 'R'`.

2. It is more lenient - we also check for the substring `FT` or `FR` one character further out in the string - i.e. starting at the fourth character in var.

### DIFF
--- a/lib/ecl/smspec_node.cpp
+++ b/lib/ecl/smspec_node.cpp
@@ -164,28 +164,61 @@ ecl_smspec_var_type ecl::smspec_node::identify_var_type(const char * var) {
     case('R'):
       {
         /*
-          The distinction between region-to-region variables and plain
-          region variables is less than clear: The current
-          interpretation is that the cases:
+          The distinction between regular region variables and region-to-region
+          variables is less than clear. The manual prescribes a rule based on
+          the characters at position 3 and 4 or 4 and 5.
 
-             1. Any variable matching:
+             R*FT*  => Region to region
+             R**FT* => Region to region
+             R*FR*  => Region to region
+             R**FR* => Region to region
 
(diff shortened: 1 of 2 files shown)


joakim-hove:
An error message from Jenkins would have been very much appreciated ...

joakim-hove:
There is clearly an issue with this - would be very grateful if you could post the error message from the failing test(s).

markusdregi:
Sorry for the delayed responses. I'm still catching up after Christmas.

The easy answer (for me) is that this is the output from Jenkins:
> 114/275 Test #118: ecl_region2region ........................................................***Failed    0.59 sec
/tmp/tmp.ki8DPBweDc/libecl/lib/ecl/tests/ecl_region2region.cpp:27 => assert( true ) failed

However, I don't know how helpful that is.

joakim-hove:
OK; now the tests pass - which is a good thing. This PR updates the region-2-region classification according to the manual, that I am certain is correct - however the internal tests explicitly tested the old region - 2 - region detection algorithm; and therefor the old detection is also retained. This is explained in the comment in the code; the current PR is my "best shot" - but tricky...

PR #553, opened 2018-12-21
Unconditionally initialize the nums vector for all keywords

joakim-hove:
This comes from: https://github.com/OPM/opm-simulators/pull/1685

The summary values - as defined in the SMSPEC files - are composed of three parts: 

1. The keyword part - stored in the `KEYWORDS` vector.
2. The well/group name stored in the `WGNAME` vector.
3. An Integer value stored in the `NUMS` vector - exactly how this integer value should be interpreted depends on the keyword part.

Depending on the keyword value the corresponding `WGNAME` and `NUMS` values might - or might not be relevant. E.g the keyword `WWCT:OP_X` is a combination of the keyword `WWCT` and the `WGNAME` value `OP_X` - in this case the `NUMS` value is completely ignored. Another variety is the region pressure in region 7: `RPR:7`- this is a combination of the keyword `RPR` and the `NUMS` value 7 - in this case the `WGNAME`element is ignored.

As described in the OPM issue the `nums` value was not explicitly initiialized for situations where it was not used, leading to random unitiialized values in the `SMSPEC` files. With this PR the `smspec_node->num` value is unconditionally set - the change at line 806 is the relevant fix.

### DIFF
--- a/lib/ecl/smspec_node.cpp
+++ b/lib/ecl/smspec_node.cpp
@@ -71,6 +71,8 @@ static const char* special_vars[] = {"NEWTON",
                                      "WNEWTON"};
 
 
+static const int nums_unused = 0;
+
 /**
    This struct contains meta-information about one element in the smspec
    file; the content is based on the smspec vectors WGNAMES, KEYWORDS, UNIT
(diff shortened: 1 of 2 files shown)


joakim-hove:
jenkins build this please

markusdregi:
test this please

markusdregi:
Well spotted!

PR #554, opened 2018-12-22
Check for pandas function assert_frame_equal

joakim-hove:
Pandas has a function `assert_frame_equal` which is used unconditionally in a test. The `assert_frame_equal` function is only in quite new versions of Pandas, in this PR it is checked whether we have the function.

### DIFF
--- a/python/tests/ecl_tests/test_sum.py
+++ b/python/tests/ecl_tests/test_sum.py
@@ -23,7 +23,16 @@ import shutil
 import cwrap
 import stat
 import pandas
-from pandas.testing import assert_frame_equal
+
+def assert_frame_equal(a,b):
+    if not a.equals(b):
+        raise AssertionError("Expected dataframes to be equal")
+
+try:
+    from pandas.testing import assert_frame_equal
+except ImportError:
+    pass
+
 from contextlib import contextmanager
 from unittest import skipIf, skipUnless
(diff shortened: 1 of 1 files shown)


SteinarFoss:
Is there not already a check for pandas version in the cmake code?

python/tests/ecl_tests/test_sum.py
```diff
@@ -23,7 +23,12 @@
 import cwrap
 import stat
 import pandas
-from pandas.testing import assert_frame_equal
+
+try:
+    from pandas.testing import assert_frame_equal
+except ImportError:
+    assert_frame_equal = None
```
markusdregi: I'm a bit uncomfortable with skipping the test for older pandas versions. The `DataFrame.equals` method is much older. How about defining `assert_frame_equal` as :
```python
def assert_frame_equal(a, b):
    if not a.equals(b):
        raise AssertionError('Expected dataframes to be equal')

In case it is missing?
```

joakim-hove: Good idea - done!

markusdregi:
Can you try to merge this one @joakim-hove? I think it should work now...

joakim-hove:
> I think it should work now...

Looks like it - thank you!

PR #558, opened 2019-01-03
Build doc

joakim-hove:
The documentation in libecl has for a long time been neglected / dormant; in this PR the building of the documentation has been cleaned up significantly; see the commit message,

There are no changes to the actual documentation as part of thiis PR.


joakim-hove:
Thank you.

PR #561, opened 2019-01-10
Escape [ and ] in order to get rid of FutureWarning

### LINKED ISSUE #560: FutureWarning issued by EclNPV re.compile
The current [re.compile command](https://github.com/Statoil/libecl/blob/08272b1065691f847421c3ada5f461268c563a0d/python/ecl/summary/ecl_npv.py#L139) in `EclNPV` gives `FutureWarning: Possible nested set at position 1` on Python 3.7.

anders-kiaer:
**Issue**
Resolves #560 

**Approach**
Solved by escaping `[` and `]` where they are part of the character set to match, and not to be considered regular expression metacharacters.

More information regarding why the `FutureWarning` has been implemented:

- https://github.com/python/cpython/commit/05cb728d68a278d11466f9a6c8258d914135c96c#diff-8721d1362294720bad1dbbbaa4cc06e3
- https://bugs.python.org/issue30349

### DIFF
--- a/python/ecl/summary/ecl_npv.py
+++ b/python/ecl/summary/ecl_npv.py
@@ -138,6 +138,7 @@ class NPVPriceVector(object):
 class EclNPV(object):
     sumKeyRE = re.compile("[\[]([\w:,]+)[\]]")
 
+
     def __init__(self, baseCase):
         sum = EclSum(baseCase)
         if sum:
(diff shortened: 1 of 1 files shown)


abrautomate:

markusdregi:
Ok to test

PR #562, opened 2019-01-15
Convert TestArea to Cxx

joakim-hove:
The `TestArea` which is used to create and populate a random directory in `/tmp` is extremely well suited for C++ automatic destructors. 

~~A small initial step which should be merged first is here: #563~~

**Update:** OK - now I consider this ~complete. The current implementation has both a C++ and a C api, the C api must be retained for Python compatibility, and also for external dependencies in libres/ert/opm. The current *implementation* is a quite ugly mix of `const char *` and `std::string` - but the API is at least consistent. Have only converted one actual test to use this; will make a followup PR changing all the tests when this is merged.

~~Please note that will *not* be green on Travis before:  Statoil/ert#233 is merged.~~

### DIFF
--- a/lib/include/ert/util/test_work_area.hpp
+++ b/lib/include/ert/util/test_work_area.hpp
@@ -20,33 +20,58 @@
 #ifndef ERT_TEST_WORK_AREA_H
 #define ERT_TEST_WORK_AREA_H
 
+#include <string>
+
+namespace ecl {
+namespace util {
+
+class TestArea {
+public:
+    TestArea(const std::string& test_name, bool store_area = false);
+    ~TestArea();
+    const std::string& test_cwd() const;
+    const std::string& original_cwd() const;
+
+    void copy_directory(const std::string input_directory) const;
+    void copy_directory_content(const std::string input_directory) const;
+    bool copy_parent(const std::string input_path) const;
+    bool copy_parent_content(const std::string original_path) const;
+
+    void copy_file(const std::string& input_src_file) const;
+    std::string original_path(const std::string& input_path) const;
+
+private:
+    bool store;
+    std::string cwd;
+    std::string org_cwd;
+};
+
+
+}
+}
+
+typedef ecl::util::TestArea test_work_area_type;
+
 #ifdef __cplusplus
 extern "C" {
 #endif
 
 #include <stdbool.h>
 
-#include <ert/util/type_macros.hpp>
-
-  typedef struct test_work_area_struct test_work_area_type;
-
   char                * test_work_area_alloc_input_path( const test_work_area_type * work_area , const char * input_path );
   test_work_area_type * test_work_area_alloc(const char * test_name );
+  test_work_area_type * test_work_area_alloc__(const char * test_name, bool store_area );
   test_work_area_type * test_work_area_alloc_relative(const char * prefix , const char * test_path);
-  void                  test_work_area_set_store( test_work_area_type * work_area , bool store);
 
   void                  test_work_area_free(test_work_area_type * work_area);
   const char          * test_work_area_get_cwd( const test_work_area_type * work_area );
   const char          * test_work_area_get_original_cwd( const test_work_area_type * work_area );
-  void                  test_work_area_install_file( test_work_area_type * work_area , const char * input_src_file );
-  void                  test_work_area_copy_directory( test_work_area_type * work_area , const char * input_directory);
-  void                  test_work_area_copy_directory_content( test_work_area_type * work_area , const char * input_directory);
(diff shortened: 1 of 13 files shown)


joakim-hove:
Currently the cross-module building in travis does not work, because the organisation has changed name from `Statoil`to `equinor` - are you interested in patches to the `build_total.py` script, or are you heading in a different direction for building on Travis?

joakim-hove:
This can (and should ....) be more polished; but on a general note - do you see yourself merging this eventually?

markusdregi:
> This can (and should ....) be more polished; but on a general note - do you see yourself merging this eventually?

Yes, I do :+1:

markusdregi:
> or are you heading in a different direction for building on Travis?

Yes, we are moving towards building against specific releases and detect master breakage with nightly  _Jenkins_ builds instead... So the `build_total.py` is to go away and should not be patched. This style of developments suits the team better and it also makes the process around _ERT_ and _Everest_ the same (which makes life less complicated).

joakim-hove:
>So the build_total.py is to go away and should not be patched

OK - but then please note that this current PR will not be green on Travis before the PR in ert is merged; that can be merged right away through.

lib/include/ert/util/test_work_area.hpp
```diff
@@ -38,15 +79,12 @@ extern "C" {
   void                  test_work_area_free(test_work_area_type * work_area);
   const char          * test_work_area_get_cwd( const test_work_area_type * work_area );
   const char          * test_work_area_get_original_cwd( const test_work_area_type * work_area );
-  void                  test_work_area_install_file( test_work_area_type * work_area , const char * input_src_file );
-  void                  test_work_area_copy_directory( test_work_area_type * work_area , const char * input_directory);
-  void                  test_work_area_copy_directory_content( test_work_area_type * work_area , const char * input_directory);
-  void                  test_work_area_copy_file( test_work_area_type * work_area , const char * input_file);
-  bool                  test_work_area_copy_parent_directory( test_work_area_type * work_area , const char * input_path);
-  bool                  test_work_area_copy_parent_content( test_work_area_type * work_area , const char * input_path);
-  void                  test_work_area_sync( test_work_area_type * work_area);
```
markusdregi: Appears that you are removing the sync-function from the C-API, the tests and the usage of it, but still exposing it in the `C++`-api?

lib/include/ert/util/test_work_area.hpp
```diff
@@ -38,15 +79,12 @@ extern "C" {
   void                  test_work_area_free(test_work_area_type * work_area);
   const char          * test_work_area_get_cwd( const test_work_area_type * work_area );
   const char          * test_work_area_get_original_cwd( const test_work_area_type * work_area );
-  void                  test_work_area_install_file( test_work_area_type * work_area , const char * input_src_file );
```
markusdregi: Isn't this function conflicting a bit with the idea of a `tmp` test area? It does not seems to be much used either. What would you think about keeping it out of the C++ API?

lib/include/ert/util/test_work_area.hpp
joakim-hove: OK - completely removed.

lib/include/ert/util/test_work_area.hpp
joakim-hove: The purpose of this functionality is to copy a file - with leading directory components, from somewhere down to the testarea. It is used in a couple of tests in libres; would prefer to leave it - at least until the tests in libres can be modified.

markusdregi:
Dear Joakim. Sorry for the long silence on this one! I'll try to make the remainder of the process a bit quicker...

I think this is a nice cleanup of the test frame work, I would however like to use the opportunity to shrink the C++ API a bit, as it seems like some of the functions are (almost) never used.

I suggest the following changes to the publig C++ API:
 - rename `org_cwd` to `original_cwd`, I at least found the short name confusing
 - I'm still a bit skeptical to the `install_file`-method, in particular the absolute path behavior. How about having the C++ implementation raise an exception in this case and then patch this to get the old behavior in the `C` api?
  - Remove the `set_store`
  - Change the name of `input_path`. It is not clear to me what this means until I look at the implementation. I struggled a bit to find a good suggestion, so you're welcome to come up with a name, the best I could find was `original_path`.

Again, sorry for the delay. I'm hoping for a good and rapid discussion from this point on... :+1:

markusdregi:
When we agree on the comments above, I'm ready to merge this one.

joakim-hove:
> I'm hoping for a good and rapid discussion from this point on...

Good - I am not able to complete this right away; but I will try to address your comments this afternoon.

joakim-hove:
OK - I think all comments should be addressed.

PR #570, opened 2019-01-28
Travis testing

markusdregi:
**Issue**
Resolves #<Issue id>

**Approach**


jokva:
Why export/install both dist- and site-packages? And what happened to the python 3.6 variable that's set, but since then ignored?

lars-petter-hauge:
Ah, this is better!

I understood the reasons for the other setup, but at this point it just brought more pain than gain

markusdregi:
> Why export/install both dist- and site-packages?

Because the _macOS_ builds and the _linux_ builds installs to _site-packages_ and _dist-packages_ respectively. I could of course have added them conditionally, but I don't really see the benefit. 

> And what happened to the python 3.6 variable that's set, but since then ignored?
It was to be used to deduce the elements that was to be added to the `PYTHONPATH`, but it turned out that _libecl_ ignores the current Python version, and hence I had to always add `..../lib/python2.7/...`. I made an issue for this: #571

.travis.yml
```diff
@@ -9,13 +9,14 @@ os:
...
-sudo: false
+sudo: true
 dist: trusty
 
 env:
   global:
     - ERT_SHOW_BACKTRACE=1
-    - LD_LIBRARY_PATH="$(pwd)/install/lib64"
+    - INSTALL_DIR="$(pwd)/install"
+    - LD_LIBRARY_PATH="$INSTALL_DIR/lib64"
```
joakim-hove: The library suffix - i.e. `lib64` in this case - depends on platform, it is `lib64` on 64 bit RedHat but e.g. `lib` on Ubuntu. CMake wise this is set [here](https://github.com/equinor/libecl/blob/master/cmake/Modules/UseMultiArch.cmake), so for the current travis environment I do not think the `LD_LIBRARY_PATH=$INSTALL_DIR/lib64` will be correct.

markusdregi: > OS/Arch:      linux/amd64

So actually I think we're good :) But I'll add the `lib` version as well to the path.

PR #577, opened 2019-02-01
Fix code owners

markusdregi:
**Issue**
Apparently the syntax of the `.github/CODE_OWNERS` file was wrong and hence only the last one, being @jokva was interpreted. This PR should resolve that issue, more info in: https://help.github.com/articles/about-code-owners/

PR #578, opened 2019-02-01
Add xjules as code owner

markusdregi:
**Issue**
Test of #577. Add xjules as a code owner


markusdregi:
Cool, seems to be working now 👍

PR #579, opened 2019-02-04
Don't Redeclare 'errno'

bska:
Not all platforms implement 'errno' as a global object.  Since we already include `<errno.h>`, there is no need to explicitly declare the object itself.  This, for instance, fixes a linkage problem when
building ResInsight on Windows (MSVC) in the context of ResInsight:
```
2>c:\users\bska\documents\projects\software\repos\opm\resinsight\thirdparty\ert\lib\ecl\fortio.c(33): warning C4273: '_errno': inconsistent dll linkage
```

**Approach**
_Remove a potentially conflicting object declaration_

### DIFF
--- a/lib/ecl/fortio.c
+++ b/lib/ecl/fortio.c
@@ -30,8 +30,6 @@
 
 #define FORTIO_ID  345116
 
-extern int errno;
-
 /**
 The fortio struct is implemented to handle fortran io. The problem is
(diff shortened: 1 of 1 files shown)


abrautomate:

markusdregi:
Test this please

markusdregi:
Will get back to you when Jenkins finishes. This looks good :+1:

PR #586, opened 2019-03-20
WIP: Prepare to replace @TYPE@_vector with std::vector<T>

joakim-hove:
**Issue**
The libecl codebase has home-made vector which has some familiarity with `std::vector<T>`. The most important difference is what happens when out of bounds indices are accessed; for `std::vector<T>` accessing out of bounds indices is a bug and will result in undefined behavior. The  `@TYPE@_vector` is created *with* a default value, reading out of bounds will return the default value and writing to out of bounds values will automatically grow the vector.

**Approach**
For this work I have just added `util_abort()` call (not committed) where the automatic growth is invoked, and then rewritten the failing code. 

Long way to go!


markusdregi:
Closed due to inactivity. Feel free to reopen it if you want to continue the effort.

PR #588, opened 2019-03-20
Use C++ containers in the ecl_grav code

joakim-hove:
Replaced `hash_type`from libecl with `std::map`and `vector_type`with `std::vector` in the gravity related code.

### DIFF
--- a/lib/ecl/ecl_grav.cpp
+++ b/lib/ecl/ecl_grav.cpp
@@ -20,9 +20,11 @@
 #include <math.h>
 #include <stdbool.h>
 
+#include <unordered_map>
+#include <vector>
+#include <string>
+
 #include <ert/util/util.h>
-#include <ert/util/hash.hpp>
-#include <ert/util/vector.hpp>
 
 #include <ert/ecl/ecl_kw.hpp>
 #include <ert/ecl/ecl_util.hpp>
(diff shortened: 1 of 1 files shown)


lib/ecl/ecl_grav.cpp
```diff
@@ -530,29 +521,25 @@ static ecl_grav_survey_type * ecl_grav_survey_alloc_RFIP(ecl_grav_type * ecl_gra
...
+
 static double ecl_grav_survey_eval( const ecl_grav_survey_type * base_survey,
                                     const ecl_grav_survey_type * monitor_survey ,
                                     ecl_region_type * region ,
                                     double utm_x , double utm_y , double depth, int phase_mask) {
-  int phase_nr;
   double deltag = 0;
-  for (phase_nr = 0; phase_nr < vector_get_size( base_survey->phase_list ); phase_nr++) {
-    ecl_grav_phase_type * base_phase    = (ecl_grav_phase_type*)vector_iget( base_survey->phase_list , phase_nr );
+  for (int phase_nr = 0; phase_nr < static_cast<int>(base_survey->phase_list.size()); phase_nr++) {
```
markusdregi: Should this not be the other way around, changing the type of `phase_nr` instead?

lib/ecl/ecl_grav.cpp
```diff
@@ -673,23 +656,27 @@ void ecl_grav_new_std_density( ecl_grav_type * grav , ecl_phase_enum phase , dou
    density to 0.75, but in PVT regions 2 and 7 the density is
    different:
 
-      ecl_grav_new_std_density( grav , ECL_GAS_PHASE , 0.75 );
+      ecl_grav_new_std_density( grav , ECL_GAS_PHASE , 0.75 )
```
markusdregi: ?

markusdregi:
Had two small comments, besides that, this looks good :+1:

lib/ecl/ecl_grav.cpp
joakim-hove: Yes agree - fixed.

lib/ecl/ecl_grav.cpp
joakim-hove: Reverted

joakim-hove:
OK - this should be fixed!

PR #590, opened 2019-03-21
Use std::map instead of hash_type in ecl_grid

joakim-hove:
The final `hash_type` &rightarrow; `std::map<>`conversion.

If all of #588, #589, #590 are merged the liobecl library will have no internal users of the `hash_type` (it is still used in libres though).

### DIFF
--- a/lib/ecl/ecl_grid.cpp
+++ b/lib/ecl/ecl_grid.cpp
@@ -21,12 +21,14 @@
 #include <stdio.h>
 #include <stdbool.h>
 #include <math.h>
+
 #include <vector>
+#include <unordered_map>
+#include <string>
 
 #include <ert/util/util.h>
 #include <ert/util/double_vector.hpp>
 #include <ert/util/int_vector.hpp>
-#include <ert/util/hash.hpp>
 #include <ert/util/vector.hpp>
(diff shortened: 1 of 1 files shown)


lib/ecl/ecl_grid.cpp
```diff
@@ -5290,7 +5287,7 @@ ecl_grid_type * ecl_grid_get_lgr(const ecl_grid_type * main_grid, const char * _
   __assert_main_grid( main_grid );
   {
     char * lgr_name          = util_alloc_strip_copy( __lgr_name );
-    ecl_grid_type * lgr_grid = (ecl_grid_type*)hash_get(main_grid->LGR_hash , lgr_name);
+    ecl_grid_type * lgr_grid = main_grid->LGR_hash.at(lgr_name);
```
andreabrambilla: Note that the `at` method throws if the element is not found. Is this the intended behaviour?
Alternatively you could use `find` which returns an iterator

joakim-hove:
Thank you for the review.

> Depending on the required performance, you my consider also unordered_map instead of map

This map will be checked O(1) times - so from a performance perspective it really should not make a difference; but when the sorted behavior of `std::map` is not called for I agree that `std::unordered_map` should be used - changing.

PR #593, opened 2019-03-24
Mem leaks - tests

joakim-hove:
These are memory leaks found in tests using a combination of valgrind and compiler option `-fsanitize=address`. A related PR with memory leaks/undefined behavior in the library code is here: #594

### DIFF
--- a/lib/util/tests/ert_util_strcat_test.cpp
+++ b/lib/util/tests/ert_util_strcat_test.cpp
@@ -28,8 +28,11 @@ void test_strcat(char * s1 , const char *s2 , const char * expected) {
   char * cat = util_strcat_realloc(s1 , s2 );
   if (test_check_string_equal( cat , expected ))
     free( cat );
-  else
-    test_error_exit("util_strcat_realloc(%s,%s) Got:%s  expected:%s \n",s1,s2,cat , expected);
+  else {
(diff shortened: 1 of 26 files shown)


lib/util/tests/ert_util_strcat_test.cpp
```diff
@@ -28,8 +28,10 @@ void test_strcat(char * s1 , const char *s2 , const char * expected) {
   char * cat = util_strcat_realloc(s1 , s2 );
   if (test_check_string_equal( cat , expected ))
     free( cat );
-  else
-    test_error_exit("util_strcat_realloc(%s,%s) Got:%s  expected:%s \n",s1,s2,cat , expected);
+  else {
+    fprintf(stderr, "util_strcat_realloc(%s,%s) Got:%s  expected:%s \n",s1,s2,cat , expected);
+    exit(1);
```
andreabrambilla: I am not sure I fully understand this change

joakim-hove: Hehe - that I can understand - the change was not even complete. The point was that the `expected` pointer leaked because the `test_error_exit( )` function terminated with a valid pointer, I guess the sanitizer stopped complaining when the `exit(1)` was "nearby"(?) - anyway I have now also added the missing `free( expected )`.

joakim-hove:
Thank you for the reviews.

PR #594, opened 2019-03-25
Mem leaks library

joakim-hove:
These are memory leaks & undefined behaviors found with a combination of valgrind and compiler flag `-fsanitize=address`

### DIFF
--- a/lib/util/buffer.cpp
+++ b/lib/util/buffer.cpp
@@ -563,12 +563,36 @@ void buffer_replace_string( buffer_type * buffer , size_t offset , size_t old_si
    touching internal state.
 */
 
+namespace {
+
+  /*
+    This homemade strstr() implementation is used here because we can not
+    guarantee that the buffer->data is '\0' terminated and then normal strstr()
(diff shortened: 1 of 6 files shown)


lib/util/buffer.cpp
```diff
@@ -563,16 +563,42 @@ void buffer_replace_string( buffer_type * buffer , size_t offset , size_t old_si
...
 
   if (strlen(expr) > 0) {
-    char * match_ptr = strstr( &buffer->data[buffer->pos] , expr );
+    const char * match_ptr = memcmp_strstr( &buffer->data[buffer->pos], buffer->content_size - buffer->pos, expr );
     if (match_ptr) {
       buffer->pos += match_ptr - &buffer->data[buffer->pos];
       match = true;
     }
+    if (match_ptr)
+      return true;
```
andreabrambilla: minor comment: it looks like these two lines are useless

joakim-hove: Agree - fixed

PR #597, opened 2019-04-01
#596 Output mapaxes transform if available even if use_mapaxes is false

### LINKED ISSUE #596: Mapaxes transform is not output in ecl_grid_fprintf_grdecl2() if use_mapaxes = false
This is despite the comment in ecl_grid_init_mapaxes() (the emphasis is mine):
/*
      If the apply_mapaxes variable is false we will not apply the
      transformation; however we will still internalize the mapaxes
      information, and **also output that if the file is saved again**.
*/

We are currently implementing writing of GRDECL data in ResInsight, using libecl. This is handled by creating an ecl_grid_type from a ResInsight grid, then writing it out using a combination of ecl_grid_alloc_GRID_data() and ecl_grid_fprintf_grdecl2().

We're testing this by importing an existing GRDECL file and then writing it out again and ideally we should get identical output, bar rounding errors and floating point text formatting differences.

However: the only way of getting the mapaxes data into the ecl_grid_type using the API is to pass it in using apply_mapaxes and the mapaxes array in ecl_grid_alloc_GRID_data() and if you set apply_mapaxes = true, libecl will apply the regular transform instead of the **inverse transform** we need to get the points back into the coordinate system in the original GRDECL.

Reading the comment for apply_mapaxes suggests we can perform the inverse transform ourselves, send in the transformed points and mapaxes and provide apply_mapaxes = false and it should export the mapaxes keyword even if it doesn't actually perform the transformation.

However: in ecl_grid_fprintf_grdecl2 the code checks for use_mapaxis (which is set directly from the input parameter apply_mapaxis) before exporting the keyword, meaning the mapaxes variable will not be exported if we pass in apply_mapaxis = false:

if (grid->use_mapaxes) {
    ecl_kw_type * mapaxes_kw = ecl_grid_alloc_mapaxes_kw( grid );
    ecl_kw_fprintf_grdecl( mapaxes_kw , stream );
    ecl_kw_free( mapaxes_kw );
  }

Changing to the following would mean the comment in ecl_grid_init_mapaxes() was honoured. Otherwise we'll have to perform the inverse transformation **twice** to counter for the transformation that is performed if apply_mapaxis is set to true

if (grid->mapaxes != NULL) {
    ecl_kw_type * mapaxes_kw = ecl_grid_alloc_mapaxes_kw( grid );
    ecl_kw_fprintf_grdecl( mapaxes_kw , stream );
    ecl_kw_free( mapaxes_kw );
  }

lindkvis:
**Issue**
Resolves #596

**Approach**
Outputs mapaxes into file even if use_mapaxes == false as long as the mapaxes pointer is not NULL.

### DIFF
--- a/lib/ecl/ecl_grid.cpp
+++ b/lib/ecl/ecl_grid.cpp
@@ -6949,7 +6949,7 @@ void ecl_grid_fprintf_grdecl2(ecl_grid_type * grid , FILE * stream , ert_ecl_uni
     fprintf(stream , "\n");
   }
 
-  if (grid->use_mapaxes) {
+  if (grid->mapaxes != NULL) {
     ecl_kw_type * mapaxes_kw = ecl_grid_alloc_mapaxes_kw( grid );
     ecl_kw_fprintf_grdecl( mapaxes_kw , stream );
     ecl_kw_free( mapaxes_kw );


abrautomate:

markusdregi:
Test this please

markusdregi:
@lars-petter-hauge Do you see any problems with this?

markusdregi:
I was a bit short in the comment earlier @lindkvis! Thank you for the extensive explanation in the issue you wrote... This all sounds correct and the code looks good. Could you please provide a test as well, ensuring that this behaviour works also in the future?

PR #601, opened 2019-04-01
Guard null pointer access on Windows

### LINKED ISSUE #600: Guard null pointer access on Windows
On Windows, `bool util_access(const char * entry, int mode)` causes crash if entry is a null pointer.

magnesj:
**Issue**
Resolves #600

### DIFF
--- a/lib/util/util.c
+++ b/lib/util/util.c
@@ -2155,6 +2155,11 @@ int util_fmove( FILE * stream , long offset , long shift) {
 #ifdef HAVE_WINDOWS__ACCESS
 
 bool util_access(const char * entry, int mode) {
+  if (!entry)
+  {
+      return false;
+  }
+
   return (_access(entry, mode) == 0);
 }
 


joakim-hove:
I am not the one to decide, but I would rather guard at the callsite (somewhere else in libecl) - this is a bit symptom fixing. Do you have a backtrace?

magnesj:
The root for the null pointer is `smspec_header = ecl_util_alloc_exfilename` in ecl_smspec.cpp, where the path given as input parameter is not found on the file system.

So an alternative to my pull request is to check for null pointer at this level. In my opinion, we should avoid hard crash if it is possible to guard, and communicate error situations using other means.

image

markusdregi:
> I am not the one to decide, but I would rather guard at the callsite (somewhere else in libecl) - this is a bit symptom fixing.

@joakim-hove Might be, but your voice still carries some weight on these issues due to competence ;)

@magnesj As far as I can see, the proposed fix does not propagate any error handling does it? And this error occurs on faulty input data in the sense that the restart path is broken?

> In my opinion, we should avoid hard crash if it is possible to guard, and communicate error situations using other means.

I completely agree, but that requires a scope of work that is not at the table right now in _libecl_.

joakim-hove:
> The root for the null pointer is smspec_header = ecl_util_alloc_exfilename in ecl_smspec.cpp, where the path given as input parameter is not found on the file system.

That is where we should intercept - I can make a PR.

joakim-hove:
OK - test this: #602 ?

markusdregi:
Thanks for the suggestion @joakim-hove!

Correct me if I'm wrong, but is not the scenario where this problem kicks in when you have requested to load restart data and the path to the restart file is invalid?

Because if so I would not really classify it as a bug. It is of course unfortunate that it crashes hard, but silently not loading the requested data is also not a very good option. We are really trying to keep the behaviour of _libecl_ frozen, but I would be inclined to change the hard crash to an exception. How would that be for you @magnesj? You could try loading the restart data, if you get an exception in return, you can recover by not requesting the restart data, which in the end gives you the exact same behaviour as the PR by Joakim?

joakim-hove:
Your summary is correct Markus, but it should be added that it used to fail with a message on stderr, now it crashes hard - so this PR is actually to recover old behavior.

markusdregi:
After a discussion with Joakim I've been informed that this was actually the behavior of `ecl` prior to `2.3` and it is a fallout of the `C++`-ification. I suggest that we follow up the #603 and close this one.

PR #603, opened 2019-04-02
Handle file not found when looading restarted summary

### LINKED ISSUE #602: ecl_smspec_install_special_keys : Crash when well name is null pointer
The attached summary files causes crash when using the summary.x tool for version 2.4.0
[summary_crash.zip](https://github.com/equinor/libecl/files/3033065/summary_crash.zip)

image

joakim-hove:
**Issue**
Resolves #602

**Approach**
Check return value - which can legitematly be NULL.

### DIFF
--- a/lib/ecl/tests/ecl_sum_restart.cpp
+++ b/lib/ecl/tests/ecl_sum_restart.cpp
@@ -1,0 +1,167 @@
+#include <vector>
+
+#include <ert/util/test_work_area.h>
+#include <ert/util/test_util.hpp>
+#include <ert/util/double_vector.h>
+#include <ert/util/vector.h>
+
+#include <ert/ecl/ecl_endian_flip.hpp>
+#include <ert/ecl/ecl_type.hpp>
+#include <ert/ecl/ecl_kw.hpp>
+#include <ert/ecl/ecl_file.hpp>
+#include <ert/ecl/ecl_sum.hpp>
+#include <ert/ecl/ecl_sum_vector.hpp>
+
+
+/*
+
+CASE1
+---------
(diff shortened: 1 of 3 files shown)


lib/ecl/tests/ecl_sum_restart.cpp
```diff
@@ -0,0 +1,167 @@
...
+    ecl_sum_free(ecl_sum);
+
+    util_unlink("CASE1.SMSPEC");
+
+    verify_CASE2(num_dates);
+  }
+}
+
+int main() {
+  write_CASE2();
```
markusdregi: Should we consider running all the tests here ;)

lib/ecl/well_state.cpp
```diff
@@ -649,7 +649,10 @@ time_t well_state_get_sim_time( const well_state_type * well_state ) {
    Will return NULL if no wellhead in this grid.
 */
 const well_conn_type * well_state_iget_wellhead( const well_state_type * well_state , int grid_nr) {
-  return well_state->index_wellhead[grid_nr];
```
markusdregi: This change is already in master no?

magnesj:
@joakim-hove : The current name of this PR is misleading, can you please rename it?

lib/ecl/tests/ecl_sum_restart.cpp
joakim-hove: It is recursive - case2 calls case1

lib/ecl/well_state.cpp
joakim-hove: Think so - will do a rebase later

PR #606, opened 2019-04-03
When creating a grid with external actnum the in file ACTNUM is ignored

### LINKED ISSUE #605: External ACTNUM : External values are combined with values from file
When simulation data is exported from Intersect to EGRID, the ACTNUM data is not correct. When ResInsight tries to import exported data from **Intersect,** the **PORV** keyword is inspected, and ACTNUM is set to 1 for cells with PORV > 0.0. This ACTNUM data is passed to `ecl_grid_alloc_EGRID__` using the parameter `ext_actnum`. Inside this function, the externally specified ACTNUM is combined with file internally values (by multiplying external ACTNUM values with file values).

This behaviour is causing ResInsight not to be able to import data from Intersect. We want to be able to apply the externally specified ACTNUM without modifications.

This behaviour was introduced in https://github.com/equinor/libecl/pull/445

**Reference to code**
https://github.com/equinor/libecl/blob/5a62567956c3266faa2eb36b5cf7808bed9e1ce0/lib/ecl/ecl_grid.cpp#L3025-L3041

joakim-hove:
**Issue**
Resolves #605

**Approach**
Just remove the code multiplying external and internal actnum. As mentioned in #605 the original implementation was a best guess - seems it was wrong.

NB: Before this is merged it would be good if @magnesj could verify that the number of elements in the solution vectors like `PRESSURE` and `SWAT`agree with the number of 1 values in the `ext_actnum` field.

### DIFF
--- a/lib/ecl/ecl_grid.cpp
+++ b/lib/ecl/ecl_grid.cpp
@@ -3022,23 +3022,19 @@ static ecl_grid_type * ecl_grid_alloc_EGRID__( ecl_grid_type * main_grid , const
     eclipse_version = main_grid->eclipse_version;
   }
 
-  // If ACTNUM and ext_actnum are not present - that is is interpreted as all active. 
-  const int * actnum_data;
-  std::vector<int> actnum_product;
-  if (ecl_file_get_num_named_kw(ecl_file , ACTNUM_KW) > grid_nr) {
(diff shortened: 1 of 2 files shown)


magnesj:
@joakim-hove, I can confirm the number of values in ext-actnum is matching the number of values in static properties.

PR #607, opened 2019-04-04
Silence warnings on MSVC

magnesj:
Compiling on Windows using Visual Studio 2017 generates  quite many warnings. This pull request will silence some of these warnings. The source code should be adjusted so this suppression list can be made shorter.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -120,6 +120,7 @@ endif()
 
 if (MSVC)
     add_definitions( "/W3 /D_CRT_SECURE_NO_WARNINGS /wd4996" )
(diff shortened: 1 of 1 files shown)


abrautomate:

markusdregi:
@magnesj Could you explain why this should be merged? Hiding warnings is in general a bad idea... Instead we try to make as many of them as possible errors...

We have addressed quite a few down the road, but only for `gcc` and `clang` on `Linux` and `MacOS`. Patches resolving warnings are of course welcome ;)

magnesj:
We fully understand the request for pull requests, but we do not have time right now to fix these issues correctly. This pull request is a temporary workaround to reduce warning noise when we compile libecl as part of a complete ResInsight build.

markusdregi:
> We fully understand the request for pull requests, but we do not have time right now to fix these issues correctly. This pull request is a temporary workaround to reduce warning noise when we compile libecl as part of a complete ResInsight build.

Completely understandable. But can you then not just deactivate them in your build process?

markusdregi:
I'm closing this one, please reopen if you disagree.

PR #608, opened 2019-04-04
Safely ignore entries with blank well name

### LINKED ISSUE #602: ecl_smspec_install_special_keys : Crash when well name is null pointer

joakim-hove:
**Issue**
Resolves #602

**Approach**
In most cases the dummy well ':+:+:+:+' is used in situations where well/group name does not make sense; however we have also encountered the blank string as an invalid well name; when this is trimmed we get NULL (C) or "" (C++).


markusdregi:
Could you please elaborate on where these files come from @magnesj before we merge this one?

magnesj:
Seen in simulations from the Smørbukk field.

markusdregi:
@magnesj Thank you! Do you know if they where using intersect or Eclipse?

magnesj:
@markusdregi I think this is an Eclipse simulation

PR #610, opened 2019-04-09
FWrite EGRID2: Infer Format Flag From Filename

bska:
This commit extends the ecl_grid_fwrite_EGRID2 library function to consult the filename in order to infer whether or not to create a formatted EGRID file.  Previously, this function would always create an unformatted EGRID output file, irrespective of filename extension.

**Issue**
Resolves part of OPM/opm-simulators#1784.

**Approach**
_Consult the filename's extension to determine whether or not to create a formatted output file_

### DIFF
--- a/lib/ecl/ecl_grid.cpp
+++ b/lib/ecl/ecl_grid.cpp
@@ -6854,6 +6854,12 @@ static void ecl_grid_fwrite_EGRID__( ecl_grid_type * grid , fortio_type * fortio
 
 void ecl_grid_fwrite_EGRID2( ecl_grid_type * grid , const char * filename, ert_ecl_unit_enum output_unit) {
   bool fmt_file        = false;
+  {
+    bool is_fmt;
+
+    if (ecl_util_get_file_type( filename , &is_fmt, NULL ) != ECL_OTHER_FILE)
+      fmt_file = is_fmt;
+  }
   fortio_type * fortio = fortio_open_writer( filename , fmt_file , ECL_ENDIAN_FLIP );
 
   ecl_grid_fwrite_EGRID__( grid , fortio, output_unit);
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -329,6 +329,7 @@ foreach (name   ecl_alloc_cpgrid
                 ecl_grid_copy
                 ecl_grid_create
                 ecl_grid_DEPTHZ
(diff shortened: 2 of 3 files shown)


abrautomate:

lib/ecl/ecl_grid.cpp
```diff
@@ -6854,6 +6854,13 @@ static void ecl_grid_fwrite_EGRID__( ecl_grid_type * grid , fortio_type * fortio
 
 void ecl_grid_fwrite_EGRID2( ecl_grid_type * grid , const char * filename, ert_ecl_unit_enum output_unit) {
   bool fmt_file        = false;
+  {
+    bool is_fmt;
+    int rpt_nr = -1;
```
markusdregi: This can be removed and then you can just pass a `NULL`-pointer instead.

markusdregi:
Thank you for the pull request! This looks like a good start 👍 

Could you please add a test?

joakim-hove:
> Could you please add a test?

Tip for a test:
```C++
{
    ecl::util::TestArea ta("fmt_file");
    ecl_grid = ecl_grid_alloc_rectangular(5,5,5,1,1,1,nullptr);
    ecl_grid_fwrite_EGRID2( ecl_grid, "CASE.FEGRID", ECL_METRIC_UNITS);
    test_assert_true( util_fmt_bit8("CASE.FEGRID"));
}
```

lib/ecl/ecl_grid.cpp
bska: > This can be removed and then you can just pass a `NULL`-pointer instead.

Okay, thanks for the hint.  I'll update accordingly.

bska:
> > Could you please add a test?
> 
> Tip for a test:

Thanks.  I'm away from my development box now, but I'll add the test first thing in the morning when I'm back in the office.

markusdregi:
Test this please

bska:
Pull request updated per review comments.

markusdregi:
Retest this please

markusdregi:
> Pull request updated per review comments

Thank you for the update! It looks good...

Let us wait for the Jenkins tests to finish before I approve it.

markusdregi:
ok to test

markusdregi:
Looks good to me 👍

bska:
> Looks good to me

Thanks, much appreciated!

PR #611, opened 2019-04-12
Update module hierarchy used in import of Python examples

blattms:
At least for installed python modules there is no ecl.ecl anymore that contains all imported files. This PR updates the examples to use the correct hierarchy.


abrautomate:

andreabrambilla:
ok to test

PR #612, opened 2019-04-15
[bugfix] Use correct ijk for NNC in python example

blattms:


abrautomate:

andreabrambilla:
ok to test

PR #614, opened 2019-05-02
Make global index used consistent in cmp_nnc.py

### LINKED ISSUE #613: treatment of global index wrong in nnc_cmp.py example
The example uses the global indices read from the *.EGRID directly and passes it to libecl to convert it to an ijk index. The problem is that in the eclipse files the global index is 1-based (1, 2, ...) as in FORTAN but the libecl functions expect it to be 0-based (0, 1, 2, ...). Therefore the ijk indices printed are wrong.

blattms:
**Issue**
Resolves #613 

**Approach**
We simply decrement the 1-based globale indices read from the eclipse files before passing them to the libel functions.

In addition the second commit resorts to printing the ijk indices as 1-based. That is a matter of taste of course but seems less surprising to me,

### DIFF
--- a/python/docs/examples/cmp_nnc.py
+++ b/python/docs/examples/cmp_nnc.py
@@ -23,6 +23,10 @@ if __name__ == "__main__":
 
     nnc_list = sorted(nnc_list, key = itemgetter(0))
     for (g1,g2,T) in nnc_list:
+        # grid_ijk assumes 0-based indexing, g1/g2 are 1-based (FORTRAN)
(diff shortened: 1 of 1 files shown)


abrautomate:

andreabrambilla:
ok to test

lars-petter-hauge:
Yeh, the `NNC1` and `NNC2` arrays are given by cell number (`ICELL` ) which is defined as 1-based.. While the `grid.get_ijk` indeed returns 0-based indexes.

Issue #585, opened 2019-03-13
Various problems with the new smspec-code


markusdregi:
**Setting**
There has been some problems occurring with the new smspec-implementation after moving to _C++_.

**Related**
Insufficient workaround in ResInsight;
https://github.com/OPM/ResInsight/issues/4172

OPM-issues:
https://github.com/OPM/ResInsight/commit/dfd45397866442146cab59fbe046cf03d3d8b56d
https://github.com/OPM/ResInsight/commit/b120d0a2c1f0aef4a4811f6bd9b416023a3bfc20

dotfloat:
If the ResInsight PRs haven't been upstreamed in this repo, then we should do it.

PR #615, opened 2019-05-08
Improve robustness

magnesj:
**Issue**
Multiple test cases has triggered issues related to null pointer access and assignment of NULL to a std::string causing crash.

Resolves parts of #585

**Approach**
Guard null pointer access
Do not assign NULL to a std::string

### DIFF
--- a/lib/ecl/ecl_smspec.cpp
+++ b/lib/ecl/ecl_smspec.cpp
@@ -699,10 +699,15 @@ static void ecl_smspec_install_special_keys( ecl_smspec_type * ecl_smspec , cons
       ecl_smspec_get_well_var_index( smspec , well_name , var );
   */
 
-  const char * well            = smspec_node_get_wgname( &smspec_node );
-  const char * group           = well;
-  const int num                = smspec_node_get_num(&smspec_node);
-  const char * keyword         = smspec_node_get_keyword(&smspec_node);
+  std::string well;
+  if (smspec_node_get_wgname(&smspec_node)) well = smspec_node_get_wgname(&smspec_node);
+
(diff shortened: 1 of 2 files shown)


andreabrambilla:
ok to test

markusdregi:
Could you add a test @magnesj verifying that the new implementation behaves differently then the old, so that this behavior is preserved also in the future?

magnesj:
@markusdregi I do not have the test system configured, so I do not have time available to create a test related to my changes.

markusdregi:
Dear @magnesj.

I've investigated this one a bit further. To me it seems like the `DOUBHEAD` is indeed to be present in a valid file. Of course, if _libecl_ could internalize the present data, compile a list of errors, let you act accordingly and still behave in a consistent manner that would be the best scenario. However, the rest of the code is not currently written to work in this way and hence you might get dubious errors down the road when poking the loaded data.

Could you please elaborate on how these files are made?

Cheers,
Markus

markusdregi:
@magnesj Did pull request #617 in any way affect the need of this patch?

The functions trying to return `NULL` as a `string` should definitively be fixed, but I'm not sure if this is the right way to go about it...

jokva:
I don't think is the right fix. The right fix is to revert it to again return const char*, not std::string.

markusdregi:
I agree @jokva. Will make a new PR.

jokva:
#621 takes this back to return const char*. These values are set in the gen_key1/2 of smspec_node, but like before silently fail (i.e. do nothing) if the smspec_alloc functions should ever return nullptr.

PR #616, opened 2019-05-08
Guard null pointer access for custom generated Eclipse files used in geomech

magnesj:
**Issue**
Geomech workflow generates some special Eclipse files. They are not complete, and crashes when trying to import into libecl.

**Approach**
Guard null pointer access

### DIFF
--- a/lib/ecl/ecl_rsthead.cpp
+++ b/lib/ecl/ecl_rsthead.cpp
@@ -88,7 +88,10 @@ ecl_rsthead_type * ecl_rsthead_alloc_from_kw( int report_step , const ecl_kw_typ
       // The only derived quantity
       rsthead->sim_time  = rsthead_date( rsthead->day , rsthead->month , rsthead->year );
   }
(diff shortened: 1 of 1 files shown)


markusdregi:
Thank you for the patch @magnesj. I will have to investigate this one a bit further.

In particular, if `DOUBHEAD` is a required part of the header, although this might fix your scenario, this might lead to strange errors for other users down the road. I'm assuming, without having checked yet, that the rest of the code is not robust enough to handle an invalid header, and then an early failure is to be preferred.

markusdregi:
Dear @magnesj.

I've investigated this one a bit further. To me it seems like the `DOUBHEAD` is indeed to be present in a valid file. Of course, if _libecl_ could internalize the present data, compile a list of errors, let you act accordingly and still behave in a consistent manner that would be the best scenario. However, the rest of the code is not currently written to work in this way and hence you might get dubious errors down the road when poking the loaded data.

Could you please elaborate on how these files are made?

Cheers,
Markus

markusdregi:
Closed due to inactivity. Feel free to reopen it if you want to continue the effort.

PR #617, opened 2019-05-12
Ecl sum invalid key

joakim-hove:
If you ask the ecl_sum object for an invalid key like:
```
double value = ecl_sum_get_general_var(ecl_sum , 1, "NO_SUCH_KEY");
```
you would previously be met with the venerable `util_abort()`- in current master you will silently get an out of range (corresponding to a lookup index of value -1) value. With this PR invalid values will be met with `std::out_of_range()`exception.

### DIFF
--- a/lib/ecl/ecl_smspec.cpp
+++ b/lib/ecl/ecl_smspec.cpp
@@ -1287,18 +1287,28 @@ int ecl_smspec_get_num_regions(const ecl_smspec_type * ecl_smspec) {
 
 /*****************************************************************/
 
-#define NODE_RETURN_INDEX(node_ptr)                     \
-  if (!node_ptr)                                        \
-    return -1;                                          \
-  else                                                \
-    return smspec_node_get_params_index( node_ptr );
 
 
-#define NODE_RETURN_EXISTS(node_ptr)            \
-  if (!node_ptr)                                \
-    return false;                               \
-  else                                          \
-    return true;
+
+namespace {
+
+  bool node_exists(const ecl::smspec_node * node_ptr) {
+    if (node_ptr)
+      return true;
+
+    return false;
+  }
+
+
+  int node_valid_index(const ecl::smspec_node * node_ptr) {
+    if (node_ptr)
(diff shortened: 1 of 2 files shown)


lib/ecl/ecl_smspec.cpp
```diff
@@ -1287,18 +1287,28 @@ int ecl_smspec_get_num_regions(const ecl_smspec_type * ecl_smspec) {
...
 
-#define NODE_RETURN_EXISTS(node_ptr)            \
-  if (!node_ptr)                                \
-    return false;                               \
-  else                                          \
-    return true;
+
+namespace {
+
+  bool NODE_EXISTS(const ecl::smspec_node * node_ptr) {
```
markusdregi: Should we make these styled as function names instead of MACROS?

joakim-hove: Yes, that makes sense - I just wanted to keep the amount of change down. Will update the PR

lib/ecl/ecl_smspec.cpp
```diff
@@ -1287,18 +1287,28 @@ int ecl_smspec_get_num_regions(const ecl_smspec_type * ecl_smspec) {
...
+
+    return false;
+  }
+
+
+  int NODE_VALID_INDEX(const ecl::smspec_node * node_ptr) {
+    if (node_ptr)
+      return node_ptr->get_params_index();
+
+    throw std::invalid_argument("Invalid lookup summary object");
```
markusdregi: I think an `out_of_bounds` exception would be more in line with STL behavior at least, i.e. `map::at`.

markusdregi:
Thank you for the contribution. This looks like a good approach!

lib/ecl/ecl_smspec.cpp
joakim-hove: I have updated to `std::out_of_range()` - feel the `std::` exception hierarchy is slightly limited - and nearly always end up on `std::invalid_argument` - but yes; I agree `std::out_of_range()` is a better choice.

PR #620, opened 2019-05-21
Set Micro version

lars-petter-hauge:


jokva:
I presume there should be a tag, too?

lars-petter-hauge:
Yes, will be added afterwards

PR #621, opened 2019-05-22
Revert smspec return types from string to char*

markusdregi:
When converting return types to `std::string` some of the changed functions did continue to return `NULL` at occasions. As this is not valid, action is necessary. We propose a rollback to the previous API with a `char *` and then let it be until a more extensive evaluation of the API is done.

### DIFF
--- a/lib/ecl/smspec_node.cpp
+++ b/lib/ecl/smspec_node.cpp
@@ -22,9 +22,11 @@
 #include <math.h>
 #include <time.h>
 
+#include <cstdlib>
 #include <string>
 #include <set>
 #include <array>
+#include <memory>
 #include <stdexcept>
 
 #include <ert/util/hash.hpp>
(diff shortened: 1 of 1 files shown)


markusdregi:
@magnesj Will this fix your problem from #615?

jokva:
Have you looked over call-sites too, and checked that it doesn't leak now, or is that for me to do?

markusdregi:
> or is that for me to do

That is for me to do. Will get back to you when I've gone through it. I guess @magnesj should still be able to do a test with the failing case to see how it goes...

jokva:
Other than that, this obviously looks excellent. Whenever you feel it's fine, and magnesj has gotten to give some feedback.

magnesj:
This PR fixes some of the issues we have in ResInsight. Please merge this one, and I will create a separate issue for other crash situations.

PR #622, opened 2019-05-23
Init basic ecl3 directory structure

jokva:
Initialise the basic project layout of ecl3, effectively an isolated and
independent project in the ecl3/ dir.

It adds a separate catch2 dependency, which at later should be replaced
with a centralised one as the ecl3 library matures.

**Issue**
Resolves #<Issue id>

**Approach**


ecl3/lib/CMakeLists.txt
```diff
@@ -0,0 +1,63 @@
+cmake_minimum_required(VERSION 3.5)
+project(ecl3-lib LANGUAGES C CXX)
+
+set(ECL3_LIB_BINARY_DIR ${CMAKE_CURRENT_BINARY_DIR} CACHE PATH "")
+
+set(CMAKE_CXX_STANDARD 11)
+
+add_library(ecl3 STATIC
```
andreabrambilla: If I understand correctly, we have an `ecl` library that links to a static `ecl3`. Is this a temporary solution (to drag in `ecl3` functionality into `ecl` incrementally i guess), or is the plan to have it like this also in the long run? In the latter case, what would be the main difference between `ecl` and `ecl3`?

jokva: This is a temporary solution indeed, the STATIC option will go away once ecl3 is the main citizen.

PR #627, opened 2019-06-03
Active cells from PORV : Assign combined bit mask

### LINKED ISSUE #626: Dual porosity models : Computation of active cells based on pore volume fails
When PORV is used as criteria, active cells are marked with the bit mask with value 1. For dual porosity models, the fracture active cell has bit mask of value 2. No active cells for fracture models will be reported.

To fix this, use the combination of the two bit mask, value 3.

magnesj:
When PORV is used as criteria, make sure all active cells are assigned both active matrix state and active fracture state. This will make sure that both single porosity models and dual porosity models are initialized with the correct bit mask.

**Issue**
Resolves #626

### DIFF
--- a/lib/ecl/ecl_kw_functions.cpp
+++ b/lib/ecl/ecl_kw_functions.cpp
@@ -136,7 +136,7 @@ ecl_kw_type * ecl_kw_alloc_actnum(const ecl_kw_type * porv_kw, float porv_limit)
     return NULL;
 
   if (!util_string_equal(PORV_KW, ecl_kw_get_header(porv_kw)))
-    return NULL;
+        return NULL;
 
   const int size = ecl_kw_get_size(porv_kw);
   ecl_kw_type * actnum_kw = ecl_kw_alloc(ACTNUM_KW, size, ECL_INT);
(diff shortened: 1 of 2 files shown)


abrautomate:

lib/ecl/ecl_kw_functions.cpp
```diff
@@ -143,9 +143,16 @@ ecl_kw_type * ecl_kw_alloc_actnum(const ecl_kw_type * porv_kw, float porv_limit)
   const float * porv_values = ecl_kw_get_float_ptr(porv_kw);
   int * actnum_values = ecl_kw_get_int_ptr( actnum_kw);
 
+  // When PORV is used as criteria, make sure all active cells are assigned both
+  // active matrix state and active fracture state. This will make sure that
+  // both single porosity models and dual porosity models are initialized with
+  // the correct bit mask. See documentation in top of ecl_grid.cpp
```
joakim-hove: This fix assumes that all cells have *both* fracture and matrix active simultaneously; that is correct in 95% of the cases - but unfortunately not in all cases - then there would be no point in having the bitmask functionality. I think *really* solving this requires taking a step back - whether it is worth it?

magnesj: This functionality is intended to be a helper function in some situations from ResInsight. ResInsight will be responsible to make sure it will be valid to use this function. A fallback to ACTNUM keyword is still needed.

What keywords creates models with difference in active cells between matrix and fracture? Do you have any examples we can try?

jokva: Sure, but other applications won't. I'm not sure if they exist, but if they do this might break.

I have a few questions:

1. Can resinsight not just-in-time adjust the model to also address Joakim's issue and set fractures accordingly
2. If not, can resinsight use a different function for this, with which sets both matrix and fracture active?

magnesj: One solution from our point of view can be to add the bitmask to the list of parameters
`ecl_kw_alloc_actnum(const ecl_kw_type * porv_kw, float porv_limit, int bitmask)`

jokva: That's the "different function" approach, but I will have no problems accepting that patch.

joakim-hove: > One solution from our point of view can be to add the bitmask to the list of parameters

That will certainly be an improvement - but it will not solve the fundamental problem, namely that some cells should actnum value 1 (i.e. only matrix), some should have value 2 (i.e. only fracture) and some should have value 3 - for both matrix and fracture.

I will shut the f... up now.

magnesj: This function is no way intended to handle the use case where active cells differ between matrix and fracture. A totally different function/parameters is relevant in that case. I can update the PR.

PR #630, opened 2019-06-18
update redhat packaging

akva2:
add ecl3

**Approach**
Add new files to packaging definition

### DIFF
--- a/redhat/ecl.spec
+++ b/redhat/ecl.spec
@@ -2,10 +2,10 @@
 # spec file for package ecl
 #
 
-%define tag rc1
+%define tag git1
 
 Name:           ecl
-Version:        2018.10
+Version:        2019.10
 Release:        0
 Summary:        ECL library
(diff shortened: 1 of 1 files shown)


abrautomate:

jokva:
There's no need for this, does red hat packaging complain? ecl3 is not ready for packaging nor use.

akva2:
it complains about installed but unpackaged files..

jokva:
Hm, the default-install was done in order to make CI simpler, do you actually package HEAD?

akva2:
yes, we have nightly builds for opm where we build ecl head.

jokva:
My goodness.

When ecl2 is discontinued, the RH package files will be moved. I'd rather see this is a small change in the install script instead, so you can opt-out of building ecl3.

akva2:
okay. i opened #631 which adds an option and use this to disable in packaging.

PR #631, opened 2019-06-18
Ecl3 option

akva2:
**Issue**
Due to the new ecl3 target, packaging breaks due to installed but unpackaged files.

**Approach**
Add an option to disable ecl3, and use this is packaging.


jokva:
I'll merge once travis does its job.

abrautomate:

PR #634, opened 2019-06-28
Restore previous abort handler if it was present

mortalisk:
In order to have both c and python stack traces on errors, we
call pythons abort handler after done with util_abort. Also
move the check for flag ECL_SKIP_SIGNAL, since the method
to install signal handler was called several places.

Also fixes an error in util_abort where it would deadlock if it could not open a file, and then called util_abort again.

### DIFF
--- a/lib/util/util_abort_gnu.c
+++ b/lib/util/util_abort_gnu.c
@@ -248,8 +248,10 @@ static char* __abort_program_message;
 
 void util_abort__(const char * file , const char * function , int line , const char * fmt , ...) {
   util_abort_test_intercept( function );
-  pthread_mutex_lock( &__abort_mutex ); /* Abort before unlock() */
-  {
+  /* if we cannot lock, presumably we could not open a file and we are calling
+   * util_abort recursively */
+  int lock_status = pthread_mutex_trylock( &__abort_mutex );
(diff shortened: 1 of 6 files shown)


abrautomate:

andreabrambilla:
jenkins test this please

lib/util/util_abort_gnu.c
```diff
@@ -313,10 +315,12 @@ void util_abort__(const char * file , const char * function , int line , const c
     }
     chmod(filename, 00644); // -rw-r--r--
     free(filename);
+    pthread_mutex_unlock(&__abort_mutex);
+  } else {
+    /* Failure in util_abort. Print basic error message and exit. */
+    fprintf(stderr, "Error while handling error. Exiting. File: %s Function: %s, Line: %d\n", file, function, line);
```
andreabrambilla: This is fine, but we are not printing the content of the `va_list`. One option could be to move the lines 267-275 outside the if and make it print to a string instead. Then, if we can lock the mutex, we write the string to file, otherwise we display it on stderr

mortalisk: I'll just repeat those lines needed for diplaying the error if that is ok.

PR #637, opened 2019-07-04
Check for sighandler_t and redifine only if needed

joakim-hove:
The PR: #634 seems to have broken the build of opm: https://github.com/OPM/opm-common/pull/884

This PR restores that build. The changes in the PR are:

1. There is `try_compile()` based cmake check to look for the `sighandler_t` type, and it is only redeclared if needed.

2. The management of previous signal handler is implemented in the source file instead of the header file.


joakim-hove:
Thank you for merging

PR #642, opened 2019-08-01
Use six elements from summary STARTDAT keyword

joakim-hove:
In a summary SMSPEC file the start time of the simulation is encoded in the `STARTDAT` keyword. In current version of libecl this is encoded as a three element integer with DAY, MONTH, YEAR, but it turns out that it can optionally also be stored as a six element vector with addition elements HOUR, MINUTE, MICRO_SECOND.

This PR adds the three new elements HOUR, MINUTE, MICRO_SECOND - and optionally reads them if they are present in the file when reading.

### DIFF
--- a/lib/ecl/ecl_smspec.cpp
+++ b/lib/ecl/ecl_smspec.cpp
@@ -425,12 +425,15 @@ static void ecl_smspec_fwrite_DIMENS(const ecl_smspec_type * smspec, fortio_type
 
 static void ecl_smspec_fwrite_STARTDAT(const ecl_smspec_type * smspec, fortio_type * fortio) {
   ecl_kw_type * startdat_kw = ecl_kw_alloc( STARTDAT_KW , STARTDAT_SIZE , ECL_INT );
-  int day,month,year;
-  ecl_util_set_date_values( smspec->sim_start_time , &day, &month , &year);
+  int second,minute,hour,mday,month,year;
+  ecl_util_set_datetime_values(smspec->sim_start_time, &second, &minute, &hour, &mday,&month,&year);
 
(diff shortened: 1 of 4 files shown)


andreabrambilla:
@joakim-hove @jokva 
It appears that this PR has broken a couple of tests in libres related to the simulation context. The issue is a bit nasty to debug as it occures only on osx on travis (see https://travis-ci.org/equinor/libres/jobs/571240984)
Do you have any idea what might be the connection between this specific PR and those tests?

joakim-hove:
> It appears that this PR has broken a couple of tests in libres related to the simulation context. The issue is a bit nasty to debug as it occures only on osx on travis (see https://travis-ci.org/equinor/libres/jobs/571240984)
Do you have any idea what might be the connection between this specific PR and those tests?

I have no idea - and quite frankly I find it a bit hard to believe that there is such a relation; if anything my hunch would be that the current PR provokes existing buggy behavior in libres - but I find also that quite hard to believe. 

A bit tight now; but I can assist in debugging this tomorrow.

andreabrambilla:
I agree it's absurd, but I am sure it's triggered by this (directly or indirectly). I have created a custom branch with this commit reverted, and built libres against that, and all the tests passed.

The osx image we use in the libres build is quite outdated (I think there was a reason for that, but can't remember now). I'll try updating it and see if the issue disappears

andreabrambilla:
Ok, updating the osx image worked, sorry for bothering you!

joakim-hove:
👍

Den ons. 14. aug. 2019, 12:03 skrev Andrea Brambilla <
notifications@github.com>:

PR #647, opened 2019-08-28
Include argument in select_method example

lars-petter-hauge:
**Issue**
 No issue created - user reported

**Approach**
Include necessary argument in the documentation for the `select_method`

### DIFF
--- a/python/ecl/grid/ecl_region.py
+++ b/python/ecl/grid/ecl_region.py
@@ -52,18 +52,22 @@ def select_method(select):
     Consider this example:
 
        region = EclRegion( grid , False )
-       region.select_islice(0 , 5)     # Selects all cells with i:[0:5]
-       region.select_jslice(0 , 5)     # Selects all cells with j:[0:5]
+       region.select_islice(0, 5)     # Selects all cells with i[0:5]
+       region.select_jslice(0, 5)     # Selects all cells with j[0:5]
 
     When these two calls have completed selection will contain all
     the cells which are either in i-interval [0:5] or in
     j-interval [0:5]. If we supply the @intersect argument in the
     second call the j selection will only be applied to the cells
-    in i:[0:5] interval:
+    in i[0:5] interval:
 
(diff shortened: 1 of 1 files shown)


python/ecl/grid/ecl_region.py
```diff
@@ -62,8 +62,12 @@ def select_method(select):
     in i:[0:5] interval:
 
        region = EclRegion( grid , False )
-       region.select_islice(0 , 5)     # Selects all cells with i:[0:5]
-       region.select_jslice(0 , 5)     # Selects all cells with j:[0:5] AND i:[0:5]
+
+       # Select all cells with i:[0:5]:
```
markusdregi: What is up with the colons?

lars-petter-hauge: Probably some sort of preference? How about now? Removed some colons and some whitespace

python/ecl/grid/ecl_region.py
```diff
@@ -52,18 +52,22 @@ def select_method(select):
...
 
        region = EclRegion( grid , False )
-       region.select_islice(0 , 5)     # Selects all cells with i:[0:5]
-       region.select_jslice(0 , 5)     # Selects all cells with j:[0:5] AND i:[0:5]
+
+       # Select all cells with i[0:5]:
+       region.select_islice(0, 5)
+
+       # Select all cells with j[0:5] AND i[0:5]:
+       region.select_jslice(0, 5, intersect = True)
```
markusdregi: If you are fixing white spaces, then maybe `intersect=True` while you're at it?

python/ecl/grid/ecl_region.py
markusdregi: Still don't get what the colons are about, but I guess it is not mission critical anyway.

python/ecl/grid/ecl_region.py
lars-petter-hauge: fixed

PR #649, opened 2019-08-28
Remove obsolete documentation

### LINKED ISSUE #584: Documentation on read the docs
Should we build `libecl` documentation on read the docs or similar solution?

markusdregi:
**Issue**
Resolves #584 and https://github.com/equinor/ert/issues/447


python/docs/code/python/index.rst
```diff
@@ -1,11 +1,1260 @@
...
 
-Python documentation
-====================
+.. contents::
+   :depth: 2
+   :local: 
 
-.. toctree::
-   :maxdepth: 2
+The :code:`ert.ecl` package for working with Eclipse files is quite
```
lars-petter-hauge: `ert.ecl` is deprecated right? Then we ought to have only the `ecl` entry

python/docs/code/python/index.rst
```diff
@@ -1,11 +1,1260 @@
...
+:code:`__setitem__()` methods which are used to implement access using
+the :code:`[ ]` notation, and the :code:`__len__()` method which gives
+the size of the :code:`EclKW` instance. In the example below we load
+an INIT file and extract the PERMX and PORO keywords, we then
+forcefully set permeability to zero for all elements where the
+porosity is below a limit:
+
+.. code:: python
+
+   from ert.ecl import EclFile,EclGrid
```
lars-petter-hauge: And changing the `ert.ecl`, all of the imports in the examples must be changed as well
```
from ecl.file import EclFile
from ecl.grid import EclGrid
```

lars-petter-hauge:
I very much agree on the changes here, but I haven't looked through it as such to see if there are any outdated information.

Main objection so far is the legacy `ert.ecl` usage in the documentation.

markusdregi:
> I haven't looked through it as such to see if there are any outdated information.

Me neither. But I thought that a good first step was to remove the things that should not be there... And then we can improve upon the existing documentation afterwards.

> Main objection so far is the legacy ert.ecl usage in the documentation.

Well, for some definition of deprecated (where deprecated is an idea among the developers combined with an already existing alternative). With a fresh build of `libecl`:

```python
>>> from ert.ecl import EclFile
>>> 
```

I would say that it is up for discussion whether it is worth starting the entire import deprecation process with `ecl3` as a hard break in the making. I suggest that we save this discussion for later and just remove the clearly obsolete stuff for now.

lars-petter-hauge:
>where deprecated is an idea among the developers

I was under the impression that we had deprecation warning for `ert.ecl` usage, but of course that's for the camelcase/snakecase change. In that case I agree that we should avoid trying to move users to a new API which will be superseded in a not too distant future

PR #651, opened 2019-09-04
Call correct time function

joakim-hove:
Followup fix to: #642

A central element in #642 is to break down a `time_t` value to a full datetime (year, month, day, hour, minute, second) - where it was previously only broken down to (year,month,day). Unfortunately there was a bug in #642 which lead to the (hour, minute, second) part to be undefined.


bska:
Does this return the seconds as seconds (0..59) or as microseconds (0..59,999,999)?

joakim-hove:
> Does this return the seconds as seconds (0..59) or as microseconds (0..59,999,999)?

seconds (0..59) - it is ultimately based on unix `gmtime_r` which has seconds resolution; and it goes in and out of the function as unix `time_t`.

PR #653, opened 2019-09-06
Allow mixed case filenames

### LINKED ISSUE #582: Aborts if mixed case is used in datafile name
Eclise does not handle mixed case in filenames, while Intersect does. Libecl tests for mixed case, which might not make sense as we are only to read the data anyway. Should we remove these guards?

Could you pinpoint the relevant code @lars-petter-hauge?

sondreso:
**Issue**
Resolves #582 

**Approach**
Remove mixed case checking in filename. Preserve checking of upper case such that the file extensions added by `ecl_util_alloc_filename` will be consistent with previous behavior:
```
ecl_util_alloc_filename(NULL, "UPPER", ECL_EGRID_FILE, false, -1) -->  UPPER.EGRID
ecl_util_alloc_filename(NULL , "lower", ECL_EGRID_FILE, false, -1) --> lower.egrid
```
Mixed case will get uppercase extension
```
ecl_util_alloc_filename(NULL, "miXed", ECL_EGRID_FILE, false, -1) -->  miXed.EGRID
```

### DIFF
--- a/lib/ecl/ecl_util.cpp
+++ b/lib/ecl/ecl_util.cpp
@@ -335,30 +335,18 @@ const char * ecl_util_file_type_name( ecl_file_enum file_type ) {
   return NULL;
 }
 
-static bool valid_base(const char * input_base, bool * upper_case) {
-  bool upper = false;
-  bool lower = false;
+static bool base_has_upper(const char * input_base) {
   const char * base = strrchr(input_base, UTIL_PATH_SEP_CHAR);
   if (base == NULL)
     base = input_base;
 
   for (size_t i=0; i < strlen(base); i++) {
-    unsigned char c = base[i];
-
-    if (isupper(c))
-      upper = true;
-
-    if (islower(c))
-      lower = true;
-
+    if (isupper(base[i]))
+      return true;
   }
 
-  if (upper_case)
-    *upper_case = upper;
-  return !(lower && upper);
+  return false;
 }
-
-
 
 
 
@@ -373,11 +361,6 @@ static bool valid_base(const char * input_base, bool * upper_case) {
 */
 
 static char * ecl_util_alloc_filename_static(const char * path, const char * base , ecl_file_enum file_type , bool fmt_file, int report_nr, bool must_exist) {
-  bool upper_case;
-  if (!valid_base(base, &upper_case))
-    return NULL;
-
-
   char * filename;
   char * ext;
   switch (file_type) {
@@ -454,7 +437,7 @@ static char * ecl_util_alloc_filename_static(const char * path, const char * bas
     ext        = NULL;
   }
 
-  if (!upper_case) {
+  if (!base_has_upper(base)) {
     for (size_t i=0; i < strlen(ext); i++)
       ext[i] = tolower(ext[i]);
   }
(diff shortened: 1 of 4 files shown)


lib/ecl/ecl_util.cpp
```diff
@@ -373,10 +361,7 @@ static bool valid_base(const char * input_base, bool * upper_case) {
 */
 
 static char * ecl_util_alloc_filename_static(const char * path, const char * base , ecl_file_enum file_type , bool fmt_file, int report_nr, bool must_exist) {
-  bool upper_case;
-  if (!valid_base(base, &upper_case))
-    return NULL;
-
+  bool upper_case = base_has_upper(base);
```
markusdregi: I suggest that you move this call to the place where `upper_case` is used. Also, the standard is `ALL_CAPS_NAMES`, so we should really check if `AlL_CaPs_NameS` should yield `AlL_CaPs_NameS.DATA`, `AlL_CaPs_NameS.DaTa` or `AlL_CaPs_NameS.data`. I would expect the first option.

@magnesj Could you comment on this?

magnesj: I think that filenames on disk can come from a vast range of sources, and import of these files should be as flexible as possible.

lib/ecl/ecl_util.cpp
```diff
@@ -645,34 +630,29 @@ static int ecl_util_select_predicate_filelist(const char * path, const char * ba
...
-  if (base)
-    valid_case = valid_base(base, &upper_case);
+  if (file_type == ECL_SUMMARY_FILE || file_type == ECL_RESTART_FILE)
+    return ecl_util_select_predicate_filelist(path, base, file_type, fmt_file, upper_case, filelist);
 
-  if (valid_case) {
-    if (file_type == ECL_SUMMARY_FILE || file_type == ECL_RESTART_FILE)
-      return ecl_util_select_predicate_filelist(path, base, file_type, fmt_file, upper_case, filelist);
+  char * ext_pattern = util_alloc_string_copy(ecl_util_get_file_pattern( file_type , fmt_file ));
+  char * file_pattern;
```
markusdregi: Please move to where it is used.

lib/ecl/ecl_util.cpp
markusdregi: The problem is that it is an established assumption in _libecl_ that the extension casing can be deduced from the base casing. Hence, there is a huge difference in effort if we are to support the base `MixedCase` and we can assume that when mixed case the datafile extension is either always i.e. `DATA` or always i.e. `data`. For example:
 - `ALL_UPPER` -> `ALL_UPPER.DATA`,
 - `all_lower` -> `all_lower.data` and
 - `SOME_lower` -> `SOME_lower.DATA`.

Of course, we can also change it so that `SOME_lower` always assumes `SOME_lower.data`. The problem with allowing both is that then information regarding extensions have to be persisted and propagated through the code. In addition, the consumer of the API would have to specify when writing the summary data what extension they want.

I suggest we finish of this PR with the continued assumption that extension casing is entirely deduced by base name and that this is validated. You still get full flexibility regarding the base name.

Then we can discuss if you want support beyond this afterwards as this will require quite a bit more... Does this sound good?

magnesj: Thanks for the explanation. This PR is an improvement, and I think this will fix the import issues for many users. Thanks.

markusdregi: No problem at all! Sorry for the staleness of this PR. I'll make sure the task is well defined and follow it up so that we get it in soon :+1:

markusdregi:
My suggestion for this PR is then that we do the following. We continue to require that the extension casing is deducible from base casing and extend the rules in the following manner to allow for mixed case base names:

- all base names are valid,
- if base name contains a single upper case character the extension is upper case and
- otherwise, when there are no upper case character, the extension is lower case.

We should validate that the above holds, as parts of reading and in particular writing, will fail hard with the current implementation if this is not respected. In particular, `BASE.DATA`, `Base.DATA`, `baSE.DATA` and `base.data` are all valid, while `BASE.DaTA`, `BASE.data`, `Base.data` and `base.DATA` are all invalid.

PR #655, opened 2019-09-17
Compile with -fPIC switch

joakim-hove:
Needed when a shared library `somelib.so` links to `libecl.a` library.

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -199,7 +199,8 @@ endif ()
 
 set_target_properties(ecl PROPERTIES
                           VERSION ${ECL_VERSION_MAJOR}.${ECL_VERSION_MINOR}
-                          SOVERSION ${ECL_VERSION_MAJOR})
+                          SOVERSION ${ECL_VERSION_MAJOR}
+                          POSITION_INDEPENDENT_CODE ON)
 
 install(TARGETS ecl
         EXPORT  ecl-config


jokva:
Why not building shared if you want shared?

joakim-hove:
> Why not building shared if you want shared?

That would also work of course; but this is for a python extension where the extension library `xxx_python.so` would link functions from libecl. If the latter is compiled statically the extension can load runtime without considering the runtime loading of symbols with e.g. `LD_LIBRARY_PATH`.

jokva:
Sure, but if you package it correctly (wheel with a .libs dir and pathelf dancing) you can accomplish that. Or simply use bespoke flags for python-oriented builds.

joakim-hove:
> Sure, but if you package it correctly (wheel with a .libs dir and pathelf dancing) you can accomplish that. Or simply use bespoke flags for python-oriented builds.

I have never built a wheel; you might be right that this can be done in better ways. But anyway if I:

1. Build and install libecl statically.
2. Build a Python extension with plain `setup.py`- which links to the installed version of libecl.

I get these link errors:
```
/usr/bin/ld: /home/hove/work/ERT/install/lib/libecl.a(util.c.o): relocation R_X86_64_PC32 against symbol `stderr@@GLIBC_2.2.5' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: /home/hove/work/ERT/install/lib/libecl.a(util_abort_gnu.c.o): relocation R_X86_64_PC32 against symbol `stderr@@GLIBC_2.2.5' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: /home/hove/work/ERT/install/lib/libecl.a(util_spawn.cpp.o): relocation R_X86_64_PC32 against symbol `environ@@GLIBC_2.2.5' can not be used when making a shared object; recompile with -fPIC
```
when generating the final extension library. These linker errors are fixed with this PR.

jokva:
Sure, but they're also fixed by passing -DCMAKE_POSITION_INDEPENDENT_CODE=ON when configuring.

joakim-hove:
> Sure, but they're also fixed by passing -DCMAKE_POSITION_INDEPENDENT_CODE=ON when configuring.

Yes - we agree on that; my point was just that with this PR any default built `libecl.a` library lying around could be used in a .so file.

jokva:
And I believe that's the wrong default. If you want shared or PIC, ask for it, but cmake defaults predictable.

jokva:
Other maintainers, feel free to chime in. I vote no.

lars-petter-hauge:
I don't have that much experience as to the consequences of enabling/disabling fpic (especially what the consequences are for external libraries using libecl), but I believe we should be as predictable as possible and use default behaviour where suitable.

> ..the shared library compilation and linking flags must have -fPIC..

> ..static libraries, the common case is not to have relocatable code..

> In other words, if both a shared and a static library is being built, each source unit (*.c, for example, for C files) will need to be compiled twice, for the normal case.

https://www.debian.org/doc/debian-policy/ch-files.html#fr84

If we force the use of fPIC, it is impossible for the user to change behaviour from the command line.   That would make it troublesome for users that e.g. wish to build static and without positional.

A fairly affordable approach has been suggested which would not impose any changes to other external libraries.

I do not agree to this change.

jondequinor:
The CMake defaults match the debian policy, `-DCMAKE_POSITION_INDEPENDENT_CODE=ON` is the answer here (or a patch for the downstream project). I vote no.

PR #664, opened 2019-09-25
#663 Fix crash when calling ecl_grid_reset_actnum

### LINKED ISSUE #663: Crash when calling ecl_grid_reset_actnum()
We have a dataset I'm not yet (at least) able to disclose which crashes due to a ecl_grid->inv_index_map nullpointer which comes from the ecl_grid->total_active being set to 0 before inv_index_map is reallocated.

This occurs in ecl_grid_set_active_index because the active_index is never incremented if the following conditions  are met:
1. You're calling ecl_grid_reset_actnum after a grid has already been loaded (to read PORV). This means the cells active_index has been initialized.
2. There are coarse cells, so the fast path is NOT taken.
3. The cells all have a course group, so the `cell->coarse_group == COARSE_GROUP_NONE` is false.

The code will then use ecl_coarse_cell_update_index() which checks that the active_index is -1 before allowing it to be set, meaning it will never set the index and never increment the number of active cells leaving it at 0, causing the nullpointer.

This behaviour doesn't match the comment to ecl_grid_set_activ_index which says:
`/*
  This function goes through the entire grid and sets the active_index
  of all the cells. The functione ecl_grid_realloc_index_map()
  subsequently reads this to create and initialize the index map.
*/`

... because it currently only guarantees to set the active_index if the active_index hasn't already been set.

lindkvis:
**Issue**
Resolves #663 

**Approach**
* Following review: reset active_index across the grid for the involved path in ecl_grid_set_active_index.
* It does not appear necessary to do this for the fast path because that sets active_index even if it wasn't -1 before hand.

### DIFF
--- a/lib/ecl/ecl_coarse_cell.cpp
+++ b/lib/ecl/ecl_coarse_cell.cpp
@@ -290,17 +290,13 @@ const int * ecl_coarse_cell_get_box_ptr( const ecl_coarse_cell_type * coarse_cel
 
 void ecl_coarse_cell_update_index( ecl_coarse_cell_type * coarse_cell , int global_index , int * active_index , int * active_fracture_index , int active_value) {
   if (active_value & CELL_ACTIVE_MATRIX) {
-    if (coarse_cell->active_index == -1) {
-      coarse_cell->active_index = *active_index;
-      (*active_index) += 1;
-    }
+    coarse_cell->active_index = *active_index;
+    (*active_index) += 1;
   }
 
   if (active_value & CELL_ACTIVE_FRACTURE) {
-    if (coarse_cell->active_fracture_index == -1) {
-      coarse_cell->active_fracture_index = *active_fracture_index;
-      (*active_fracture_index) += 1;
-    }
+    coarse_cell->active_fracture_index = *active_fracture_index;
+    (*active_fracture_index) += 1;
   }
 
   int_vector_append( coarse_cell->active_cells , global_index );
--- a/lib/ecl/ecl_coarse_cell.cpp
+++ b/lib/ecl/ecl_coarse_cell.cpp
@@ -290,13 +290,17 @@ const int * ecl_coarse_cell_get_box_ptr( const ecl_coarse_cell_type * coarse_cel
 
 void ecl_coarse_cell_update_index( ecl_coarse_cell_type * coarse_cell , int global_index , int * active_index , int * active_fracture_index , int active_value) {
   if (active_value & CELL_ACTIVE_MATRIX) {
-    coarse_cell->active_index = *active_index;
-    (*active_index) += 1;
+    if (coarse_cell->active_index == -1) {
+      coarse_cell->active_index = *active_index;
+      (*active_index) += 1;
+    }
   }
 
   if (active_value & CELL_ACTIVE_FRACTURE) {
-    coarse_cell->active_fracture_index = *active_fracture_index;
-    (*active_fracture_index) += 1;
+    if (coarse_cell->active_fracture_index == -1) {
+      coarse_cell->active_fracture_index = *active_fracture_index;
+      (*active_fracture_index) += 1;
+    }
   }
 
   int_vector_append( coarse_cell->active_cells , global_index );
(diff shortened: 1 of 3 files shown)


ertomatic:

lindkvis:
I was assigned to this, but I am the submitter. If there is any need to modify the patch, please let me know, but AFAIK this hasn't been evaluated yet.

markusdregi:
Thank you for your patch @lindkvis :+1: We will need some time to go through the consequences of the change. We will get back to you as soon as possible.

markusdregi:
@lindkvis Sorry for the late reply!

We have looked at the relevant code now and we think there might be an issue with your patch. In particular, `ecl_coarse_cell_update_index` is using `active_index` as an indicator for whether this course cell has already been updated. We suggest that instead of your edit, you reset the active index across the entire grid to `-1` at the start of `ecl_grid_set_active_index`.

Also, could you please provide a test that fails on the current implementation that verifies the new patch?

markusdregi:
Dear @lindkvis. We are preparing a new release of libecl now. Are you interested in pushing this issue so that we get it in before that? I know that you are not necessarily following the releases of libecl, but I think potential bug-fixes etc would be easier to handle if we sync up a bit...

lindkvis:
@markusdregi . I have updated the pull request following review. Because we have to alter the coarse cell indices as well, which is in a private scope inaccessible from ecl_grid that led to a new method in ecl_coarse_cell. Please advice if this isn't what you meant.

As for data, we are not at liberty to share this and have no sharable data. I can at least ask @hhgs if he is able to make it available to you for testing.

lib/ecl/ecl_grid.cpp
```diff
@@ -1871,7 +1871,29 @@ static void ecl_grid_set_active_index(ecl_grid_type * ecl_grid) {
...
+        {
+            if (cell->active & CELL_ACTIVE_MATRIX)
+            {
+                cell->active_index[MATRIX_INDEX] = -1;
+            }
+
+            if (cell->active & CELL_ACTIVE_FRACTURE)
+            {
+                cell->active_index[FRACTURE_INDEX] = -1;
+            }
```
andreabrambilla: Unless I missed something, the `cell->active_index` and `active_fracture_index` are just overwritten further down (lines 1906 and 1911 at the time of this review), with no checks about their previous values. I suspect we can remove the entire `if` and keep just the lines in the `else` block

lib/ecl/ecl_coarse_cell.cpp
```diff
@@ -287,6 +287,17 @@ const int * ecl_coarse_cell_get_box_ptr( const ecl_coarse_cell_type * coarse_cel
 
 
 /*****************************************************************/
+void ecl_coarse_cell_reset_active_index(ecl_coarse_cell_type * coarse_cell, int active_value) {
+    if (active_value & CELL_ACTIVE_MATRIX)
```
andreabrambilla: I presume the `active_value` will be used to determine whether this `coarse_cell` has to be updated during an update. Here, we can probably disregard that value an directly set both the active indices to -1

lib/ecl/ecl_grid.cpp
```diff
@@ -1871,7 +1871,29 @@ static void ecl_grid_set_active_index(ecl_grid_type * ecl_grid) {
...
+        } else {
+            ecl_coarse_cell_type * coarse_cell = ecl_grid_iget_coarse_group(ecl_grid, cell->coarse_group);
+            ecl_coarse_cell_reset_active_index(coarse_cell, cell->active);
```
andreabrambilla: If my comment in `ecl_coarse_grid.cpp` is correct, then we do not need to loop through all the cells in order to retrieve the coarse cells. We can directly access the list of coarse cells in the grid and reset them all

lib/ecl/ecl_grid.cpp
lindkvis: that looks reasonable

lib/ecl/ecl_coarse_cell.cpp
lindkvis: I'm removing the active_value parameter and just setting them unconditionally inside the function.

lib/ecl/ecl_grid.cpp
lindkvis: I've updated the pull request. I don't really have any more time to put into this so I hope this is acceptable.

andreabrambilla:
jenkins test this please

PR #666, opened 2019-10-11
Remove warning regarding python 3

oysteoh:


python/cmake/Modules/init_python.cmake
```diff
@@ -62,9 +62,7 @@ macro(init_python target_version)
       return()
    endif()
 
-   if (PYTHON_VERSION_MAJOR EQUAL 3)
-      message(WARNING "libecl is not PYTHON 3 compatible. Python wrappers are enabled, but will fail horribly.")
-   elseif (NOT "${PYTHON_VERSION_MAJOR}.${PYTHON_VERSION_MINOR}" STREQUAL "${target_version}")
+  if (PYTHON_VERSION_MAJOR EQUAL 2 AND NOT "${PYTHON_VERSION_MAJOR}.${PYTHON_VERSION_MINOR}" STREQUAL "${target_version}")
```
markusdregi: Are you sure you want to add the `PYTHON_VERSION_MAJOR EQUAL 2`? Should this not be evaluated for every supported Python version and hence now also for Python 3.x?

oysteoh: Target version is fixed to 2.7 currently in the config, and this check from my point of view ends up with verifying python version is 2.7. All versions above should be ok... But - i might be wrong :)

markusdregi: Alright, so if we have Python 2.x for x <= 6, we should turn of Python bindings. That is fine. Should we also warn if the Python 3 version is not among the supported ones? Currently we only test on Travis for 3.6....

python/cmake/Modules/init_python.cmake
```diff
@@ -62,9 +62,11 @@ macro(init_python target_version)
...
 
-   if (PYTHON_VERSION_MAJOR EQUAL 3)
-      message(WARNING "libecl is not PYTHON 3 compatible. Python wrappers are enabled, but will fail horribly.")
-   elseif (NOT "${PYTHON_VERSION_MAJOR}.${PYTHON_VERSION_MINOR}" STREQUAL "${target_version}")
+  if (PYTHON_VERSION_MAJOR EQUAL 3)
+      if (NOT PYTHON_VERSION_MINOR EQUAL 6)
+          message(WARNING "libecl for Python 3 is only tested and verified against Python 3.6. Python wrappers are enabled, but not guaranteed to work.")
+      endif()
+  elseif (NOT "${PYTHON_VERSION_MAJOR}.${PYTHON_VERSION_MINOR}" STREQUAL "${target_version}")
       message(WARNING "Need Python version ${target_version}, found version: ${PYTHON_VERSION_MAJOR}.${PYTHON_VERSION_MINOR} - Python wrappers not enabled")
```
markusdregi: I suggest we change this to be explicit on that if MAJOR is 2, we only support `2.7`. Now, that message sounds like we only support `2.7` period.

PR #670, opened 2019-10-23
Change WARNING -> FATAL_ERROR in python cmake

### LINKED ISSUE #669: CMake should exit if ENABLE_PYTHON is ON but the Python dependencies aren't found.
If the user chose to `-DENABLE_PYTHON=ON` when compiling, then they likely don't want CMake to quietly warn the user and `set(ENABLE_PYTHON OFF)` when the dependencies are missing.

dotfloat:
**Issue**
Resolves #669 

**Approach**
Changing to FATAL_ERROR is enough to have CMake exit with an error after printing.


ertomatic:

PR #676, opened 2019-12-10
Fix bugs found using asan, ubsan

dotfloat:


ertomatic:

lib/ecl/ecl_file_kw.cpp
```diff
@@ -403,16 +403,16 @@ ecl_file_kw_type ** ecl_file_kw_fread_alloc_multiple( FILE * stream , int num) {
         buffer_offset += ECL_STRING8_LENGTH;
       }
 
-      kw_size = *((int *) &buffer[ buffer_offset ]);
+      memcpy(&kw_size, &buffer[ buffer_offset ], sizeof kw_size);
```
xjules: why we need to copy here? it's used in `ecl_file_kw_alloc0` which only passes the value further on and not the reference

lib/ecl/ecl_grid.cpp
```diff
@@ -1533,6 +1534,14 @@ static ecl_grid_type * ecl_grid_alloc_empty(ecl_grid_type * global_grid,
                                             int nz,
                                             int lgr_nr,
                                             bool init_valid) {
+  /* Check for int overflow in a roundabout way */
```
xjules: Why the grid size needs to be `< INT_MAT`?

lib/ecl/ecl_file_kw.cpp
dotfloat: Both versions do a copy. The problem with casting it instead of explicitly copying is undefined behaviour due to aligned issues. Some architectures don't support `mov`-ing data from misaligned addresses and will crash. This isn't the case for either x86 or any other remotely relevant architecture for us, but it's still UB.

`*((int *) &x)` roughly means "assume that x is an int", which will create `mov` instructions for copying the "int" `x` into a different memory location. Suppose that the architecture only supports integers that are aligned to 16 bytes (that is, their hex memory address _must_ end with 0). `*((int *) &x)` cannot guarantee this, and so it will crash if we first read 1 byte and then try to read an int from the buffer. However, `memcpy` is a simple byte for byte copy, and so will always work.

Both version generate the same assembly with GCC on x86. https://godbolt.org/z/XDwbYF

lib/ecl/ecl_grid.cpp
dotfloat: https://github.com/equinor/libecl/blob/f1443476bc784f5b1d052e33cf1ea3235c41a00a/python/tests/ecl_tests/test_grid.py#L129-L135

The `size` (an int) overflows in one of the test due to x = y = z = 2000, and 2000<sup>3</sup> > `INT_MAX`. This is undefined behaviour. Because `INT_MAX`<sup>3</sup> > `ULONG_MAX`, we can't simply cast each `int` to `uint64_t` and do the multiplication, so we have to test each multiplication in this roundabout way.

The previous version had a `size` that overflowed and then would `util_abort` when trying to malloc <strong>-</strong>589934592 bytes, so adding this check will make this more robust.

lib/ecl/ecl_file_kw.cpp
xjules: valid point!

lib/ecl/ecl_grid.cpp
xjules: We need to find out what's the maximum grid size in use

joakim-hove: > We need to find out what's the maximum grid size in use

The `.EGRID` file format dictates that the number of elements in the `ACTNUM` array can be represented with a 32bit signed integer - i.e. `nx*ny*nz < 2**31 - 1`

lib/util/tests/ert_util_strcat_test.cpp
```diff
@@ -40,17 +40,20 @@ int main(int argc , char ** argv) {
...
-    test_strcat(NULL, util_alloc_string_copy(s) , s);
+    const char *s = "Hei";
+    char *c = util_alloc_string_copy(s);
+    test_strcat(NULL, c, s);
+    free(c);
   }
   {
     const char * s = "Hei";
-    test_strcat(util_alloc_string_copy(s) , NULL , s);
+    test_strcat(util_alloc_string_copy(s), NULL, s);
```
xjules: can u revert back the whitespaces?

PR #682, opened 2020-01-09
CMake: Use current git tag for libecl version

### LINKED ISSUE #680: Have CMAKE extract version from git tag
Instead of versions being set explicitly in the CMakeList.txt, we should extract that information from the git tag.

dotfloat:
Closes #680 
Same approach as https://github.com/equinor/libres/pull/862


ertomatic:

PR #685, opened 2020-01-14
Avoid installing with .c.x suffix, only .x

### LINKED ISSUE #684: Installation suffixes for binary files
Due to a change in 2018, some of the applications, like "ecl_unpack" now installs as `ecl_unpack.c.x` instead of `ecl_unpack.x`.

Other applications, like `summary.x` is still `summary.x` as before. Perhaps the `.c.x` was not intentional.

berland:
**Issue**
Resolves #684 

**Intention**

Change the file suffix for the binary targets 
`ecl_pack,  ecl_unpack,  kw_extract, grid_info, grid_dump, grid_dump_ascii, select_test, load_test`
from `*.c.x` or `*.cpp.x` to only `*.x`

### DIFF
--- a/applications/CMakeLists.txt
+++ b/applications/CMakeLists.txt
@@ -5,24 +5,24 @@ if (BUILD_APPLICATIONS)
     add_executable(make_grid   ecl/make_grid.c)
     add_executable(grdecl_grid ecl/grdecl_grid.c)
     add_executable(summary     ecl/view_summary.cpp)
+    add_executable(kw_extract  ecl/kw_extract.cpp)
     target_link_libraries(sum_write     ecl)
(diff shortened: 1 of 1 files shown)


ertomatic:

pgdr:
Can you split your patch into two commits and state in the PR description what your intention is?

It's not clear from the issue that you link to what is the intended consequence of this patch.

berland:
I fail to see how to split the commit into two (meaningful) commits (?).

Added text to PR describing the intention.

markusdregi:
I've tested the installation procedure on my machine and it seems to work here also. Waiting for further updates from @pgdr

PR #690, opened 2020-01-27
Add deprecation warnings when import ecl.ecl or import ert.ecl

oysteoh:


python/ecl/ecl/__init__.py
```diff
@@ -9,4 +9,12 @@
...
 
+import warnings
+
+warnings.simplefilter('always', DeprecationWarning)
+warnings.warn(
+    "Importing from ert is deprecated and will be removed in a future release, use: \'from ecl import\' instead",
+    DeprecationWarning
+)
+
 from ert.ecl import *
```
sondreso: Will this not hit both deprication warnings because of this line?

python/ecl/ecl/__init__.py
```diff
@@ -9,4 +9,12 @@
 # Both of these legacy forms are controlled by the cmake switch
 # INSTALL_ERT_LEGACY.
 
+import warnings
+
+warnings.simplefilter('always', DeprecationWarning)
+warnings.warn(
+    "Importing from ert is deprecated and will be removed in a future release, use: \'from ecl import\' instead",
```
sondreso: from ecl.ecl :slightly_smiling_face:

python/ecl/ecl/__init__.py
oysteoh: does not seems like it does..... and anyways - better having two deprecationwarnings instead of changing the implementation.

python/ert/__init__.py
```diff
@@ -1,3 +1,11 @@
+import warnings
```
sondreso: I think this file will be overwritten with ert/__init__.py from libres, will that have any unintended side-effects?

sondreso:
Note: Related to https://github.com/equinor/libecl/issues/679

PR #691, opened 2020-02-05
pip install

### LINKED ISSUE #686: Make ecl pip installable

dotfloat:
This PR makes it possible to `pip install libecl` while keeping the old `cmake`-based install approach (ie. not a breaking change).
_(Note: I talk about Linux here but everything also applies to macOS or any other POSIX-compliant OS. WINAPI support can probably be added relatively easily, but I haven't investigated it much.)_

In particular, this PR does the following:
1. Fix tests so that it's possible to use pytest
2. Add `setup.py`, `pyproject.toml`, etc so that it's possible to do `pip install .` (Must have recent-ish `pip`)
3. Use https://github.com/matthew-brett/multibuild as to build wheels for Python 2.7 and 3.6, deploying to PyPI on tags
4. Installs `libecl.so` into the Python module's directory under `.libs` (meaning `LD_LIBRARY_PATH` is entirely ignored and `import ecl` loads the bundled `libecl.so`)
5. Installs headers (ie. `include/`) into the module's `.include` directory

### Changes to development of ecl

The new build pipeline requires a new workflow when it comes to development as it's no longer recommended to rely entirely on CMake. Rather, from now on, it is required to run `pip install -e .` in order to generate `libecl.so`. (`-e` is short for `--edit`) This does the following automatically:
1. Uninstall `libecl` if it was installed previously
2. Read `pyproject.toml` and install setup dependencies
3. Run `python setup.py build`, which will call CMake in a directory called `_skbuild`
4. Installs extensions into current working directory (that is, `python/ecl/.libs/`)

This means setting `PYTHONPATH=$PWD/python` will allow `ecl` to find `libecl.so`. Note that this happens automatically when calling `pytest` on the local tests directory. This must be done every time C sources or `CMakeLists.txt`s are changed.

Alternatively, one can use `setup.py` directly. Install setup requirements manually with `pip install -r requirements.txt` (`pip install .` will install these, but they aren't available after installation). Then it's possible to use `setup.py`:

`python setup.py develop`: does the same as `pip install -e .`
`python setup.py build`: only builds `libecl.so`

Both of these can be used with flags:

`python setup.py develop --build-type=debug`: run CMake with `-DCMAKE_BUILD_TYPE=Debug`
`python setup.py develop -- -DBUILD_TESTS=ON`: run CMake with `-DBUILD_TESTS=ON` (any other CMake flag can be supplied after `--`)

More info can be found in scikit-build docs: https://scikit-build.readthedocs.io/en/latest/

### Changes to users of ecl

The Python components should behave as before and no change is required if the user is only using the Python bindings (except maybe adding `"libecl"` to requirements). However, linking to `libecl.so` is slightly more complicated.

Since `libecl.so` exists only in a `site-packages` directory somewhere, it won't be able to be found by the linker `ld.so`. But, we dynamically open `libecl.so` in `ecl/__init__.py` with the flag `RTLD_GLOBAL` which makes every single symbol found in the library available in the global symbol table. This means that any unsatisfied symbols in the client library (eg. `util_abort__`) are already loaded and so a lookup doesn't happen.

Users still need the headers, though. Their path can be found via:
```python
import ecl
ecl.get_include()  #=> "/usr/lib/python3.6/site-packages/ecl/.include"
```
This value can be passed to eg. CMake, which needs to do the following, assuming it was called with `-DECL_INCLUDE_DIRS={ecl.get_include()}`:
```cmake
target_include_directories(my-project PRIVATE ${ECL_INCLUDE_DIRS})
```
Note that the projects must not try to link with `libecl.so`. This would mean that the only way to use the user's library is via Python, by first importing the ecl module. Non-python libraries are still supported via the old cmake install pipeline, where `libecl.so` is installed somewhere where the linker can find it (`/usr/local/lib64/libecl.so.2`).

### RE: Using Python exceptions in `util_abort`
It is possible to throw a Python exception in `util_abort` and have it be caught in Python. However, none of the C functions provided by Python allow for unwinding of the stack, which is required in order to get execution out of `libecl` and into Python again. This is obvious in hindsight since Python would otherwise leak like crazy as C can't clean up after itself.

`PyErr_SetString(PyExc_RuntimeError, "util_abort");` -- which is one of way to set that a Python exception occurred from C -- only sets a global variable. It's the program's duty to manually return somewhere where the exception can be handled.

If we replace `cwrap` with `pybind11`, we'd be able to use C++ exceptions which do unwind to `pybind11` which rethrows them as Python exceptions. However, this requires that we replace _all_ instances of `EclPrototype` with a proper function call to the C extension. Luckily, this can be accomplished by a search-and-replace script and automatically generating some C++ code. Unluckily, this could mean breaking change to users of our library, since we will no longer use `RTLD_GLOBAL` to load `libecl.so`.

---

Resolves #686


markusdregi:
🚀 PR of the year 🚀

requirements.txt
```diff
@@ -1,7 +1,12 @@
 cwrap
-numpy
-pandas
-sphinx
+functools32;python_version=='2.7'
 future
+ninja
+numpy<=1.16.6
```
markusdregi: Any particular reason for upper bounding the numpy and pandas version?

.travis.yml
```diff
@@ -1,69 +1,130 @@
...
-  - sudo make install
-  - which python
-  - export PYTHONPATH="/usr/local/lib/python$MB_PYTHON_VERSION/site-packages:/usr/local/lib/python$MB_PYTHON_VERSION/dist-packages:$PYTHONPATH"
-  - python -c "import sys; print('\n'.join(sys.path))"
-  - set -e; python -c "import ecl"; set +e
-  - ctest --output-on-failure $TEST_SUITE
-  - popd
+  - |
+    if [ -z "${LEGACY_INSTALL}" ]; then
+      install_run $PLAT
```
markusdregi: Where does `install_run` come from?

markusdregi: Is it multi build magic?

markusdregi:
As mentioned earlier, this is really cool 😎

markusdregi:
Have you looked into why Jenkins is failing?

requirements.txt
dotfloat: Due to Python 2.7 -- Pandas 1.0 dropped support for it and the latest is 0.25.3 .

Had problems with latest numpy+py2, so I put an upper bound on the one that worked for me, but it's not very scientific.

.travis.yml
dotfloat: multibuild magic. It's in `travis_steps.sh`

dotfloat:
> I agree with the changes to the Python tests and Python build system. The Travis pipeline does look good, but I'm not familiar enough with multibuild to approve it immediately.
> 
> Should we also update the readme, and possibly the docs, to reflect the changes to building?

We can discuss multibuild in more detail when you're able. Readme and docs should definitely be updated. Writer's block is my excuse for not having done it already. ;)

requirements.txt
markusdregi: I don't mind an upper bound for Python 2.7, but then we should make it explicit for `python_version < "3"` and then have no upper bound for `python_version >= "3"

dotfloat:
@markusdregi, I edited the README

requirements.txt
dotfloat: Should be good now.

setup.cfg
```diff
@@ -0,0 +1,2 @@
+[aliases]
+test = pytest
```
markusdregi: Missing new line....

dotfloat:
TODOs:
- [ ] `libecl.so` is not dependent on python version, so it should be distributed as a `py{2,3}-none-manylinux2010-x86_64` wheel and not versioned as it is now. (Look at cmake pypi pkg)
- [ ] Remove `set -x` from travis

mortalisk:
I pip-installed this, equinor-libres and ert.
Running snake-oil, it is going very slow, and there are a lot of deprecation warnings.

setup.py
```diff
@@ -1,52 +1,70 @@
...
+from setuptools_scm import get_version
 
 
-class CMakeExtension(Extension):
-    def __init__(self, name, sourcedir=''):
-        Extension.__init__(self, name, sources=[])
-        self.sourcedir = os.path.abspath(sourcedir)
+def src(x):
+    root = os.path.dirname(__file__)
+    return os.path.abspath(os.path.join(root, "python", x))
```
mortalisk: remove this

everapparatus:

PR #693, opened 2020-02-13
Add komodo ci test

jondequinor:
unit test running on komodo jenkins

build: https://ci.equinor.com/komodo/job/komodo-test-libecl/4/console


markusdregi:
This looks like a good start and I'm really happy that we are getting to the stage of automatic testing towards build environments 👏 

I would however like to discuss the interface of these scripts, as the plan should be to do similar work in all of the other repositories. Did you consider moving the responsibility of setting up the environment, figuring out which version should be built etc to the central unit? So that the only thing a project has to do, to be able to hook into the test pipeline is to write a script that is provided with an already setup environment builds and runs the tests. Because, I guess we should fetch the correct version of the script, so that we already have to clone the repo and checkout the tag?

jondequinor:
Yes, we figured that isolated jobs could be re-used, as opposed to a central job. E.g. this job could run for each commit to `libecl@master`, giving valuable feedback on how a commit would fare were it deployed to komodo.

markusdregi:
Aight, brain has been applied...

markusdregi:
I see that RELEASE_PATH etc has been parameterised, would it be an idea to do the same with the environment (/opt/rh/devtoolset-7/enable) and such to extract the Equinor'ism of it?

jondequinor:
@markusdregi it has been done as you have willeth. There's nothing new under the sun though.

PR #704, opened 2020-03-23
Fix incorrect build steps in README

### LINKED ISSUE #702: cmake error: could not load cache
I'm using WLS running Ubuntu inside Windows 10 and trying to use ecl_summary to plot well curves simulated with flow (according to [OPM project's first tutorial](https://opm-project.org/?page_id=197&page=5) libecl should already be installed, since I have flow, but I can't find it).

I'm trying to follow Alternative 3 step-by-step to install libecl and set the BUILD_ECL_SUMMARY switch to on in the CMake configuration (according to [OPM project's wiki](https://wiki.opm-project.org/index.php?title=Main_Page#Text_based_solution:_ecl_summary) this is necessary to get the ecl_summary application).

When I run the command
`cmake libecl -DENABLE_PYTHON=ON -DBUILD_ECL_SUMMARY=ON`
`cmake --build libecl --target install`
I get the error:
`Error: could not load cache`

I'm running everything from the root folder.

dotfloat:
**Issue**
Resolves #702

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,14 +31,20 @@ compiler such as GNU GCC, the CMake build system and, optionally, zlib.
 
 ```bash
 $ git clone https://github.com/Equinor/libecl
-$ cmake libecl
-$ cmake --build libecl --target install
+$ mkdir libecl/build
+$ cd libecl/build
+$ cmake ..
+$ make
+$ make install
 ```
 
 To install *libecl* in a non-standard location, add
-`-DCMAKE_INSTALL_PATH=/path/to/install` to the first `cmake` command. If you
-intend to develop and change *libecl* you should build the tests by passing
-`-DBUILD_TESTS=ON` and run the tests with `ctest`.
+`-DCMAKE_INSTALL_PREFIX=/path/to/install` to the first `cmake` command. Remember
+to set `LD_LIBRARY_PATH=/path/to/install/lib64:$LD_LIBRARY_PATH` if you do use a
+non-standard location for your program to find `libecl.so`.
+
+If you intend to develop and change *libecl* you should build the tests by
+passing `-DBUILD_TESTS=ON` and run the tests with `ctest`.
 
 ### Alternative 3: C library with Python bindings ###
 It is also possible to install both the C library and Python bindings using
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -31,8 +31,11 @@ compiler such as GNU GCC, the CMake build system and, optionally, zlib.
 
 ```bash
 $ git clone https://github.com/Equinor/libecl
-$ cmake libecl
-$ cmake --build libecl --target install
+$ mkdir libecl/build
+$ cd libecl/build
+$ cmake ..
+$ make
+$ sudo make install
```
markusdregi: Why are we recommending to install using `sudo`? That sounds like horrible advice one would only find on the internet...

README.md
```diff
@@ -48,9 +51,12 @@ compiler, CMake and, optionally, zlib.
 
 ```bash
 $ git clone https://github.com/Equinor/libecl
-$ pip install -r libecl/requirements.txt
-$ cmake libecl -DENABLE_PYTHON=ON
-$ cmake --build libecl --target install
+$ mkdir libecl/build
+$ cd libecl/build
+$ pip install -r ../requirements.txt  # You need write-access to Python's packages
```
markusdregi: I think the comment is not necessary. If someone cannot pip install something, the great internet will explain how they can mitigate

README.md
dotfloat: What would you suggest as an alternative? The «nice» way to do this would be to provide ‘.deb’ and ‘.rpm’. CMake has tools to generate these.

README.md
dotfloat: The issue is that it’s not immediately obvious that pip has failed, and our CMake falls back to disabling Python if it can’t find Python stuff instead of erroring.

README.md
dotfloat: I actually checked now and our CMake Python stuff is so incorrect that `sudo make install` doesn’t do the correct thing.

`CMAKE_INSTALL_PREFIX` is a pretty advanced feature, how would you suggest they use it? Install into a virtualenv?

markusdregi: Either into a virtualenv or into an install prefix that they append to their paths. The next sentence in the readme goes:

> Then, you must tell Python where to find the package[1]:

I think most users will be happy with pip-installing it. And the ones who need the C-API will probably be fine with dealing with `LD_LIBRARY_PATH`?

dotfloat: I’m not comfortable asking users who need these instructions to mess around with environment variables. Very user unfriendly. The correct place for user-managed software is `/usr/local`, which is the default for `make install`, but requires root.

I’ll ping @jondequinor for his opinion on this. Maybe also @jokva .

README.md
markusdregi: I agree. But at the same time there are multiple ways pip can fail, perhaps a general warning about checking that the pip installation was successful is better? Or maybe even inspecting the exit code of pip?

README.md
markusdregi: I kind of like how `segyio` has done it:
https://github.com/equinor/segyio#build-segyio

jondequinor: I think `sudo` is superfluous, and not often seen in build instructions. I think the instructions should assume as little as possible about the user's environment. So the assumption that the user is going to install into some pre-existing prefix is fine by me. Setting `LD_LIBRARY_PATH` is not good in my opinion: unsafe and unnecessary.

dotfloat:
Remove `sudo`, added description on how to install into a virtualenv and how to use `LD_LIBRARY_PATH`.

PR #707, opened 2020-04-15
Expanded list of rate identifiers

### LINKED ISSUE #706: smpec_node is_rate is lacking several rate identifiers
The rate identifier array in  [smspec_node.cpp](https://github.com/equinor/libecl/blob/6f89be14d0c8c62c606d2da17c580115cfce3b75/lib/ecl/smspec_node.cpp#L444-L458) is not very comprehensive. In addition it includes some (WCT and GOR), but not all (OGR, LGR, WGR) production ratios, which are unitless and not really rates, but closely connected. 
Propose to extend the list and include the rest of the production ratios, as GOR and WCT is already there and removing them _may_ influence users.

asnyv:
**Issue**
Resolves #706 

**Approach**
Added several additional rate identifiers in addition to remaining production ratios (addition to already existing water-cut (WCT) and gas-oil ratio (GOR). Structured them into groups per line with a small comment section to make it a bit more human friendly to read. Also wanted to include some var_type verification similar to what is done in [smspec_node_identify_total](https://github.com/equinor/libecl/blob/6f89be14d0c8c62c606d2da17c580115cfce3b75/lib/ecl/smspec_node.cpp#L461-L509), but the additional required input parameter broke (at least) the python API and I couldn't (fairly quickly) identify why. Limited experience with C++, so I'll blame that 😉  Also not really sure if that var_type logic provides a lot of extra benefits. If someone wants to add it (or help me on the way), feel free.

I didn't add absolutely all rate vectors in the manual, but (most of) the relevant ones should hopefully be in. At least a lot more than there has been previously.
Might be an idea to take a similar look at the `smspec_node_identify_total` if you find this ok, though that list seems more comprehensive already.  (did add a couple on is_total, but just some that I came across when looking at rates)

### DIFF
--- a/lib/ecl/smspec_node.cpp
+++ b/lib/ecl/smspec_node.cpp
@@ -28,6 +28,8 @@
 #include <array>
 #include <memory>
 #include <stdexcept>
+#include <algorithm>
+#include <vector>
 
 #include <ert/util/hash.hpp>
 #include <ert/util/util.h>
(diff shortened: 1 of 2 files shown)


ertomatic:

sondreso:
ok to test

dotfloat:
This looks fine to me, though we could convert this code to more proper C++.

Regarding `var_type`: it's not difficult to add this parameter. I could hijack this PR and perform these changes myself or we could discuss it on video.

asnyv:
> This looks fine to me, though we could convert this code to more proper C++.
> Regarding var_type: it's not difficult to add this parameter. I could hijack this PR and perform these changes myself or we could discuss it on video.

@dotfloat If you have some minutes for a chat, that would be appreciated 😄 Wanted to change as little as possible before I knew if you found the idea ok or not. Not sure what is considered _proper C++_, but I would at least prefer to finish the char pointer arrays with a `nullptr`, skip the integer division for array length (`num_rate_vars`) and loop with a while loop until `nullptr,` as I found that pointer length solution confusing for a non-c++ speaker 😉

dotfloat:
```c++
bool smspec_node_identify_rate(const char * keyword) {
    static const std::unordered_set<std::string> rate_vars {{
        "OPR" , "OIR", "OPRF", "OPRS", "OVPR", "OVIR", "OFR", "OPP", "OPI", "OMR",
        "GPR" , "GIR", "GPRF", "GPRS", "GVPR", "GVIR", "GFR", "GPP", "GPI", "GMR",
        "WPR" , "WIR", "WVPR", "WVIR", "WFR", "WPP", "WPI", "WMR",
        "LPR", "LIR",
        "VPR", "VIR", "VFR",
        "GLIR", "RGR", "EGR", "EXGR", "SGR", "GRS", "FGR", "GIMR", "GCR"
        "NPR", "NIR", "CPR", "CIR", "SIR", "SPR",
        "EPR", "EIR",
        "GOR" , "WCT", "OGR", "WGR", "GLR"
    }};

    // keyword is probably not an empty string, but let's not assume things
    if (!keyword[0])
        return false;

    return rate_vars.count(&keyword[1]) > 0;
}
```
:wink:

lib/ecl/smspec_node.cpp
```diff
@@ -439,73 +441,147 @@ char * smspec_alloc_local_completion_key( const char * join_string, const std::s
 }
 
 /*****************************************************************/
+static bool match_keyword_vector(int ipos, const std::vector<std::string> &vars, const std::string &keyword) {
+    //Validate string length    
+    if (!keyword[ipos]){
```
dotfloat, line 446: This is incorrect. Use `keyword.size() < ipos` to check that the size of the string is at least `ipos` bytes long.

dotfloat: The reason I did `!keyword[0]` to test for emptiness is that C strings are required to terminate with a null (`'\0'`), meaning that they're at least one byte long. Strings aren't allowed to contain non-terminating null terminators for obvious reason, so this is a simple check, but it doesn't scale.

lib/ecl/smspec_node.cpp
```diff
@@ -439,73 +441,147 @@ char * smspec_alloc_local_completion_key( const char * join_string, const std::s
 }
 
 /*****************************************************************/
+static bool match_keyword_vector(int ipos, const std::vector<std::string> &vars, const std::string &keyword) {
```
dotfloat, line 444: Type of `ipos` should be `size_t`, rather than `int`. C++ standard classes (ie. `std::string`, `std::vector`) use `size_t` (unsigned 64-bit ints) to denote size. It doesn't make sense to ask for a negative position, so using `int` (signed 32-bit int) is incorrect.

Note that this is a matter of correctness and will most likely generate the same machine code. The compiler should've warned you that you're comparing an unsigned to a signed value when compiling this.

PR #722, opened 2020-08-07
Allow user specified signal to be raised from util_abort

mortalisk:
Resolves: https://github.com/equinor/scout/issues/101

### DIFF
--- a/python/ecl/util/util/install_abort_signals.py
+++ b/python/ecl/util/util/install_abort_signals.py
@@ -2,8 +2,9 @@ from ecl import EclPrototype
 import sys
 import os
 
+
 def installAbortSignals():
-    if sys.version_info.major < 3 and not os.getenv('ECL_SKIP_SIGNAL'):
+    if sys.version_info.major < 3 and not os.getenv("ECL_SKIP_SIGNAL"):
         install_signals = EclPrototype("void util_install_signals()")
         install_signals()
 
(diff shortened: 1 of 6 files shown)


jondequinor:
I'd like to test this. Any suggestion on how to effectively test?

mortalisk:
Check out my libres branch as well: https://github.com/equinor/libres/pull/980
That is right now hard coded to use this branch in unit tests.
If you install that, add some util_abort somewhere, you should see it work (hopefully)

mortalisk:
Did you manage to test this?

dotfloat:
I wanna revive this and get it merged soonish. Is the idea that *libecl* via Python will have the old behaviour of `util_abort` being useless, but for *libres* it's actually gonna dump a Python stack trace via the `faulthandler` module?

Anyway, this seems like a decent solution, although I'd put the `black`ing in its own commit to appease the github history gods.

python/ecl/util/util/install_abort_signals.py
```diff
@@ -12,6 +13,11 @@ def updateAbortSignals():
     """
     Will install the util_abort_signal for all UNMODIFIED signals.
     """
-    if sys.version_info.major < 3 and not os.getenv('ECL_SKIP_SIGNAL'):
+    if sys.version_info.major < 3 and not os.getenv("ECL_SKIP_SIGNAL"):
         update_signals = EclPrototype("void util_update_signals()")
         update_signals()
+
+
+def setSignalOnUtilAbort(sig):
```
dotfloat, line 21: snake case

mortalisk: Shouldn't it be consistent with how the file already is?

dotfloat:
I did 5 seconds of thinking and instead of letting the user set a specific signal, I think we should just set the faulthandler by default and deprecate the `install/updateAbortSignals` entirely, making them a no-op. Lets not hide the ERT error output by default.

mortalisk:
Yeah, this went into the forgetbook.
The idea is that one can have the python faulthandler invoked after util_abort has done it's thing.

Did it this way since there are other consumers of libecl. ERT would by default get both util_aborts wonders, and faulthander, once we merge the libres change related to this.

dotfloat:
I agree that we should be careful, but I reckon *how* libecl crashes isn't part of its public API. We should still have it `SIGABRT`, but otherwise it shouldn't be an issue.

mortalisk:
I'm unable to see how to do this in another way that let's us SIGABRT like before, and also invoke faulthander afterwards.

dotfloat:
Closing due to https://github.com/equinor/ecl/issues/767

PR #728, opened 2020-09-17
Add feature to retry if test path already exists when using TestArea

jepebe:
Based on a hunch from @markusdregi, this PR changes the behavior of TestArea to not fail hard if the randomly generated test path already exists. The test path is regenerated until a non-conflicting path is created.

### DIFF
--- a/lib/util/test_work_area.cpp
+++ b/lib/util/test_work_area.cpp
@@ -147,14 +147,36 @@ static bool test_work_area_copy_parent__( const TestArea * work_area , const std
   }
 }
 
-
-TestArea::TestArea(const std::string& test_name, bool store_area) :
-    store(store_area)
-{
-    char * prefix = test_work_area_alloc_prefix();
+static char *create_test_path(const std::string &test_name, const char *prefix,
+                              const char *user_name) {
+  const int MAX_TRIES = 10;
+  for (int try_count = 0; try_count < MAX_TRIES; try_count++) {
     unsigned int random_int;
-    util_fread_dev_urandom( sizeof random_int, (char *) &random_int);
+    util_fread_dev_urandom(sizeof random_int, (char *) &random_int);
     random_int = random_int % 100000000;
+
+    char *test_path = util_alloc_sprintf(
+            TEST_PATH_FMT, user_name, test_name.c_str(), random_int
+    );
+
+    char *test_cwd = util_alloc_sprintf(FULL_PATH_FMT, prefix, test_path);
+    free(test_path);
+
+    if(util_mkdir_p(test_cwd)) {
+      return test_cwd;
+    }
+    free(test_cwd);
+  }
+
+  util_abort(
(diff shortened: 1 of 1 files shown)


ertomatic:

markusdregi:
I'm positive towards this, but a first request is that it only tries a constant number of times. That is how much I trust the generator and how difficult it will be to debug an infinite loop due to a fully exhausted tmp folder. Paranoia for the win 🎏

jepebe:
Added limit fo retries to 10. Abort with `util_abort` if `try_count == 11`

lib/util/test_work_area.cpp
```diff
@@ -147,14 +147,36 @@ static bool test_work_area_copy_parent__( const TestArea * work_area , const std
   }
 }
 
+static char *create_test_path(const std::string &test_name, const char *prefix,
```
markusdregi, line 150: Standard C++ indentation in libecl is 2 spaces 🤷

markusdregi: Also, would the function read nicer as a for-loop? Just I suggestion; I leave the decision to you ;)

```c++
static char *create_test_path(
    const std::string &test_name,
    const char *prefix,
    const char *user_name
) {
  const int MAX_TRIES = 10;
  for (int try_count = 0; try_count < MAX_TRIES; ++try_count) {
    unsigned int random_int;
    util_fread_dev_urandom(sizeof random_int, (char *) &random_int);
    random_int = random_int % 100000000;

    char *test_path = util_alloc_sprintf(
        TEST_PATH_FMT, user_name, test_name.c_str(), random_int
    );

    char *test_cwd = util_alloc_sprintf(FULL_PATH_FMT, prefix, test_path);
    free(test_path);

    if(util_mkdir_p(test_cwd))
      return test_cwd;
    free(test_cwd);
  }

  util_abort(
      "%s: failed to make test directory after %d tries: %s - aborting\n",
      __func__, MAX_TRIES, test_name.c_str()
  );
  return NULL;
}
```

jepebe: > Standard C++ indentation in libecl is 2 spaces 🤷
For the rest of the file, this is true, but not for the TestArea class.

jepebe: I took your subtle "hint" and implemented it your way. With some minor changes of course.

markusdregi: It was not intended as a subtle "hint", but as an invitation to a discussion for which you are free to ignore. If you prefer the previous implementation I'm perfectly fine with merging that ;)

jepebe: No I liked your way better. Just wanted to show some attitude.

markusdregi: I like it ;)

markusdregi (approved):
Increasing test suite stability is always nice 🎉

PR #734, opened 2020-10-14
Remove timer class

jepebe:
The *timer_type* class is only used in two applications: *load_test* and *grdecl_test*. They only use basic functionality of the timer class.


dotfloat (approved):
ResInsight doesn't use this, so I think it's fine. Besides, they have the entirety of libecl in their repository for some reason, so I think we can do whatever we want anyway.

PR #736, opened 2020-10-16
Remove deprecated __int__ warning

### LINKED ISSUE #733: Implicit conversion to integers using __int__ is deprecated
Running Python3.8 with `DeprecationWarning`s turned on gives
```
WARNING:       [...]/python3.8/site-packages/cwrap/prototype.py:218: DeprecationWarning: an integer is required (got type EclFileFlagEnum).  Implicit conversion to integers using __int__ is deprecated, and may be removed in a future version of Python.
  return self._func(*args)
```

Not 100% sure if it belongs here or if it should be transferred to `libecl`?

```
pip list
libecl                       2.9.1
cwrap                        1.6.1
```

jepebe:
Protoyping with an int instead of the expected enum results in a deprecated __int__ warning on newer Python version. This should resolve #733

### DIFF
--- a/python/ecl/ecl_util.py
+++ b/python/ecl/ecl_util.py
@@ -91,10 +91,13 @@ EclUnitTypeEnum.addEnum("ECL_PVT_M_UNITS", 4)
 #-----------------------------------------------------------------
 
 class EclFileFlagEnum(BaseCEnum):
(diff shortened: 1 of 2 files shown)


oysteoh:
@markusdregi  is this something that could be merged?

dotfloat (approved):

PR #742, opened 2020-10-26
Publish ecl applications to PyPI

dotfloat:
Expose only `ecl_pack.x` and `ecl_unpack.x` because that is what subscript uses.


sondreso (approved):
LGTM!

Issue #710, opened 2020-05-14
Deploy to PyPI as `ecl` rather than `libecl`


dotfloat:
We're currently deploying this package as `libecl` when we should probably deploy it as `ecl` since we own that name. We want to use `ecl` because it has the same name as the module. There are several ways to do this, and I'd say only 2 decent ones:

1. Deploy to `ecl`, make `libecl` be a dummy package with a high version number that depends only on `ecl` and prints a deprecation warning when installed.
2. Ditto, but keep `ecl` and `libecl` at the same version just in case.
3. (Bad) deploy to both `ecl` and `libecl`. There'll be confusion about which one is which, and they'd override each other when installed, leading to major problems.

The 2. point can be done with making generating the `libecl` wheel at the `before_deploy` step of Travis.

pgdr:
I'm voting for 2.

markusdregi:
I'm also voting for 2

PR #744, opened 2020-10-27
Rename libecl to ecl

dotfloat:
Resolves: #710

Also add `libecl` package that installs `ecl` of the same version

---

To test libecl -> ecl:
1. Go to https://github.com/equinor/libecl/pull/744/checks
2. Click on the artifacts
3. Download `libecl -> ecl` zip
4. Download the zip for your version of Python
5. Unzip both
6. `pip install ./libecl-*.whl -f .`

Both `ecl` and `libecl` should now be installed.

When this is uploaded to PyPI it will suffice to write `pip install libecl` to get the correct `ecl`.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -1,4 +1,4 @@
-name: Libecl testing
+name: ecl testing
 
 on: [push, pull_request]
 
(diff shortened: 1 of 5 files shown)


sondreso (approved):
LGTM, remmber to squash!

PR #746, opened 2020-11-03
 Fix segfault with coarse cell group id

jepebe:
Coarse cell group numbering is not necessarily in consecutive order. A dataset that skips, for example, index 0 for coarse cell group id and rather starts at 1 will fail hard when ecl_file reads the file. Since ecl_file assumes that all groups ids are present it segfaults when it tries to read index 0 which is, of course, NULL.


markusdregi (approved):
🚢

PR #752, opened 2020-11-18
Find libecl.dll on Windows

crayxt:
It seems like when installed from a wheel (built manually, as it is missing from pypi), it is not possible to load package on Windows, due to platform being not listed.


ertomatic:

dotfloat (approved):
Looks good, thanks a bunch! :)

PR #757, opened 2020-11-26
Run the c tests on komodo pr

eivindjahren:
Currently the komodo job is responsible for running the internal test data, so it must also run the c tests.


dotfloat (approved):
Remember to squash

PR #759, opened 2020-11-27
Use non-equinor data whenever possible

eivindjahren:
Several of the tests ran by cmake does not rely on any specific data. To reduce the amount of complexity, these either just use generated data or are moved to using data from test-data/ECLIPSE whenever possible.

Some tests of internal data are removed simply because there was already a counterpart using generated data.


eivindjahren:
retest this please

lib/CMakeLists.txt
```diff
@@ -458,29 +457,32 @@ foreach(name ecl_coarse_test
   add_executable(${name} ecl/tests/${name}.cpp)
   target_link_libraries(${name} ecl)
   target_include_directories(${name} PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/private-include)
-endforeach()
+endforeach(name)
```
dotfloat, line 460: Miiinor style nitpick, but I personally dislike having anything inside the parens of block-end tags. Ie, `endforeach(name)` -> `endforeach()`. I'm up for discussing this.

dotfloat:
Why does this delete all of those `add_test`s?

eivindjahren:
So it either replaces an `add_test` with one running on `test-data/local` or it is removed because it is already ran on generated data.

dotfloat (approved):

PR #764, opened 2020-12-16
Limit pytest to python/tests

eivindjahren:
Before pytest could look through test-data/Equinor which is
quite big and on disk.


lars-petter-hauge (approved):
Lgtm!

Issue #648, opened 2019-08-28
TestAreaContext deletes tmp dir and enters infinite loop


mortalisk:
While working on [libres#653](https://github.com/equinor/libres/pull/653), we encountered this bug. When running pytest in the new test folder (job_runner) we created there, i.e. to run all the tests, TestAreaContext would consistently enter an infinite loop during construction. We don't know why, but this happens only with tests that are in the new folder python/tests/job_runner. It only happens when running pytest without specifying a test, thus letting it discover tests for itself. We have now used other means of creating a tmp working dir in those tests, but it should be possible to reproduce the error by trying to use TestAreaContext in some of those tests in libres.

It seems that the tmp dir that it creates and then chdirs to, is then deleted. After this, it tries to find the current working directory, which then fails with ENOENT. ENOENT is not handled, and the code enters a never ending loop. This happens in the function util_alloc_cwd() in util.c.

https://github.com/equinor/libecl/blob/e6d02534589d3ab0922d622a06e5adfd948ed3d1/lib/util/util.c#L694-L710

So, there are two issues it seems:

1. the tmp dir is for some reason deleted
2. the function util_alloc_cwd does not handle other errors than ERANGE, and thus enters a infinite loop befause of 1

dotfloat:
https://man7.org/linux/man-pages/man2/getcwd.2.html

mortalisk:
I was unable to reproduce the deleting of tmp dir issue. But I have at least removed the infinite loop issue in PR #776

sondreso:
Should we close this now?

PR #776, opened 2020-12-18
Fix bug in util_alloc_cwd that would loop for ever

mortalisk:
The code was assuming it could never fail to get cwd, but this is
actually possible in certain cases.

Remove the loop by allocating a large enough array to begin with,
and call util_abort, since the caller probably expects this to
always return a valid path.

**Issue**
Might solve  #648

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -70,7 +70,7 @@ add_library(ecl util/rng.cpp
                 util/hash.cpp
                 util/node_data.cpp
                 util/node_ctype.cpp
-                util/util.c
+                util/util.cpp
                 util/util_symlink.cpp
                 util/util_lfs.c
                 util/util_unlink.cpp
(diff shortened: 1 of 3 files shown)


sondreso:
Jenkins test this please

lib/util/util.c
```diff
@@ -693,21 +693,15 @@ static char * util_getcwd(char * buffer , int size) {
...
-        free(cwd);
-      }
-    }
-  } while ( result_ptr == NULL );
-  cwd = (char*)util_realloc(cwd , strlen(cwd) + 1 );
-  return cwd;
+  char cwd[PATH_MAX];
+  char * result_ptr = util_getcwd(cwd , sizeof(cwd));
+  if (result_ptr != NULL) {
+    char *result = (char *)malloc(strlen(cwd)+1);
```
dotfloat, line 699: `strdup`: duplicate string

```c
return strdup(cwd);
```

dotfloat:
Does this fix a flaky test or is it possible to confirm that this solves #648 ? Is it possible to get a test for this? (Use the new catch2 framework, @ me if you need help with it :pear: )

mortalisk:
Jenkins test this please

mortalisk:
Jenkins test this please

mortalisk:
> Does this fix a flaky test or is it possible to confirm that this solves #648 ? Is it possible to get a test for this? (Use the new catch2 framework, @ me if you need help with it 🍐 )

Not sure how to test it, really.

dotfloat (approved):

PR #788, opened 2020-12-21
Fix uninitialized header8

eivindjahren:
If a header longer than 8 characters is created header8 becomes uninitialized memory, making the python implementation crash completely. Changed it so that header8 is always the 8 first characters of the keyword header, while header is the entire string.

**Issue**
https://github.com/equinor/ecl/issues/775

### DIFF
--- a/lib/ecl/ecl_kw.cpp
+++ b/lib/ecl/ecl_kw.cpp
@@ -16,6 +16,8 @@
    for more details.
 */
 
+#include <algorithm>
+
 #include <stdlib.h>
 #include <stdio.h>
 #include <stdbool.h>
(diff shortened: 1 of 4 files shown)


lib/ecl/ecl_kw.cpp
```diff
@@ -1531,17 +1531,10 @@ void ecl_kw_free_data(ecl_kw_type *ecl_kw) {
...
-     /* Internalizing a header without the trailing spaces as well. */
-     free( ecl_kw->header );
-     ecl_kw->header = util_alloc_strip_copy( ecl_kw->header8 );
-  }
-  else {
-     ecl_kw->header = (char*)util_alloc_copy(header, strlen( header ) + 1);
-  }
+  free( ecl_kw->header );
 
+  sprintf(ecl_kw->header8 , "%.*s\0" , ECL_STRING8_LENGTH, header);
```
dotfloat, line 1536: `strncpy(ecl_kw->header8, header, ECL_STRING8_LENGTH + 1);` ?

dotfloat (approved):

lib/ecl/ecl_kw.cpp
eivindjahren: Possibly, but the damn thing is very sensitive.

PR #791, opened 2020-12-22
Avoid sphinx 3.4.0 due to bug

eivindjahren:
ecl docs does not currently work with sphinx version 3.4.0, because of sphinx-doc/sphinx#8568

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -22,7 +22,7 @@ jobs:
     - name: Build ecl
       run: |
         mkdir cmake-build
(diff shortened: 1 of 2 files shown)


dotfloat (approved):

markusdregi:
Nice! Could consider explicitly avoiding `3.4.0` as in #792 instead as there seems to be a fix on the way in `3.4.1` :)

eivindjahren:
> Nice! Could consider explicitly avoiding `3.4.0` as in #792 instead as there seems to be a fix on the way in `3.4.1` :)

Done

PR #795, opened 2021-01-12
Add missing utils to Python package

### LINKED ISSUE #794: Add missing ecl binaries to the pypi package
Currently, only `pack.x` and `unpack.x` are officially published to PyPI due to their use in https://github.com/equinor/subscript

The following has also been identified by @jcrivenaes and @eivindsm as missing:
* `convert.x`
* `summary.x`
* `rft.x`
* `kw_list.x`
* `grid_info.x`

Are there any more?

dotfloat:


ManInFez:
is there a test that checks if the ".x" binaries are available?

dotfloat:
Nope. Do we want a test for this? Note that there are no existing tests for the utilities and so we can't test if they function as intended. All we can do is check for existence and that's not particularly satisfying.

ManInFez (approved):
🥓

Issue #802, opened 2021-02-10
Can't write UNSMRY files with from_pandas() beyond year 2262


berland:
The `EclSum.from_pandas()` function goes via Pandas  datetime, which is limited to year 2262 due to nanosecond precision. This is fixable as all we need is `datetime` object which supports longer timespans.

PR #804, opened 2021-02-11
Circumvent Pandas bugs for years > 2262

berland:
**Issue**
#802 

**Approach**

Pandas is very eager to convert numpy datetime64[*] timestamps to datetime64[ns], and will error hard if a datetime64[ms] timestamp is beyond year 2262. Avoid this by giving Pandas datetime.datetime objects instead

### DIFF
--- a/python/ecl/summary/ecl_sum.py
+++ b/python/ecl/summary/ecl_sum.py
@@ -513,7 +513,7 @@ class EclSum(BaseCClass):
             raise ValueError("No valid key")
 
         if time_index is None:
-            time_index = self.numpy_dates
+            time_index = self.dates
             data = numpy.zeros([len(time_index), len(keywords)])
             EclSum._init_pandas_frame(self, keywords,data.ctypes.data_as(ctypes.POINTER(ctypes.c_double)))
         else:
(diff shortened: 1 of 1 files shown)


ertomatic:

berland:
Related to https://github.com/pandas-dev/pandas/issues/39727

dotfloat:
Can you write a unit test that tests that this does the right thing and also that existing user code doesn't break with this change? Thanks.

berland:
> Can you write a unit test that tests that this does the right thing and also that existing user code doesn't break with this change? Thanks.

Not sure I dare to touch the test-code in here. Identical codelines are tested in `ecl2df`.

lars-petter-hauge:
The pandas issue was closed (as fixed) 2 weeks ago, https://github.com/pandas-dev/pandas/issues/39727, is this still relevant for us?

berland:
I guess it is relevant until we have Pandas 2.0 (not released) in our stable environment, which is hard to predict when is feasible.

hnformentin:
@berland cam you take a look at this PR? Tests are not running and it seems to need a rebase main.

PR #805, opened 2021-02-15
Fix EclSum.iget example in docs

markusdregi:
**Issue**
The order of the arguments was wrong in the documentation. Check vs implementation to verify that it is now correct.

**Approach**
Opened the file in vim, permuted the keywords, a `:wq` and some git usage and here we are 🤷


dotfloat (approved):
Your approach is incorrect since you should be using Emacs. I'll let it slide for now though. 🚀

markusdregi:
There is always evil mode to save the lost ones ;)

PR #807, opened 2021-02-19
Use 0.0 instead of -99 to represent an undefined value

magnesj:


ertomatic:

sondreso:
Jenkins, test this please!

dotfloat:
Can you explain why you want this change? And why something like `NaN` wouldn't be a better fit?

magnesj:
When a new well is introduced based on a restart case, all summary values for this well earlier than the introduced time step are reported as -99. For plotting and data report in ResInsight, it is easier to use a value of 0.0 directly(reservoir engineers I have talked to say that 0.0 is an OK value to use for all summary data). NaN values can be difficult to handle in some situations(ie plotting using Qwt), and a value of 0.0 will be easier and less error prone than NaN.

markusdregi:
> a value of 0.0 will be easier and less error prone than NaN

Is that really true? Would it just not make it impossible to differentiate between an undefined value and a value actually being zero?

magnesj:
From a ResInsight point of view, a value of 0.0 is much better than -99. We do resampling and statistical calculations on this data too, and these algorithms must be updated if NaN is used. A value of 0.0 works without any other known side effects. Agree that NaN is the correct solution, but 0.0 is the easiest and a good enough pragmatic value.

markusdregi:
But do you not want to separate between undefined values and `0.0`? This might heavily impact your statistical calculations; expected values, variance and others...

magnesj:
I see this PR has some unintended side effects, I will address this issue in the import step in ResInsight instead in ecl.

markusdregi:
Thanks 👍

PR #815, opened 2021-06-15
Remove call to ecl_util_get_start_date

mortalisk:
**Issue**
partially-resolves https://github.com/equinor/ert/issues/1319 together with equinor/ert#1856

**Approach**
remove the calls to ecl_util_get_start_date

### DIFF
--- a/lib/ecl/ecl_util.cpp
+++ b/lib/ecl/ecl_util.cpp
@@ -1061,10 +1061,8 @@ time_t ecl_util_get_start_date(const char *data_file) {
     FILE *stream = util_fopen(data_file, "r");
     char *buffer;
 
-    if (!basic_parser_fseek_string(parser, stream, "START", true,
-                                   true)) /* Seeks case insensitive. */
-        util_abort("%s: sorry - could not find START in DATA file %s \n",
-                   __func__, data_file);
+  if (!basic_parser_fseek_string( parser , stream , "START" , true , true))   /* Seeks case insensitive. */
+    return start_date;
 
     {
         long int start_pos = util_ftell(stream);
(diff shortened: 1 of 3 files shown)


verveerpj (approved):
Looks good to me, but one question: would it make sense to log some kind of warning?

kvashchuka:
@mortalisk you also might want to change the commit message ;)

mortalisk:
> Looks good to me, but one question: would it make sense to log some kind of warning?

I don't know. Depends on whether this is something we want to discourage, or if it is perfectly normal thing to do. @markusdregi what do you think?

markusdregi:
I'm slightly sceptical towards merging this PR in its current state for multiple reasons.

First, most systems will handle a `time_t` of value `-1` as the end of `1969`. Which is a perfectly fine date for being the start of a simulation 🤷🏻 Tearing up in flames is a bad approach for a library, but returning seemingly ok values that potentially leads to wrong calculations is even worse...

Second, there are better libraries for parsing the `DATA`-files for Eclipse (like the opm-libraries). I'm not sure whether we should solidify or start removing `DATA`-file capabilities instead.

Third, according to the comment right above the function Eclipse has default start dates when a date is not specified. If anything, we should try to adhere to those...

mortalisk:
Have tried to run this on rgs node with setup from issue, which seems to work fine

markusdregi:
Since the start date is parsed only when requested I don't see how this PR is needed for the issue referred to to be solved. If we are to remove this functionality in `ecl` I think we should:

- make a conscious decision on whether we want to in general decrease the amount of input deck parsing we support,
- deprecate the functions first and
- confer with our largest consumer besides ERT, namely ResInsight on how this would impact them before we do any of the two above.

Due to this I don't think this PR should be merged in its current form at this very time.

mortalisk:
Closed according to recomendation to not merge this.

PR #817, opened 2021-06-28
Remove the deprecated ert and ecl.ecl python packages

### LINKED ISSUE #790: Don't install the 'ert' top-level module when using pip
Currently, the `ert` module is installed when it shouldn't be. This is due to `setup.py`'s `find_packages` being overzealous. It should filter it out.

verveerpj:
**Issue**
Resolves #790 
This issue is now blocking an ERT issue: https://github.com/equinor/ert/issues/1786

**Approach**
Remove the `ert` and `ecl.ecl` deprecated packages.


ertomatic:

markusdregi:
I suggest that we start by only removing the `ert` package in this PR.

DanSava (dismissed):
Looks good to me

markusdregi:
Wait a minute ⌛

verveerpj:
I have now also nuked the `ecl.ecl` package. It is a separate commit, so we can change our mind again :smile:.

Joking aside, I can squash them, but if we leave them as two separate commits, reverting to keep `ecl.ecl` but not `ert` will be easier.

markusdregi (approved):
Two commits is perfect 👌🏻

Issue #774, opened 2020-12-17
Define a C++ code style


dotfloat:
Python code uses black already.

We should create a clang-format file for C++ code.

magnesj:
You can use GitHub actions to automatically create a separate clang-format PR if there are clang-format issues in your current PR. See here for an example use in ResInsight
https://github.com/OPM/ResInsight/blob/dev/.github/workflows/clang-format.yml

PR #818, opened 2021-07-03
Add style formatting to CI

dotfloat:
This PR is made of several commits which I don't want to squish, for reasons of the massive amounts of code changes.

The first commit (`Refactor clang-format`) adds a `script/clang-format` script which behaves similarly to `black`. This was used to avoid formatting anything in `external/` (`catch2.hpp` exposed some differences between the different versions of `clang-format`). Due to unknown reasons, there is no backwards-compatibility guarantee with `clang-format`. However, the current changes should be compatible with all current versions of LLVM. I've not whitelisted version 13, since it's not released yet, but it seems to also not break anything.

The following commits each apply each their own formatting:
* C: Ran `script/clang-format` twice using version 12.
* Python: Ran `black setup.py` and `black python/ --exclude docs`. The reason for excluding docs is that they're written for Python 2 and have rotted to bits.
* CMake: Ran `cmake-format -i` on each of `*.cmake` and `CMakeLists.txt` using `find`.

The final commits adds a CI job that tests all of these. The CI job has been tested previously by submitting it running without the reformat commits.

Resolves: #774


eivindjahren (approved):
Nice! Big improvement, while we are doing cosmetic stuff, at least ecl_grid.cpp ends with a single blank comment and some new lines, might want to search for those aswell.

PR #823, opened 2021-08-13
Use C++ to define vectors rather than CMake

dotfloat:


lib/util/vector_template.cpp
```diff
@@ -0,0 +1,1768 @@
...
+                       __func__);
+    }
+}
+
+template <typename T> void vector<T>::memmove(int offset, int shift) {
+    if ((shift + offset) < 0)
+        util_abort("%s: offset:%d  left_shift:%d - invalid \n", __func__,
+                   offset, -shift);
+
+    if ((shift + this->size > this->alloc_size))
```
eivindjahren, line 303: Maybe fix this double parenthesis while here?

lib/util/vector_template.cpp
```diff
@@ -0,0 +1,1768 @@
...
+        else {
+            util_abort(
+                "%s: index:%d is invalid - only accepts positive indices.\n",
+                __func__, index);
+            return -1;
+        }
+    }
+}
+
+/** Will abort is size == 0 */
```
eivindjahren, line 645: Think this should be "if size == 0".

lib/util/vector_template.cpp
```diff
@@ -0,0 +1,1768 @@
...
+template <typename T> T vector<T>::get_last() const {
+    return this->iget(this->size - 1);
+}
+
+/** Will abort is size == 0 */
```
eivindjahren, line 650: Think this should be "if size == 0".

lib/util/vector_template.cpp
```diff
@@ -0,0 +1,1768 @@
...
+}
+
+template <typename T> const void *vector<T>::get_const_ptr() const {
+    return this->data;
+}
+
+/**
+   Observe that there is a principle difference between the get_ptr()
+   functions and alloc_data_copy() when the vector has zero size. The
+   former functions will always return valid (@TYPE@ *) pointer,
```
eivindjahren, line 823: The `(@TYPE@ *)` loooks weird now. Should at least be updated to `T*`. I think it the entire `alloc_data_copy` and its documentation should just be removed. code scars aren't really that useful in my opinion, and we are seriously unlikely to want to add that symbol to the public API at this point.

eivindjahren (changes requested):
Seems like a good change. I had some minor comments, which was more to clean up some of the old code while we are here.

I know there are no tests for this code currently, but a vector library should be pretty easy to test so writing some catch2 tests would be good.

PR #832, opened 2021-10-14
Mzran read

dafeda:
Documenting `mzran` and proposing test that draws 200 samples instead of just 2.


ertomatic:

PR #834, opened 2021-10-26
Simplify installation on corporate network

### LINKED ISSUE #833: Clean build of C libs + Python package fails with Conan problems
**Steps to reproduce**

```
git clone ..
mkdir ecl/build
cd ecl/build
pip install -r ../requirements.txt

cmake ..
```
 

1. `conan.cmake` requires an explicitly set `CMAKE_BUILD_TYPE`:

```
...
-- Downloading conan.cmake from https://github.com/conan-io/cmake-conan
-- Conan: Automatic detection of conan settings from cmake
CMake Error at build/conan.cmake:71 (message):
  Please specify in command line CMAKE_BUILD_TYPE
  (-DCMAKE_BUILD_TYPE=Release)
  build/conan.cmake:239 (_conan_detect_build_type)
  build/conan.cmake:758 (conan_cmake_settings)
  CMakeLists.txt:38 (conan_cmake_run)
```
passing `-DCMAKE_BUILD_TYPE=xxx` or adding a `set(CMAKE_BUILD_TYPE "XXX")` solves it - not clear to me what is the best way to fix this.

2. Can not find `conan` executable

Install `conan` with `pip install` solves it - but I guess information about conan could be added to the documentation - or alternative to something like `dev-requirements.txt` - if you indicate what kind of fix you wish I am happy to provide a PR.

dotfloat:
This commit sets `CONAN_CACERT_PATH` to the RHEL system-wide
certificates path, if available.

We also modified the `CMakeLists.txt` to set `CMAKE_BUILD_TYPE` to
`RelWithDebInfo` so as to stop `conan.cmake` from complaining.

Resolves #833

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -27,7 +27,7 @@ jobs:
     - name: Build ecl
       run: |
         mkdir cmake-build
-        cmake -S . -B cmake-build -DBUILD_TESTS=ON -DCMAKE_BUILD_TYPE=RelWithDebInfo -DRST_DOC=ON
+        cmake -S . -B cmake-build -DBUILD_TESTS=ON -DRST_DOC=ON
         cmake --build cmake-build
 
     - name: Run tests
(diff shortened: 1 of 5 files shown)


eivindjahren (approved):
Looks great! Install catch2 via conan right around the corner?

PR #843, opened 2021-11-24
Store result of fgetc to int

### LINKED ISSUE #842: Result of of `fgetc` should be put into an int
Saving output of `fgetc` to `char` and comparing it to `EOF` will fail on Ubuntu aarch (both gcc and clang).

https://github.com/equinor/ecl/blob/ef756e07bc30ce4d9d3cd23b849a09bd1bedcf08/lib/util/util.cpp#L578-L580

Try for example running this code on aarch:

```c
#include <iostream>
#include <stdio.h>

int main() {
	FILE *stream = fopen("test.txt", "r");
	bool cont;
	cont = true;
        char c;
	std::cout << EOF << std::endl;
	do {
		c = fgetc(stream);
		std::cout << c << std::endl;
		if (c == EOF) cont = false;
	} while (cont);
}
```

dafeda:
**Issue**
Resolves #842 

Runs on aarch Ubuntu 💯

### DIFF
--- a/lib/util/util.cpp
+++ b/lib/util/util.cpp
@@ -568,14 +568,14 @@ static char *util_fscanf_alloc_line__(FILE *stream, bool *at_eof, char *line) {
     int init_pos = util_ftell(stream);
     char *new_line;
     int len;
-    char end_char;
+    int end_char;
     bool cont;
     bool dos_newline;
 
     len = 0;
     cont = true;
     {
-        char c;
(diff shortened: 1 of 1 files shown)


ertomatic:

sondreso:
Jenkins add to whitelist

dotfloat (approved):
🤘

PR #851, opened 2021-12-27
Document unexpected date parsing through tests.

berland:
ISO dates will be parsed with unexpected results when
given to the function util_sscanf_date_utc, as it will not
attempt to validate.


ertomatic:

sondreso (approved):
LGTM!

PR #858, opened 2022-03-28
Called with wrong arguments (don't explicitly pass self)

BjarneHerland:
The wrapped function `bool_vector_count_equal()` should not explicitly pass `self` since that is done implicitly. No idea why this works at all...

### DIFF
--- a/.github/workflows/style.yml
+++ b/.github/workflows/style.yml
@@ -13,7 +13,7 @@ jobs:
 
     - name: Install dependencies
       run: |
-        sudo pip install cmake-format black==21.12b0
+        sudo pip install cmake-format black==22.1.0 'click<8;python_version<="3.6"' 'click==8.0.2;python_version >"3.6"'
 
     - name: Clang Format
       run: ./script/clang-format --check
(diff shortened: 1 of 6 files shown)


berland:
This looks to actually resolve a bug uncaught in current testing:

```
173     def test_bool_vector(self):
174         b = BoolVector()
175         b.setDefault(True)
176
177         b[4] = False
178
179         self.assertEqual(list(b), [True, True, True, True, False])
180         self.assertEqual(b.count(True), 4)
181         self.assertEqual(b.count(False), 1)
```

Adding the lines 180 and 181, it will fail unless this patch is applied.

BjarneHerland:
It's also worth noting that most of the wrapped methods in this class are not used, e.g. `_alloc`, `_alloc_copy`, `_assign`, `_equal`, `_sort_perm` etc.

berland:
Can you rewrite commit message to state that this fixes a bug with count_equal in a BoolVector (and the other change)

berland (approved):
👍🏻

PR #860, opened 2022-03-31
Fixes argument-bug in BoolVector and adds test

BjarneHerland:
The wrapped function bool_vector_count_equal() should not explicitly pass self since that is done implicitly. No idea why this works at all...


berland (approved):
👍🏻

PR #862, opened 2022-04-20
Fix ecl_util_get_num_cpu function

verveerpj:
**Issue**
The `ecl_util_get_num_cpu` function is used by ERT to parse the `PARALLEL` keyword. This fails if the `PARALLEL` word appears in the `TITLE` keyword value. The result is that ERT detects the number of cpus as equal to one in such a case.

This PR is required to be able to resolve this `ERT` issue: https://github.com/equinor/ert/issues/1905

**Approach**
While scanning the file for `PARALLEL` keywords, detect if they occur in a title and so, skip them.

**Note**
This PR also pins manylinux2014, since cause failure, as was done for `ERT`.

### DIFF
--- a/lib/ecl/tests/data/num_cpu5
+++ b/lib/ecl/tests/data/num_cpu5
@@ -1,0 +1,18 @@
+----------------------------------------------------------------------
+-- A very simple test case for the shared memory parallel option.
+-- --------------------------------------------------------------
+
+TITLE
+foo
+
+TITLE
+PARALLEL BENCHMARK
+
+TITLE
+foo
+
+PARALLEL
+-- TITLE
+-- foo
+  4  DISTRIBUTED /
+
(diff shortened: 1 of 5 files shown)


ertomatic:

sondreso:
Jenkins add to whitelist

lib/ecl/tests/data/num_cpu5
```diff
@@ -0,0 +1,16 @@
...
+TITLE
+foo
+
+TITLE
+PARALLEL BENCHMARK
+
+TITLE
+foo
+
+PARALLEL
```
berland, line 14: Could a `-- TITLE` appended to this line (or the line below) make some havoc?

verveerpj: It should not, but lets test that. I will add it to the file.

verveerpj: Seems to be fine.

berland (approved):
LGTM

PR #866, opened 2022-04-26
Add windows build to release albeit without testing.

### LINKED ISSUE #726: Using pip install libecl on Window?
Hello, I am wondering if there is any clearer instruction of installing libecl on window? Thank you.

trhallam:
**Issue**
Resolves #726 

**Approach**

Include windows MSVC build environment in github workflow but exclude windows from testing due to an error in the creation  of temporary memory.

I'm currently building these as unofficial artifacts for windows but if you feel they could be included in the main trunk then I make this pull request available.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -71,10 +71,13 @@ jobs:
       with:
         python-version: ${{ matrix.python }}
 
-    - name: Build Windows Wheel
+    - name: Setup VS-devshell
       uses: egor-tensin/vs-shell@v2
       with:
         arch: x64
+      if: matrix.os == 'windows-latest'
+
+    - name: Build Windows Wheel
       run: 
         cl.exe
         pip wheel . --no-deps -w dist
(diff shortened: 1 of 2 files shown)


ertomatic:

lars-petter-hauge:
Hi @trhallam

Thanks for the PR! Unfortunately we can not ship releases for a platform that does not have tests running. I can understand the frustration of having to compile locally every time you want to install the package though.. I believe workarounds like WSL has already been mentioned in the issue you are referring to.

trhallam:
Ok, no problem, if you fix the issue where the custom test area is hard coded into the tests to suit linux then I think the tests might pass. I'll just continue to have the shadow releases in my fork. 

I'd also suggest that WSL is not always a solution for people because some IT departments are still wary of providing WSL support. People with less knowledge also struggle to understand some of the nuance of WSL which makes ECL less accessible.

Non-the-less, appreciate all the work that has gone into this package.

PR #878, opened 2022-10-20
Restrict ISO parsing to be closer to standard

### LINKED ISSUE #877: ISO dates are parsed differently on mac vs linux
On Mac, the string "17-01-2017" passes as an ISO date, while on Linux it does not. This is likely from a difference in `timegm`.

Linux marks this as a not valid date by coincidence, since the iso parser in ecl does not require ~at least~ exactly four digits for the string:

https://github.com/equinor/ecl/blob/c69f900fcb53b6e99012f661e56198999e65fb62/lib/util/util.cpp#L2656

berland:
ISO8601 requires exactly four digits for the year, while the current ISO parser in ecl allowed any number of digits. This commit will restrict the parser on this detail.

**Issue**
Resolves #877 

**Approach**
Add an extra `sscanf` that must pass before an ISO date is validated.


ertomatic:

hnformentin:
How do you think this will impact users? Should we add labels for releases-notes to communicate it?

berland:
> How do you think this will impact users? Should we add labels for releases-notes to communicate it?

Not sure, the bug fixed by this is only present on Mac, on which we have no users, only developers (?)

hnformentin (approved):
LGTM! Good job identifying this issue.

PR #891, opened 2022-12-21
clang-format: Remove version check

berland:
This commit also speeds up `clang-format` when installed from PyPI. It detects whether the `clang-format` executable is a Python wrapper and finds where the actual binary is location.

Co-authored-by: Zohar Malamant <ZOM@equinor.com>

(This is the same as https://github.com/equinor/ert/commit/0c54472c550e0cc133a2e77418a036e15a662337)

### DIFF
--- a/.github/workflows/style.yml
+++ b/.github/workflows/style.yml
@@ -9,7 +9,7 @@ jobs:
     runs-on: "ubuntu-latest"
 
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
 
     - name: Install dependencies
       run: |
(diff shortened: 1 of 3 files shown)


ertomatic:

lars-petter-hauge:
Quick question, when would one want to clang-format the code if one has installed it through pypi? I'm just missing something..

berland:
> Quick question, when would one want to clang-format the code if one has installed it through pypi? I'm just missing something..

The text concerns the case when `clang-format` is installed from pypi, not when `ecl` is installed from pypi.

lars-petter-hauge (approved):

PR #886, opened 2022-12-05
Use Catch2 from conan

pinkwah:

PR #892, opened 2022-12-23
Pin windows compile to windows-2019 image

berland:
Last 2022-image that works is 20221127.1. Not possible to pin to this image exactly. First 2022-image that fails is 20221215.2


ertomatic:

berland:
Closing, as this problem is due to #886

PR #897, opened 2023-01-19
Circumvent Pandas bugs for years > 2262

### LINKED ISSUE #802: Can't write UNSMRY files with from_pandas() beyond year 2262

berland:
**Issue**
Resolves #802 

Taking over from #804


hnformentin (approved):
LGTM!

PR #903, opened 2023-05-11
Update location of testkomodo.sh

berland:
The location inside jenkins has been deprecated in komodo-releases


xjules (approved):
:rocket:

PR #909, opened 2023-09-27
Use datetime.utcfromtimestamp instead of first using gmtime in CTime.datetime

JHolba:
This makes datetime run about 10% faster.

measured using this python code
```
import timeit

runtime = timeit.timeit(
    "time_map = [t.datetime() for t in c_time]",
    "from ecl.summary import EclSum;"
    'summary = EclSum("/home/jholba/Downloads/BIGPOLY",include_restart=False,lazy_load=False);'
    "c_time = summary.alloc_time_vector(True)",
    number=100,
)

print(runtime)
```

### DIFF
--- a/python/ecl/util/util/ctime.py
+++ b/python/ecl/util/util/ctime.py
@@ -50,7 +50,7 @@ class CTime(BaseCValue):
         return datetime.date(*self.time()[0:3])
 
     def datetime(self):
-        return datetime.datetime(*self.time()[0:6])
+        return datetime.datetime.utcfromtimestamp(self.ctime())
 
     def __str__(self):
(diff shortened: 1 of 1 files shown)


python/ecl/util/util/ctime.py
```diff
@@ -50,7 +50,7 @@ def date(self):
         return datetime.date(*self.time()[0:3])
 
     def datetime(self):
-        return datetime.datetime(*self.time()[0:6])
+        return datetime.datetime.utcfromtimestamp(self.ctime())
```
berland, line 53: Maybe not measurably faster, but by inlining `self.ctime()` to `self.value()` you should save a function call.

pinkwah:
Stale?

PR #912, opened 2023-10-25
Rename to Resdata

eivindjahren:


andreas-el (approved):
As discussed, there is no way we can guarantee the integrity of this change by review.
Going to approve this (as agreed upon), and we should have a keen eye on the lookout for issues that may have been caused by this. 🚀

PR #930, opened 2023-11-08
Remove unused ert_test_context

eivindjahren:
Since importing the file in question has raised an `ImportError` since 3.0 and will continue to raise an `ImportError`, but with a different message since deletion, there is no need for a major version increment because of this change.


sondreso (approved):
Will you give it a slightly better commit title? The "broken so likely unused" can be moved to the body I think :slightly_smiling_face:

PR #947, opened 2024-01-19
make tests compatible with python3.11

### LINKED ISSUE #946: Some tests fail on python3.11 when running with test data onprem (Equinor)
Following test fails:
```
ResdataFileEquinorTest.test_index
SummaryTest.test_truncated_data
SummaryTest.test_truncated_smspec
```
The reason for the failing tests seems to be that python3.11 is stricter about type conversion than python3.8. Some of the failing tests are probably better solved in cwrap directly. However some failures are due to the following:
```
self = <tests.rd_tests.test_rd_sum.SummaryTest testMethod=test_truncated_smspec>

    def test_truncated_smspec(self):
        with TestAreaContext("Summary/truncated_smspec") as ta:
            ta.copy_file(self.test_file)
            ta.copy_file(self.createTestPath("Equinor/ECLIPSE/Gurbat/ECLIPSE.UNSMRY"))
    
            file_size = os.path.getsize("ECLIPSE.SMSPEC")
            with open("ECLIPSE.SMSPEC", "r+") as f:
>               f.truncate(file_size / 2)
E               TypeError: 'float' object cannot be interpreted as an integer

python/tests/rd_tests/test_rd_sum.py:88: TypeError
```
, and should probably use (// instead of /).

larsevj:
**Issue**
Resolves #946 

Converts filesize to int by using "//" instead of "/".

### DIFF
--- a/python/tests/rd_tests/test_grid.py
+++ b/python/tests/rd_tests/test_grid.py
@@ -312,7 +312,7 @@ class GridTest(ResdataTest):
 
             size = os.path.getsize("TEST.EGRID")
             with open("TEST.EGRID", "r+") as f:
-                f.truncate(size / 2)
+                f.truncate(size // 2)
 
             with self.assertRaises(IOError):
                 Grid("TEST.EGRID")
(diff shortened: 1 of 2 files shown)


sondreso (approved):

sondreso (approved):
LGTM, update commit message slightly before merging

PR #950, opened 2024-01-26
Fix a buffer overflow in restart summary read

berland:
The number of blocks of 8 character strings in the RESTART keyword can now be more than 9, and in any case the file could always have the length set to something larger, leading to stack corruption.


eivindjahren (approved):
:+1:

PR #956, opened 2024-02-20
Add ninja to the build requirements

larsevj:
Currently ninja is listed as a dependency in requirements.txt and installed in CI, but is not listed in the build dependencies in pyproject.toml.


eivindjahren (approved):
:+1:

PR #964, opened 2024-04-09
Remove remnants of fault_block_collection

eivindjahren:
fault block collection was removed 10 years ago
in https://github.com/equinor/resdata/commit/b426325ad8808ec374c3ca390e95ff3b30e48996 but these files have remained.


berland (approved):
🧹

PR #969, opened 2024-04-22
Fix some issues with deselect in rd_region

eivindjahren:
Also removes unused private functions


andreas-el (approved):
Looks good to me! ✈️

PR #971, opened 2024-04-23
Remove broken replace_kw from ResdataFile

eivindjahren:
This function referenced the field data_owner from
ResdataKW which is never set, except after the lookup
and then it is set to False.

The only possible outcome of calling the function is
AttributeError so we are only removing undefined behavior
in terms of semver.

Also removes deprecated functions:
These functions have thrown NotImplementedError for a long
time so it should now be considered removing undefined
behavior in terms of semver.


yngve-sk (approved):
Very cool💯

PR #974, opened 2024-04-29
Refactor README

pinkwah:


CMakeLists.txt
```diff
@@ -35,47 +44,50 @@ endif()
...
-  file(
-    DOWNLOAD
-    "https://raw.githubusercontent.com/conan-io/cmake-conan/0.18.1/conan.cmake"
-    "${CMAKE_BINARY_DIR}/conan.cmake")
-endif()
+if(USE_CONAN)
+  if(NOT EXISTS "${CMAKE_BINARY_DIR}/conan.cmake")
+    message(
+      STATUS
+        "Downloading conan.cmake from https://github.com/conan-io/cmake-conan")
```
eivindjahren, line 51: Seems like there is both a change to cmake and a change to readme in the same commit titled "Refactor Readme". Would be nice to change the commit message to "Refactored cmake build".

eivindjahren (approved):
One simple comment, otherwise looks good :)

PR #975, opened 2024-04-29
Remove Equinor tests

pinkwah:
These tests aren't being run anymore and are taking up valuable brain space.


berland:
Is the code tested elsewhere?

If not, should we fix the CI so these tests are run?

eivindjahren:
I am closing this as we are actually relying on these tests to check that everything is working.

PR #977, opened 2024-05-14
Remove broken arg_pack.py

eivindjahren:
It refers to c functions `arg_pack_*` which do not exist. It probably existed at some point as it is refered to here: 
https://github.com/equinor/resdata/blob/b3cf05c890c941156baa9b91357de55e4af88b7d/lib/util/dependencies#L10


andreas-el (approved):
Keeping it tidy! 👍

PR #979, opened 2024-05-28
Fix segfault in WellStateglobalConnections

eivindjahren:
Fixes a bug where None would be given to `WellState._segment_collection_iget` would return None, which would become a nullptr and dereferenced in `self._segment_collection_iget` when called from `WellState.globalConnections`.


andreas-el (approved):
💯

PR #981, opened 2024-05-29
Remove nonsensical time_interval.hpp

eivindjahren:
The header just includes itself which just blows up whenever used.


dafeda (approved):

yngve-sk (approved):
↻↻↻↻↻

PR #991, opened 2024-08-07
Fix an issue where util_sscanf_int("") parsed as 0

### LINKED ISSUE #959: util_sscanf_int treats empty string as 0 without errors
`util_sscanf_int("", &value)` returns ok, with value set to 0. This is has the effect that

```python
from resdata import ResdataUtil

def test_failure():
    data_contents = "PARALLEL '' /\n"
    with open("TEST.DATA", "w") as f:
        f.write(data_contents)
    print(ResdataUtil.get_num_cpu("TEST.DATA"))
```
prints `0` rather than `1`.

eivindjahren:
**Issue**
Resolves #959 

the function is used in
* rd_get_start_date
* rd_get_num_parallel_cpu__
* string_util_sscanf_alloc_active_list
* rd_version_is_devel_version

Neither of which where interpreting "" as 0 is appropriate.

It is also used in `stringlist_iget_as_int`, but that function is getting removed as it is not used anywhere.

### DIFF
--- a/lib/util/tests/ert_util_sscan_test.cpp
+++ b/lib/util/tests/ert_util_sscan_test.cpp
@@ -174,6 +174,12 @@ void test_sscanf_int() {
     test_assert_false(util_sscanf_int("7.5", &value));
     test_assert_int_equal(value, 1);
 
+    test_assert_false(util_sscanf_int("abc1", &value));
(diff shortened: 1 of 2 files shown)


andreas-el (approved):
Nice one!
We could have tested for multiple spaces and such, but that might be too far.

Issue #954, opened 2024-02-16
Make c/c++ interface private


sondreso:
We will make the `c`/`c++` interface of resdata (formerly known as ecl/libecl) private.

The rationale is that there are a lot of functionality that is only exposed in c/c++ which is in little use, and we would like to clean this up in order to have a library that is easier to maintain. [ert](https://github.com/equinor/ert) used to be a big consumer of this interface, but this is not the case anymore. Other users 

The next step is ta create a new major version bump of Resdata with the `c`/`c++` interface as private. This means all comments in README and documentation about the c/c++ interface will disappear. Also the following lines will be removed: 
https://github.com/equinor/resdata/blob/c62737ab1bcae238c36346e526d8742772c3e7e1/python/resdata/__init__.py#L69

https://github.com/equinor/resdata/blob/c62737ab1bcae238c36346e526d8742772c3e7e1/setup.py#L86

Note: The binaries exposed by resdata is still considered public:
https://github.com/equinor/resdata/blob/c62737ab1bcae238c36346e526d8742772c3e7e1/setup.py#L37-L53

magnesj:
Hi, due to historical reasons, we have a copy of the `resdata` repository in `ResInsight` using the C++ interface only, not the Python interface. We are asked to evaluate using the official `resdata` repository instead of a local copy. For `ResInsight` it is required to be able to access and use the C++ code in `resdata`  using `CMake.` Please make sure that using C++ still will be possible.

Here is our copy (4-5 years since last sync with the public repository)
https://github.com/OPM/ResInsight/tree/dev/ThirdParty/Ert

eivindjahren:
>Here is our copy (4-5 years since last sync with the public repository)
>https://github.com/OPM/ResInsight/tree/dev/ThirdParty/Ert

@magnesj seeing as your copy and resdata has diverged quite a bit, would it not be simpler to keep these projects separate? You can still apply patches from resdata without using the cmake file, but given how much divergence there is, that might very rarely be possible.

Also, the goal of resdata is to maintain the backwards compatability of the python interface, where historical artifacts like `util_abort` become difficult to deal with. Freed from the need to maintain the python interface, resinsight could deal with that in a standard c++ way: smart pointers, RAII, and exceptions.

PR #994, opened 2024-08-07
Remove deprecated .h headers

eivindjahren:
Another step in the direction of #954 

Removes all cases of there being both `.h` and `.hpp` files.


sondreso (approved):
:broom:

PR #949, opened 2024-01-26
Remove unused c code

eivindjahren:
Removes unused c code now that only the python api and executables are considered public interface:

The current files that have been searched for unused c code
- [x] lib/util/buffer.cpp
- [x] lib/util/hash.cpp
- [x] lib/util/path_stack.cpp
- [x] lib/util/parser.cpp
- [x] lib/util/statistics.cpp
- [x] lib/util/cxx_string_util.cpp
- [x] lib/util/util.cpp
- [x] lib/util/util_opendir.cpp
- [x] lib/util/lookup_table.cpp
- [x] lib/util/util_unlink.cpp
- [x] lib/util/node_ctype.cpp
- [x] lib/util/string_util.cpp
- [x] lib/util/path.cpp
- [x] lib/util/util_lockf.cpp
- [x] lib/util/perm_vector.cpp
- [x] lib/util/util_symlink.cpp
- [x] lib/util/hash_sll.cpp
- [x] lib/util/util_abort.cpp
- [x] lib/util/util_zlib.cpp
- [x] lib/util/util_endian.cpp
- [x] lib/util/node_data.cpp
- [x] lib/util/test_work_area.cpp
- [x] lib/util/util_spawn.cpp
- [x] lib/util/mzran.cpp
- [x] lib/util/util_getuid.cpp
- [x] lib/util/test_util.cpp
- [x] lib/util/rd_version.cpp
- [x] lib/util/rng.cpp
- [x] lib/util/stringlist.cpp
- [x] lib/util/vector.cpp
- [x] lib/util/hash_node.cpp
- [x] lib/util/type_vector_functions.cpp
- [x] lib/resdata/rd_util.cpp
- [x] lib/resdata/rd_io_config.cpp
- [x] lib/resdata/rd_sum_tstep.cpp
- [x] lib/resdata/grid_dims.cpp
- [x] lib/resdata/rd_kw_functions.cpp
- [x] lib/resdata/rd_init_file.cpp
- [x] lib/resdata/well_state.cpp
- [x] lib/resdata/well_info.cpp
- [x] lib/resdata/smspec_node.cpp
- [x] lib/resdata/rd_rft_file.cpp
- [x] lib/resdata/rd_grav_common.cpp
- [x] lib/resdata/rd_rsthead.cpp
- [x] lib/resdata/rd_region.cpp
- [x] lib/resdata/rd_rft_cell.cpp
- [x] lib/resdata/rd_file_kw.cpp
- [x] lib/resdata/nnc_info.cpp
[...]

PR #997, opened 2024-08-08
Remove unused node_data c functions

eivindjahren:
In relation to https://github.com/equinor/resdata/issues/954 we will remove c code that is not used to implement python functionality.

This PR is does only part of that and you can see the overview here: https://github.com/equinor/resdata/pull/949


sondreso (approved):
Needs rebasing after https://github.com/equinor/resdata/pull/995#pullrequestreview-2226904807 is merged.

PR #1000, opened 2024-08-12
Fix issue with from_pandas

### LINKED ISSUE #944: Bug in `_compile_headers_list`
In `_compile_headers_list` there is a check if the first character of wgnames is numeric or not. If it's numeric then it's assumed to be either cell/block coordinate (i, j, k) or region/aquifer index.  

https://github.com/equinor/resdata/blob/6a113a4d42536e3f05cbe61a2ef4f10ff8ce6941/python/resdata/summary/rd_sum.py#L619

This will fail if for example the well name is starting with number.

Suggestion, add more check to the keyword, for example, ensure they are not well vectors (WOPR, WWCT, etc) and not a segment vectors (SOFR, SWFR etc)
````python
if lst[1][0].isdigit() and kw[0] not in ['W', 'S']:
    nums = re.split(",", lst[1])
else:
    wgname = lst[1]
````

eivindjahren:
Resolves #944

The PR duplicates the generator code from the ert test code. I think that is preferable to trying to reuse this code in both places as test failures will be difficult to debug, and we would likely want to modify the generators so that `summaries()` generates a `resdata.summary.Summary`

This breaks one test case from https://github.com/equinor/resdata/commit/1a41876ce2363a718e769a2ac896a212d4cb848a , but it seems like this synthetic data does not reflect how these fields should be interpreted.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -107,7 +107,7 @@ jobs:
     - name: Run Python tests
       run: |
         # Runs tests on installed distribution from an empty directory
-        python -m pip install pytest
+        python -m pip install -r test_requirements.txt
 
         # pytest adds every directory up-to and including python/ into sys.path,
         # meaning that "import resdata" will import python/resdata and not the installed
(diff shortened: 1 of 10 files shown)


sondreso (approved):
I think this looks good, but could you split it into one commit for adding the test data generator and one for solving the issue? :slightly_smiling_face: 

:100: :zero:

eivindjahren:
> I think this looks good, but could you split it into one commit for adding the test data generator and one for solving the issue? 🙂

Split it now

sondreso:
> > I think this looks good, but could you split it into one commit for adding the test data generator and one for solving the issue? 🙂
> 
> Split it now

:rocket: Go for merge

PR #1001, opened 2024-08-13
Remove more unused c code

eivindjahren:
In relation to https://github.com/equinor/resdata/issues/954 we will remove c code that is not used to implement python functionality.

This PR is does only part of that and you can see the overview here: https://github.com/equinor/resdata/pull/949


andreas-el (approved):
Deserve a medal sir! 🎖️

PR #1017, opened 2024-08-30
Remove test using removed rd_sum_report_step_equal

eivindjahren:
In d1c2b93b1589faad378e71ad83f7927ab8bc24b4 , the test rd_sum_report_step_equal was removed. However, some corresponding ctest commands were not removed.


berland (approved):
✂️

PR #1020, opened 2024-09-02
Fix flaky test_to_from_pandas

eivindjahren:
Fix flaky test https://github.com/equinor/resdata/actions/runs/10628849221/job/29464486920

### DIFF
--- a/python/tests/rd_tests/test_rd_sum.py
+++ b/python/tests/rd_tests/test_rd_sum.py
@@ -177,9 +177,12 @@ def test_to_from_pandas(summary):
         "TEST", baseline, dims=(smspec.nx, smspec.ny, smspec.nz)
     ).pandas_frame()
 
-    # round to hours
-    roundtrip.index = roundtrip.index.round(freq="H")
-    baseline.index = baseline.index.round(freq="H")
+    # Round to days because of precision loss due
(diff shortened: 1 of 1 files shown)


python/tests/rd_tests/test_rd_sum.py
```diff
@@ -177,9 +177,9 @@ def test_to_from_pandas(summary):
         "TEST", baseline, dims=(smspec.nx, smspec.ny, smspec.nz)
     ).pandas_frame()
 
-    # round to hours
-    roundtrip.index = roundtrip.index.round(freq="H")
-    baseline.index = baseline.index.round(freq="H")
+    # round to days
```
berland, line 180: Could we leave an explanation here for why we round (so much). Related to some int32 handling of dates?

eivindjahren: Sure, its because there is so much conversion.

eivindjahren: Updated

berland: not pushed yet (?)

eivindjahren: Forgot to push, is updated now.

berland (approved):
Great!

PR #1021, opened 2024-09-02
Fix pytest.mark typo and associated tmpdir-bug

berland:
* Fix bug that only occurs when the fixture is actually activated.
* Enforce strict markers


eivindjahren (approved):
:+1:

PR #1023, opened 2024-09-03
Fix missing import for equinor-testdata

berland:
**Issue**
Resolves failure when running with Equinor testdata present. Tested locally confirming it should work with Equinor testdata.

**Approach**
Do the import.


eivindjahren (approved):
:+1:

PR #1032, opened 2025-01-06
Fix bug with rporv

### LINKED ISSUE #962: rd_grav_survey_assert_RPORV has infinite loop if no init_porv > 0
The problem is that it will try to check 100 random non-zero points, but there may be fewer than 100 non-zero points in which case it is just an infinite loop: https://github.com/equinor/resdata/blob/c76fecd7e9e909c509e94547cb37206901d83cfb/lib/resdata/rd_grav.cpp#L376

eivindjahren:
**Issue**
Resolves #962

### DIFF
--- a/lib/include/resdata/rd_kw_magic.hpp
+++ b/lib/include/resdata/rd_kw_magic.hpp
@@ -17,7 +17,6 @@ extern "C" {
 /*****************************************************************/
 
 #define PORO_KW "PORO"
-#define PORV_KW "PORV"
 #define AQUIFER_KW "AQUIFERN"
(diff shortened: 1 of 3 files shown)


berland (changes requested):
This is no duplicate 👀

lib/include/resdata/rd_kw_magic.hpp
```diff
@@ -17,7 +17,6 @@ extern "C" {
 /*****************************************************************/
 
 #define PORO_KW "PORO"
-#define PORV_KW "PORV"
```
berland, line 20: This is no duplicate..

Issue #1035, opened 2025-02-14
Deprecation and/or removal of resdata applications


berland:
The applications in `resdata/applications/resdata` installed through `setup.py` serves various purposes. Some of these should be deprecated and/or possibly removed:

* `CF_dump`: Any user of this is better of using `res2df` and its `rft` module, that extracts the same information and more
* `convert.x`: This might have some use cases, but could be deprecated in favour of a recommentation to use `resfo` instead.
* `rd_pack.x`/`rd_unpack.x`: This is used by `subscript`'s `restartthinner`. If subscript uses resfo instead, we can deprecate it.
* `grdecl_grid`: Should be deprecated or outright removed. It parses input decks, but very poorly. And then it does nothing.
* `grdecl_test`: Times the loading of a grdecl-file. Limited use for this, deprecate.
* `grid_dump.x`/`grid_dump_ascii.x`: Prints xyz for all 8 corners of every cornerpoint-cell. Has probably limited use-cases. Deprecate.
* `grid_info.x`: Will print some simple aggregated numbers for a grid. Attainable in short lines of python from res2df also. Can also compare two grids. for equivalence. 
* [more to come]

PR #1036, opened 2025-02-14
Deprecate applications with replacements or with little use

berland:
**Issue**
Partially solves #1035 

**Approach**
`fprintf`


eivindjahren (commented):
This seems like what we discussed, I think we just want a final say from @sondreso

berland:
I have compiled all binaries and verified that they emit the warning and do not crash (more than before).

PR #1039, opened 2025-04-02
Restrict cmake to < 4

JHolba:
restricting cmake to <4 due to errors caused by the upgrade

some packages built by conan give errors due to them not setting CMake minimum version to >= 3.5. 
> Compatibility with CMake < 3.5 has been removed from CMake.


berland (approved):
yay

PR #1045, opened 2025-05-14
Resolve Python 3.12 SyntaxWarning

berland:
**Issue**
Resolves Python 3.12 SyntaxWarning when building bytecode files


andreas-el (approved):
🎖️

PR #1088, opened 2025-11-17
Type cast cell properties for backwards compatibility

HakonSohoel:


HakonSohoel:
planing to upgrade resfo-utilities version before rebase and merge

eivindjahren (approved):

PR #1108, opened 2026-01-20
Use Conan2

JHolba:
Update conan dependency to conan 2.

Use cmake presets made by conan for configuring and building resdata


eivindjahren (approved):
Looks good! A bit disappointing that you have to do so much mucking about in setup.py to get conan2 working, but I guess that is just the way it is.

JHolba:
I agree. Could be done easier in scikit-build-core with scikit-build-core conan plugin, but that plugin is a community plugin with version < 1.0

PR #1110, opened 2026-01-20
Simplify github actions

JHolba:
Use cibuildwheel to build all wheels.
Python tests are now done by cibuildwheel.
Merge windows building into the same action as other OS.


.github/workflows/testing.yml
```diff
@@ -5,7 +5,7 @@ on:
    branches:
      - main
      - 'version-**'
-   tags: "*"
+   tags: ["*"]
```
eivindjahren: What is the effect of this change?

PR #1119, opened 2026-02-12
Add tests for layer.cpp

eivindjahren:


lib/tests/test_layer.cpp
```diff
@@ -0,0 +1,663 @@
...
+                }
+
+                THEN("Tracing from a cell with non-matching value fails") {
+                    REQUIRE_FALSE(layer_trace_block_content(layer, false, 6, 6,
+                                                            0, i_list, j_list));
+                    REQUIRE_FALSE(layer_trace_block_content(
+                        layer, false, 5, 5, 99, i_list, j_list));
+                }
+                int_vector_free(i_list);
+                int_vector_free(j_list);
```
ajaust: Could be removed if we use smart pointer.

lib/tests/test_layer.cpp
```diff
@@ -0,0 +1,663 @@
...
+                        for (int j = 0; j < ny; j++) {
+                            bool grid_active =
+                                rd_grid_cell_active3(grid, i, j, 0);
+                            bool layer_active = layer_iget_active(layer, i, j);
+                            REQUIRE(layer_active == grid_active);
+                        }
+                    }
+                }
+            }
+            rd_grid_free(grid);
```
ajaust: Could be removed if we use a smart pointer.

lib/tests/test_layer.cpp
```diff
@@ -0,0 +1,663 @@
...
+            layer_iset_cell_value(layer, 0, 0, 5);
+
+            THEN("The cell sum is the sum of the set values") {
+                REQUIRE(layer_get_cell_sum(layer) == 20 + 5);
+            }
+        }
+
+        THEN("All cells are active by default") {
+            for (int i = 0; i < 5; i++) {
+                for (int j = 0; j < 5; j++) {
```
ajaust: Why are the upper bounds `5` and not `nx` nor `ny`? The same pattern appears several times below.

ajaust (approved):
I only have minor things and the pointers you have already fixed.  You have to see if you want to change anything based on my comments. 🙂

PR #1121, opened 2026-02-13
Fix util clear directory

eivindjahren:
This should resolve https://github.com/equinor/resdata/security/code-scanning/6

Currently blocked by the need to deprecate `resdata.util.test`. The plan is to do this in version 7.


lib/util/test_work_area.cpp
```diff
@@ -187,8 +187,13 @@ TestArea::TestArea(const std::string &test_name, bool store_area)
 }
 
 TestArea::~TestArea() {
-    if (!this->store)
-        util_clear_directory(this->cwd.c_str(), true, true);
+    if (!this->store) {
+        try {
+            std::filesystem::remove_all(this->cwd);
+        } catch (std::filesystem::filesystem_error const &ex) {
```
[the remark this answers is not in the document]
eivindjahren: According to https://man7.org/linux/man-pages/man3/rmdir.3p.html this is unspecified, but could result in EBUSY. I changed this so that orig_cwd is restored first to avoid the issue.

lib/util/test_work_area.cpp
```diff
@@ -187,8 +187,11 @@ TestArea::TestArea(const std::string &test_name, bool store_area)
 }
 
 TestArea::~TestArea() {
-    if (!this->store)
-        util_clear_directory(this->cwd.c_str(), true, true);
+    if (!this->store) {
+        std::error_code ec;
+        std::filesystem::remove_all(this->cwd, ec);
```
ajaust, line 193: - The old implementation was called with `strict_uid=true`. This would, as far as I can tell, only delete files owned by the user running the code. `std::filesystem::remove_all` will remove all files that a user has access to even if they are not the owner. Is this change in behavior intended?
- Apparently, this standard library implementation of `std::filesystem::remove_all` had the same vulnerability as resdata in the past (bug report for [LLVM](https://reviews.llvm.org/D118134), and [GCC](https://gcc.gnu.org/pipermail/gcc-cvs/2023-October/390561.html)). Do we need to verify that the compilers we use for building resdata have the fix? The issue is fixed in newer compilers and also has been backported.

eivindjahren: Ah, I missed that. That complicates this quite a bit so I think we will have to reconsider.

ajaust: I don't really know in which context the deletion is triggered. If this is only used in some kind of test cleanup it might be fine using the `remove_all`.

eivindjahren: Its probably fine since it is indeed test clean up. It is a really minor change in behavior and we could claim that changing the ownership inside the directory was never defined before or even that the test interface was never part of the public interface. I am leaning towards that we should also remove anything test related in the interface.

ajaust: Sounds good to me.

ajaust (approved):

PR #1131, opened 2026-03-03
Refactor fault block layer.cpp

eivindjahren:
Note that in particular this changes `util_abort` to `.at()` out-of-range exceptions in some cases.

### DIFF
--- a/lib/resdata/fault_block_layer.cpp
+++ b/lib/resdata/fault_block_layer.cpp
@@ -59,22 +59,20 @@ fault_block_type *fault_block_layer_add_block(fault_block_layer_type *layer,
 
 void fault_block_layer_scan_layer(fault_block_layer_type *fault_layer,
                                   layer_type *layer) {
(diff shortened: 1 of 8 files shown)


lib/resdata/fault_block_layer.cpp
```diff
@@ -111,26 +111,27 @@ bool fault_block_layer_scan_kw(fault_block_layer_type *layer,
         return false;
     else {
         int max_block_id = 0;
-        layer_type *work_layer = layer_alloc(rd_grid_get_nx(layer->grid),
-                                             rd_grid_get_ny(layer->grid));
+        std::unique_ptr<layer_type, decltype(&layer_free)> work_layer(
```
ajaust: You should `#include <memory>` to be on the safe side that `std::unique_ptr` is available in the file.

ajaust (approved):
Looks good to me. I only have minor points.

Issue #1117, opened 2026-02-11
rd_grid_alloc_mapaxes_kw returns inconsistent rd_kw when mapaxes is not used


eivindjahren:
Calling rd_grid_alloc_mapaxes_kw will always return a rd_kw_type pointer with size 6, but grid->mapaxes may be null, meaning the data is instantiated to NULL so doing

```c
for (int i = 0; i < rd_kw_get_size(mapaxes_kw);i++)
   printf("%s\n", rd_kw_iget_float(mapaxes_kw, i));
```

will segfault.

There is a warning printed:

```
.../resdata/lib/resdata/rd_kw.cpp:726:11: runtime error: null pointer passed as argument 2, which is declared to never be 
null
```

PR #1134, opened 2026-04-05
Handle missing MAPAXES export when mapaxes is absent

yasser508:
Make `rd_grid_alloc_mapaxes_kw()` handle grids without mapaxes.

I added a regression test for the case where `rd_grid_alloc_mapaxes_kw()` is called for a grid with no `MAPAXES` data, which was the issue in #1117.

Then I changed the implementation so it returns an empty keyword instead of a size-6 keyword with null backing data.


ajaust:
Thank you for the proposed changes and sorry for the late reply. We have implemented some fixes and rewrote some parts which should also handle absent mapaxes, see PR 1196. I will close this PR for now, but feel free to reopen if you think that it is still relevant.

PR #1137, opened 2026-04-16
Fix util_abort called with too few arguments

eivindjahren:
**Issue**
Resolves https://github.com/equinor/resdata/security/code-scanning/9 and two similar security reports


ajaust (approved):
Looks good.

Should we `-Wformat` to the default compile options? This might be able to detect such errors at compile time.

eivindjahren:
@ajaust we have -Wall (which gives `-Wformat=1`), but not -WError and there are a number of warnings. We should probably get that fixed also.

PR #1138, opened 2026-04-20
Deprecate resdata.util.test and set to remove in version 7.

eivindjahren:
In order to fix TestArea, it would be much simpler to apply some breaking changes: https://github.com/equinor/resdata/pull/1121 . We therefore deprecate `resdata.util.tests` which no one should really use, as all of its functionality is also in pytest/unittest, except potentially for `SummaryMock` which is not that great.

Note that since pytest errors on deprecation warnings at collect time, it was necessary to copy in the parts of `resdata.util.test` that was used by the tests straight away.


oyvindeide (approved):
LGTM!

PR #1142, opened 2026-04-24
Ensure that scanf with %s format specifier is always given width

eivindjahren:


lib/resdata/rd_kw_grdecl.cpp
```diff
@@ -133,8 +134,8 @@ static bool rd_kw_grdecl_fseek_kw__(const char *kw, FILE *stream) {
     long init_pos = util_ftell(stream);
     while (true) {
         if (rd_kw_grdecl_fseek_next_kw(stream)) {
-            char next_kw[256];
-            fscanf(stream, "%s", next_kw);
+            char next_kw[256] = {0};
+            fscanf(stream, "%255s", next_kw);
```
ajaust, line 138: I am no GRDECL expert, but is it guaranteed that a keyword is never longer than 255 chars? Most of the other input seems to allow 511 chars.

lib/resdata/rd_kw_grdecl.cpp
ajaust, line 401: Not introduced in this PR, but there might be some consistency issue. `strlen` doesn't seem to count `\0`. If the buffer has 512 chars the current conditional evaluates to `false` (512 > 512) and therefore does not abort. I think in this case we would truncate a char since the buffer will only hold 511 chars (plus `\0`, I guess).

```suggestion
    if (header && strlen(header) >= MAX_GRDECL_HEADER_SIZE)
```

lib/resdata/rd_kw_grdecl.cpp
ajaust, line 138: Not introduced in this PR, but a thought:

Should we check the return value of `fscanf` to check if we reach the end of the file or similar? Could we get stuck in an endless loop if we do not check for read failures?

ajaust (approved):
I think the fix looks good. I have some comments regarding stuff not introduced in this PR, but that is touching similar stuff (reading chars).

andreas-el (approved):
💯

lib/resdata/rd_kw_grdecl.cpp
eivindjahren: It should not exceed 8 characters actually, I just kept the size the same as to not change anything too much (yet).

eivindjahren: Yes, this is a bit of a headache as fscanf does not progress if you get a read failure, so then you get an infinite loop instead. If you have any suggestions to good fixes then I am all ears.

ajaust: A wild guess since I do not have not read and understood the whole code. Could this work?

```suggestion
            if (fscanf(stream, "%255s", next_kw) != 1) {
                util_fseek(stream, init_pos, SEEK_SET);
                return false;
            }
```
So if we do not find and assign exactly one field (for whatever reason), we reset the position the same way as if `if rd_kw_grdecl_fseek_next_kw(stream))` would have returned false.

PR #1144, opened 2026-04-24
Add more tests for grid

eivindjahren:
Very large PR in aggreggate, but that is due to things changing place and many "one word" changes *in tests* from e.g. `rd_grid_xxx(grid)` to `rd_grid_xxx(grid.get())` and similar. There is also a lot of moved code so I have split that into several commits. The actual count of new tests is 2364 lines, the remainder is some changes away from manual memory management in tests and reorganization of tests.


lib/include/resdata/rd_grid.hpp
```diff
@@ -208,5 +208,15 @@ UTIL_SAFE_CAST_HEADER(rd_grid);
 
 #ifdef __cplusplus
 }
+
+#include <memory>
+#include <filesystem>
+
+namespace fs = std::filesystem;
```
ajaust: If you know that this is save, this is fine. I think it is oftend advised against defining namespace shortcuts (or `using namespace`) in headers. As header my be included in other projects/files, it may lead to naming conflicts.

lib/tests/grid_fixtures.hpp
```diff
@@ -49,24 +53,18 @@ inline rd_grid_type *generate_coordkw_grid(
...
     }
 
-    rd_grid_type *grid = rd_grid_alloc_GRDECL_kw(num_x, num_y, num_z, zcorn_kw,
-                                                 coord_kw, NULL, NULL);
-    rd_kw_free(coord_kw);
-    rd_kw_free(zcorn_kw);
-
-    return grid;
+    return build_grdecl_grid(num_x, num_y, num_z, zcorn_kw.get(),
+                             coord_kw.get());
```
ajaust: Compilation fails from commit `Avoid manual memory management in generate_coordkw_grid` due to usage `build_grdecl_grid` which is not defined yet. It will be added in a later commit.

Found it by running:
```bash
git rebase -i upstream/main --exec "cmake --build --preset=conan-release -j 8 && ctest --preset conan-release"
```

lib/tests/test_fault_block_layer.cpp
```diff
@@ -8,8 +8,7 @@
 
 #include <catch2/catch.hpp>
 
-static std::unique_ptr<rd_grid_type, decltype(&rd_grid_free)>
-make_grid(int nx, int ny, int nz) {
+static rd_grid_ptr make_grid(int nx, int ny, int nz) {
     return {rd_grid_alloc_rectangular(nx, ny, nz, 1, 1, 1, nullptr),
             rd_grid_free};
```
ajaust: Consider changing this to

```suggestion
            &rd_grid_free};
```

for consistency.

lib/resdata/rd_grid.cpp
```diff
@@ -5663,3 +5663,9 @@ void export_corners(const rd_grid_type *grid, int index_size,
         }
     }
 }
+
+rd_grid_ptr make_rectangular_grid(int nx, int ny, int nz, double dx, double dy,
+                                  double dz, const int *actnum) {
+    return {rd_grid_alloc_rectangular(nx, ny, nz, dx, dy, dz, actnum),
+            rd_grid_free};
```
ajaust: Consider changing this to

```suggestion
            &rd_grid_free};
```

for consistency.

lib/tests/test_rd_grid_regular.cpp
ajaust, line 1: Is the term `rectangular` equivalent to `regular`? I see that both are used and wonder if there is any difference and whether it should affect the name renamed file.

lib/tests/grid_fixtures.hpp
```diff
@@ -49,24 +53,18 @@ inline rd_grid_type *generate_coordkw_grid(
...
+        rd_kw_iset_float(zcorn_kw.get(), index, z);
     }
 
-    rd_grid_type *grid = rd_grid_alloc_GRDECL_kw(num_x, num_y, num_z, zcorn_kw,
-                                                 coord_kw, NULL, NULL);
-    rd_kw_free(coord_kw);
-    rd_kw_free(zcorn_kw);
-
-    return grid;
+    return build_grdecl_grid(num_x, num_y, num_z, zcorn_kw.get(),
```
ajaust: From this commit on, the code does not compile due to the usage of `build_grdecl_grid` which is not yet defined. I think it will be defined two commits later.

This should be reproducible by

```bash
git rebase -i upstream/main --exec "cmake --build --preset=conan-release -j 8 && ctest --preset conan-release"
```

lib/include/resdata/rd_grid.hpp
```diff
@@ -210,10 +210,13 @@ UTIL_SAFE_CAST_HEADER(rd_grid);
...
 #include <memory>
+#include <filesystem>
 
-using rd_grid_ptr = std::unique_ptr<rd_grid_type, decltype(&rd_grid_free)>;
+namespace fs = std::filesystem;
 
+using rd_grid_ptr = std::unique_ptr<rd_grid_type, decltype(&rd_grid_free)>;
 rd_grid_ptr make_rectangular_grid(int nx, int ny, int nz, double dx, double dy,
                                   double dz, const int *actnum);
+rd_grid_ptr read_grid(fs::path filename);
```
ajaust: I agree with copilot that this could be

```suggestion
rd_grid_ptr read_grid(const std::filesystem::path &filename);
```

if you want to avoid a copy.

lib/tests/grid_fixtures.hpp
```diff
@@ -12,6 +16,147 @@
...
+                         const int *values, int size) {
+    auto kw = make_rd_kw(name, size, RD_INT);
+    for (int i = 0; i < size; i++)
+        rd_kw_iset_int(kw.get(), i, values[i]);
+    rd_kw_fwrite(kw.get(), fortio);
+}
+
+inline void write_int_kw(fortio_type *fortio, const char *name,
+                         const std::vector<int> &values) {
+    write_int_kw(fortio, name, values.data(), (int)values.size());
```
ajaust: This file contains some C style casts that could be C++ `static_cast`, but that is an optional change.

```suggestion
    write_int_kw(fortio, name, values.data(), static_cast<int>(values.size()));
```

lib/tests/grid_fixtures.hpp
```diff
@@ -12,6 +16,147 @@
...
+
+inline void write_grid_gridunit(const ERT::FortIO &fortio) {
+    write_char8_kw(fortio.get(), GRIDUNIT_KW, {"METRES", ""});
+}
+
+inline void write_grid_cell(const ERT::FortIO &fortio, int i, int j, int k,
+                            int global_index, int host_cell,
+                            const float corners[24]) {
+    const int coords[7] = {i + 1, j + 1,     k + 1, global_index + 1,
+                           1,     host_cell, 0};
```
ajaust, line 162: This looks like it has weird spacing that seems to be intentional. I am not sure why. This is not a GitHub rendering issue since I see the same in Vim.

lib/tests/grid_fixtures.hpp
```diff
@@ -68,3 +219,382 @@ inline rd_grid_ptr generate_coordkw_grid(
     return build_grdecl_grid(num_x, num_y, num_z, zcorn_kw.get(),
                              coord_kw.get());
 }
+
+/**
+ * Writes an EGRID file with a CORSNUM keyword describing coarse
+ * cell groups. The grid is an nx*ny*nz rectangular grid with unit cells.
+ *
+ *  corsnum must have nx*ny*nz entries. A value of 0 means the cell is
```
ajaust:
```suggestion
 * corsnum must have nx*ny*nz entries. A value of 0 means the cell is
```

lib/tests/grid_fixtures.hpp
```diff
@@ -68,3 +219,382 @@ inline rd_grid_ptr generate_coordkw_grid(
...
+}
+
+/**
+ * Writes a minimal single-porosity .GRID file containing a 1x1x1 main grid
+ * and a 1x1x1 LGR covering the single main-grid cell. The LGR_KW is written
+ * with two elements: the LGR name followed by parent_name.
+ *
+ * A parent_name of "" or "GLOBAL" is treated as "no parent".
+ * Any other non-empty value must match an existing grid name (nested LGRs).
+ */
```
ajaust: Is this a stray comment without an existing method? I don't see any method underneath the comment.

There are similar comments in line 470. Maybe the comment here belongs somewhere in that region?

lib/tests/grid_fixtures.hpp
ajaust, line 1: Some minor comments. The fixtures look meaningful, but I don't have enough knowledge of the library to say much more. If you need an in-depth analysis, let me know.

lib/tests/test_rd_grid_compare.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+    GIVEN("Two grids differing in NNC information") {
+        auto g1 = make_rectangular_grid(2, 2, 2, 1, 1, 1, nullptr);
+        auto g2 = make_rectangular_grid(2, 2, 2, 1, 1, 1, nullptr);
+        rd_grid_add_self_nnc(g1.get(), 0, 1, 0);
+
+        THEN("The grids are unequal") { REQUIRE_FALSE(grid_equal(g1, g2)); }
+        THEN("The grids are equal with include_nnc=false") {
+            REQUIRE(rd_grid_compare(g1.get(), g2.get(), false, false, true));
+        }
```
ajaust, line 70: It says "The grids are equal with include_nnc=false", but `include_lgr` is also set to false.

lib/tests/test_rd_grid_compare.cpp
```diff
@@ -0,0 +1,169 @@
...
+        auto g2 = load_egrid_with_coarse_groups(file_one, nx, ny, nz,
+                                                corsnum1.data());
+
+        THEN("The grids are unequal") { REQUIRE_FALSE(grid_equal(g1, g2)); }
+
+        THEN("Each grid is equal to itself") {
+            REQUIRE(grid_equal(g1, g1));
+            REQUIRE(grid_equal(g2, g2));
+        }
+    }
```
ajaust, line 132: Why is equality to itself tested here and in later tests, but not in the previous test?

lib/tests/test_rd_grid_regular.cpp
```diff
@@ -1,212 +1,220 @@
...
+
+    for (int j = 0; j < ny; j++) {
+        for (int i = 0; i < nx; i++) {
+            int offset = 6 * (i + j * nx);
+            rd_kw_iset_float(coord_kw.get(), offset, i);
+            rd_kw_iset_float(coord_kw.get(), offset + 1, j);
+            rd_kw_iset_float(coord_kw.get(), offset + 2, -1);
+            rd_kw_iset_float(coord_kw.get(), offset + 3, i);
+            rd_kw_iset_float(coord_kw.get(), offset + 4, j);
+            rd_kw_iset_float(coord_kw.get(), offset + 5, -1);
```
ajaust: Didn't remove such a pattern in an earlier commit by using `set_pillar` from the grid fixture class? Should we reuse `set_pillar` here?

lib/tests/test_rd_grid_regular.cpp
```diff
@@ -693,72 +707,118 @@ TEST_CASE("Test utility functions on a regular grid", "[unittest]") {
...
+                         RD_METRIC_UNITS);
+    write_fegrid_minimal(fegrid_path);
+    auto read_fegrid = read_grid(fegrid_path);
+
+    auto load_and_compare = [](const fs::path &case_path,
+                               const rd_grid_type *expected) {
+        auto loaded =
+            rd_grid_ptr(rd_grid_load_case(case_path.c_str()), &rd_grid_free);
+        REQUIRE(loaded != nullptr);
+        REQUIRE(rd_grid_compare(loaded.get(), expected, false, false, false));
```
ajaust: Why are `lgr` and `nnc` excluded from the comparison?

ajaust (commented):
Changes look good overall. I like that it adds to many tests although I did not go into all their details.

I left some comments. Let me know what you think. 🙂

lib/tests/test_rd_grid_regular.cpp
eivindjahren: Its really the same, the documentation isn't really consistent.

lib/tests/grid_fixtures.hpp
eivindjahren: Thanks, my mistake, I will fix this.

lib/tests/test_rd_grid_regular.cpp
```diff
@@ -0,0 +1,824 @@
...
+
+#include "grid_fixtures.hpp"
+#include "tmpdir.hpp"
+
+using namespace Catch;
+using namespace Matchers;
+
+TEST_CASE("Test unfractured grids", "[unittest]") {
+    GIVEN("An unfractured grid") {
+        auto grid = make_rectangular_grid(21, 11, 12, 1, 2, 3, NULL);
```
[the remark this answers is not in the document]
eivindjahren: This is just kept as is from a previous test.

lib/include/resdata/rd_grid.hpp
eivindjahren: I fixed it :)

lib/tests/test_fault_block_layer.cpp
eivindjahren: Fixed

lib/resdata/rd_grid.cpp

lib/tests/test_rd_grid_compare.cpp
eivindjahren: Fixed

lib/tests/test_rd_grid_regular.cpp
eivindjahren: Fixed

lib/tests/test_rd_grid_regular.cpp
eivindjahren: Fixed

lib/tests/test_rd_grid_compare.cpp
eivindjahren: Added self comparison everywhere.

lib/tests/grid_fixtures.hpp
eivindjahren: This seems to work correctly now at least, but I never managed to repro the problem.

lib/include/resdata/rd_grid.hpp
eivindjahren: Fixed

lib/tests/grid_fixtures.hpp
eivindjahren: Fixed

lib/tests/grid_fixtures.hpp
eivindjahren: Yea I see it too, but I think this is clang-format doing this.

lib/tests/grid_fixtures.hpp
eivindjahren: fixed

lib/tests/grid_fixtures.hpp
eivindjahren: Oups, my bad.

lib/tests/grid_fixtures.hpp
ajaust: I still have the same issue.

This commit and the following commit do not work for me, i.e, the following commits:

- a27e8a7269af7030d9b485951a5c1a6c04fffb43 "Avoid manual memory management in generate_coordkw_grid"
- 78add53bc35d960912c3221594f1faac076cd3a4 "Avoid manual memory management in test_nnc_info"

Could it be that the compiler finds the implementation in another file on your machine? I don't expect that the compiler makes a difference here.

Maybe 
``
git rebase -i upstream/main --exec "pushd build && make clean && popd && cmake --build --preset=conan-release -j 8 && ctest --preset conan-release"
```
will reproduce it by cleaning up the build directory.

File: lib/tests/test_rd_grid_xyz_lookup.cpp [hunk 16]
```diff
@@ -0,0 +1,216 @@
...
+ * global index of the cell that contains (x, y, z), or -1 if no cell does.
+ * A cell "contains" a point when the point lies inside the volume of the
+ * cell's hexahedron, or on one of the cell faces that is shared with no
+ * neighbour (so points on the outer boundary of the grid still belong to a
+ * cell, while points on an internal face belong to exactly one of the two
+ * cells that share it).
+ *
+ * Some special cases:
+ * * Cells with degenerate geometry are treated as empty;
+ * * An cell with a corner pinned at the origin is considered invalid, so contains no points;
```
ajaust:
```suggestion
 * * A cell with a corner pinned at the origin is considered invalid, so contains no points;
```

lib/tests/test_rd_grid_xyz_lookup.cpp
```diff
@@ -0,0 +1,216 @@
...
+        REQUIRE(rd_grid_get_global_index_from_xyz(grid.get(), 1.3, 1.3, 0.5,
+                                                  0) == 0);
+        REQUIRE(rd_grid_get_global_index_from_xyz(grid.get(), 1.3, 1.3, 0.5,
+                                                  -1) == 0);
+    }
+}
+
+TEST_CASE("rd_grid_get_global_index_from_xyz on a flat single cell",
+          "[unittest]") {
+    // An cell whose corners all share the same z
```
ajaust:
```suggestion
    // A cell whose corners all share the same z
```

ajaust (commented):
The changes look good overall. Just minor typo I have seen. Building the 6th and 7th commit still fails on my machine, but we discussed this.

## Remarks not directly relevant to this PR
The compiler on Mac gives the following warnings which may be relevent w.r.t. security:

- `ld: warning: object file (/Users/AEJ/.conan2/p/b/fmt4335568f4a2e9/p/lib/libfmt.a[2](format.cc.o)) was built for newer 'macOS' version (26.0) than being linked (11.0)`
- `warning: 'vsprintf' is deprecated: This function is provided for compatibility reasons only. Due to security concerns inherent in the design of sprintf(3), it is highly recommended that you use vsnprintf(3) instead. [-Wdeprecated-declarations]`
- `warning: 'sprintf' is deprecated: This function is provided for compatibility reasons only. Due to security concerns inherent in the design of sprintf(3), it is highly recommended that you use snprintf(3) instead. [-Wdeprecated-declarations]`

lib/tests/grid_fixtures.hpp
eivindjahren: Should have fixed it now.

lib/tests/test_rd_grid_xyz_lookup.cpp
eivindjahren: Done

lib/tests/test_rd_grid_xyz_lookup.cpp
eivindjahren: Done

eivindjahren:
> The changes look good overall. Just minor typo I have seen. Building the 6th and 7th commit still fails on my machine, but we discussed this.
> 
> ## Remarks not directly relevant to this PR
> The compiler on Mac gives the following warnings which may be relevent w.r.t. security:
> 
> * `ld: warning: object file (/Users/AEJ/.conan2/p/b/fmt4335568f4a2e9/p/lib/libfmt.a[2](format.cc.o)) was built for newer 'macOS' version (26.0) than being linked (11.0)`
> * `warning: 'vsprintf' is deprecated: This function is provided for compatibility reasons only. Due to security concerns inherent in the design of sprintf(3), it is highly recommended that you use vsnprintf(3) instead. [-Wdeprecated-declarations]`
> * `warning: 'sprintf' is deprecated: This function is provided for compatibility reasons only. Due to security concerns inherent in the design of sprintf(3), it is highly recommended that you use snprintf(3) instead. [-Wdeprecated-declarations]`

Should probably fix that in a separate commit then.

ajaust (approved):
Looks good. 

One interesting observation that I made when running all tests: The total runtime of the C++ test suite stays roughly the same even though many tests were added. It was ~22s before the PR on my machine and is around ~23s at the last commit.

It seems that
bc8f80e30 "Move generate_coordkw_grid into grid_fixtures.hpp" speeds up tests by a factor of roughly 10. The additional tests in a04ecc380 "Add more grid tests" bring the time needed back up to what it was before the PR.

PR #1145, opened 2026-04-27
No manual memory management in rd grid

eivindjahren:


lib/resdata/rd_grid.cpp
```diff
@@ -4656,16 +4649,17 @@ bool rd_grid_use_mapaxes(const rd_grid_type *grid) { return grid->use_mapaxes; }
 
 void rd_grid_init_mapaxes_data_double(const rd_grid_type *grid,
                                       double *mapaxes) {
-    for (int i = 0; i < 6; i++)
-        mapaxes[i] = grid->mapaxes[i];
+    if (grid->mapaxes.has_value())
+        for (int i = 0; i < 6; i++)
+            mapaxes[i] = grid->mapaxes.value()[i];
```
ajaust, line 4557: Is this equivalent? If `mapaxes` has no value, we don't do anything. The previous code would always do some assignment.

lib/resdata/rd_grid.cpp
```diff
@@ -1844,10 +1839,9 @@ static void rd_grid_init_mapaxes(rd_grid_type *rd_grid, bool apply_mapaxes,
 
 static void rd_grid_add_lgr(rd_grid_type *main_grid, rd_grid_type *lgr_grid) {
     main_grid->LGR_list.emplace_back(lgr_grid, &rd_grid_free);
-    if (lgr_grid->lgr_nr >= int_vector_size(main_grid->lgr_index_map))
-        int_vector_resize(main_grid->lgr_index_map, lgr_grid->lgr_nr + 1, 0);
-    int_vector_iset(main_grid->lgr_index_map, lgr_grid->lgr_nr,
-                    main_grid->LGR_list.size() - 1);
+    if (lgr_grid->lgr_nr >= (int)main_grid->lgr_index_map.size())
```
ajaust: I am suggesting static casts since we also used `static_cast` at other parts of the code that were changed as part of the PR.
```suggestion
    if (lgr_grid->lgr_nr >= static_cast<int>(main_grid->lgr_index_map.size()))
```

lib/resdata/rd_grid.cpp
```diff
@@ -4198,7 +4189,7 @@ bool rd_grid_has_lgr(const rd_grid_type *main_grid, const char *__lgr_name) {
 bool rd_grid_has_lgr_nr(const rd_grid_type *main_grid, int lgr_nr) {
     __assert_main_grid(main_grid);
     {
-        if (int_vector_size(main_grid->lgr_index_map) > lgr_nr)
+        if ((int)main_grid->lgr_index_map.size() > lgr_nr)
```
ajaust:
```suggestion
        if (static_cast<int>(main_grid->lgr_index_map.size()) > lgr_nr)
```

lib/resdata/rd_grid.cpp
```diff
@@ -1728,11 +1721,12 @@ static void rd_grid_update_index(rd_grid_type *rd_grid) {
...
-        vector_iset_owned_ref(rd_grid->coarse_cells, coarse_nr,
-                              rd_coarse_cell_alloc(), rd_coarse_cell_free__);
+    while ((int)rd_grid->coarse_cells.size() <= coarse_nr)
+        rd_grid->coarse_cells.emplace_back(nullptr, &rd_coarse_cell_free);
+    if (!rd_grid->coarse_cells[coarse_nr])
+        rd_grid->coarse_cells[coarse_nr].reset(rd_coarse_cell_alloc());
 
-    return (rd_coarse_cell_type *)vector_iget(rd_grid->coarse_cells, coarse_nr);
+    return rd_grid->coarse_cells[coarse_nr].get();
 }
```
ajaust, line 1734:
```suggestion
    while (static_cast<int>(rd_grid->coarse_cells.size()) <= coarse_nr)
```

lib/resdata/rd_grid.cpp
```diff
@@ -834,30 +834,30 @@ static void rd_cell_dump_ascii(rd_cell_type &cell, int i, int j, int k,
...
-        rd_kw_iset_int(coords_kw, 6, cell.coarse_group + 1);
+        rd_kw_iset_int(coords_kw.get(), 5, cell.host_cell + 1);
+        rd_kw_iset_int(coords_kw.get(), 6, cell.coarse_group + 1);
     }
 
-    rd_kw_fwrite(coords_kw, fortio);
+    rd_kw_fwrite(coords_kw.get(), fortio.get());
     {
-        float *corners = (float *)rd_kw_get_void_ptr(corners_kw);
+        float *corners = (float *)rd_kw_get_void_ptr(corners_kw.get());
```
ajaust:
```suggestion
        float *corners = static_cast<float *>(rd_kw_get_void_ptr(corners_kw.get()));
```

lib/resdata/rd_grid.cpp
```diff
@@ -4987,11 +4984,11 @@ void rd_grid_init_coord_data_double(const rd_grid_type *grid, double *coord) {
...
-    if (grid->coord_kw == NULL) {
-        grid->coord_kw =
-            rd_kw_alloc(COORD_KW, rd_grid_get_coord_size(grid), RD_FLOAT);
-        rd_grid_init_coord_data(grid,
-                                (float *)rd_kw_get_void_ptr(grid->coord_kw));
+    if (!grid->coord_kw) {
+        grid->coord_kw.reset(
+            rd_kw_alloc(COORD_KW, rd_grid_get_coord_size(grid), RD_FLOAT));
+        rd_grid_init_coord_data(
+            grid, (float *)rd_kw_get_void_ptr(grid->coord_kw.get()));
```
ajaust:
```suggestion
            grid, static_cast<float *>(rd_kw_get_void_ptr(grid->coord_kw.get())));
```

lib/geometry/geo_region.cpp
```diff
@@ -174,3 +174,8 @@ const int_vector_type *geo_region_get_index_list(geo_region_type *region) {
     geo_region_assert_index_list(region);
     return region->index_list;
 }
+
+geo_region_ptr make_geo_region(const geo_pointset_ptr &pointset,
+                               bool preselect) {
+    return {geo_region_alloc(pointset.get(), preselect), &geo_region_free};
```
ajaust, line 182: We might need a comment in the header to point out that the region is borrowing the pointer of the pointset. This is relevant because it implies that the pointset must outlive the region or otherwise the region will have a dangling pointer reference.

Alternatively, we could make sure that `geo_region_alloc` makes a copy of the pointset.

ajaust (commented):
Great work! I have a few comments, but nothing too serious. 🙂

lib/resdata/rd_grid.cpp
eivindjahren: True, but I believe it would do a nullptr dereference in that case. There is a bug for it https://github.com/equinor/resdata/issues/1117

lib/resdata/rd_grid.cpp

lib/resdata/rd_grid.cpp
eivindjahren: For some reason outdated now, but I believe it is what I fixed in rd_grid_has_lgr_nr.

lib/resdata/rd_grid.cpp
eivindjahren: fixed

lib/resdata/rd_grid.cpp
eivindjahren: fixed

lib/geometry/geo_region.cpp
eivindjahren: Added a comment, I think that is sufficient for now as this was always a problem. I think it would make more sense to use a shared_ptr to really fix it.

lib/resdata/rd_grid.cpp
ajaust: I don't understand what this suggestion is supposed to do nor where it comes from. I have no memory of making this suggestion.

ajaust (approved):
Nice!

lib/geometry/geo_region.cpp
ajaust: Yeah, a shared_ptr would be a good idea.I am happy with the current documentation by the comment if you prefer to properly fix this in the another PR.

eivindjahren: Ok, lets keep it mind for later then.

PR #1147, opened 2026-04-29
Drop support for win32 and Make grid use pybind

eivindjahren:
I plan to squash all commits regarding the switch to pybind.


lib/resdata/rd_grid_pybind.cpp
```diff
@@ -0,0 +1,423 @@
...
+
+    py::int_ address = obj.attr("_BaseCClass__c_pointer");
+    void *pointer = PyLong_AsVoidPtr(address.ptr());
+
+    return reinterpret_cast<T *>(pointer);
+}
+
+PYBIND11_MODULE(_grid, m) {
+    m.doc() = "pybind11 bindings between rd_grid.py and rd_grid.cpp";
+
```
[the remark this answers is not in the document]
eivindjahren: This clearly works so no problems.

pyproject.toml
```diff
@@ -29,7 +29,7 @@ testpaths = ["tests"]
 [tool.cibuildwheel]
 build-frontend = "build[uv]"
 build = "cp311-* cp312-* cp313-* cp314-*"
-skip = "*-musllinux_*"
+skip = "*-musllinux_* *-win32"
```
ajaust, line 32: Do you think that we need to add more reasoning to the commit message such that GitHub can generate a good release note? How do we normally do this?

lib/CMakeLists.txt
```diff
@@ -158,6 +158,34 @@ install(DIRECTORY include/ DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
 install(DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/include/
         DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
 
+# -----------------------------------------------------------------
+# Pybind11 extension module: resdata.grid._grid
+# -----------------------------------------------------------------
+find_package(pybind11 CONFIG REQUIRED)
```
ajaust, line 165: Consider changing this to 

```suggestion
set(PYBIND11_FINDPYTHON ON)
find_package(pybind11 CONFIG REQUIRED)
```

or

```suggestion
find_package(Python COMPONENTS Interpreter Development.Module)
find_package(pybind11 CONFIG REQUIRED)
```


to get rid of the following warning when building the project

```text
-- Conan: Including build module from '/Users/AEJ/.conan2/p/pybin9183fffcad0dc/p/lib/cmake/pybind11/pybind11Common.cmake'
CMake Warning (dev) at /Users/AEJ/.conan2/p/pybin9183fffcad0dc/p/lib/cmake/pybind11/FindPythonLibsNew.cmake:101 (message):
  Policy CMP0148 is not set: The FindPythonInterp and FindPythonLibs modules
  are removed.  Run "cmake --help-policy CMP0148" for policy details.  Use
  the cmake_policy command to set the policy and suppress this warning, or
  preferably upgrade to using FindPython, either by calling it explicitly
  before pybind11, or by setting PYBIND11_FINDPYTHON ON before pybind11.
  /Users/AEJ/.conan2/p/pybin9183fffcad0dc/p/lib/cmake/pybind11/pybind11Tools.cmake:50 (find_package)
  /Users/AEJ/.conan2/p/pybin9183fffcad0dc/p/lib/cmake/pybind11/pybind11Common.cmake:228 (include)
  build/pybind11-config.cmake:41 (include)
  lib/CMakeLists.txt:164 (find_package)
This warning is for project developers.  Use -Wno-dev to suppress it.
```

conanfile.py
```diff
@@ -9,6 +9,7 @@ def requirements(self):
         self.requires("backward-cpp/1.6")
         self.requires("catch2/2.13.9")
         self.requires("fmt/8.0.1")
+        self.requires("pybind11/2.13.6")
```
ajaust: Why are we using pybind11 v2 which is rather old? [pybind11 v3](https://pybind11.readthedocs.io/en/stable/changelog.html#version-3-0-0-july-10-2025) was released last year and introduced support for new Python 3.14 and 3.14t. The most recent release is [3.0.4](https://pybind11.readthedocs.io/en/stable/changelog.html#version-3-0-4-april-18-2026).

lib/resdata/rd_grid_pybind.cpp
```diff
@@ -0,0 +1,20 @@
...
+    m.def(
+        "_fread_alloc",
+        [](std::string filename, bool apply_mapaxes) {
+            return reinterpret_cast<std::uintptr_t>(
+                rd_grid_load_case__(filename.c_str(), apply_mapaxes));
+        },
+        py::return_value_policy::reference);
```
ajaust, line 34: This is a pre-existing issue, but now we could do something like this
```suggestion
     [](std::string filename, bool apply_mapaxes) {
         auto *grid = rd_grid_load_case__(filename.c_str(), apply_mapaxes);
         if (!grid)
             throw std::runtime_error(
                 "Failed to load grid from: " + filename);
         return reinterpret_cast<std::uintptr_t>(grid);
     });
```
to prevent returning a nullptr.

lib/CMakeLists.txt
ajaust, line 165: This adds pybind11 as strictly necessary dependency. Do we want this? We could introduce a `BUILD_PYTHON` flag and only build the Python extension if this required. This would allow users to build the pure C/C++ part of the library without any Python dependencies.

lib/CMakeLists.txt
```diff
@@ -158,6 +158,34 @@ install(DIRECTORY include/ DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
...
+  set(_grid_install_dir "python/resdata/grid")
+  # libresdata.so.6 is installed at python/resdata/.libs/ make _grid.so find it
+  if(APPLE)
+    set_target_properties(_grid PROPERTIES INSTALL_RPATH
+                                           "@loader_path/../.libs")
+  else()
+    set_target_properties(_grid PROPERTIES INSTALL_RPATH "$ORIGIN/../.libs")
+  endif()
+else()
+  set(_grid_install_dir "${CMAKE_INSTALL_LIBDIR}")
```
ajaust, line 180: Should we set the rpath also for other build paths?
```suggestion
  set(_grid_install_dir "${CMAKE_INSTALL_LIBDIR}")
  if(APPLE)
    set_target_properties(_grid PROPERTIES INSTALL_RPATH
                                           "@loader_path")
  else()
    set_target_properties(_grid PROPERTIES INSTALL_RPATH
                                              "$ORIGIN")
  endif()                                                                                     
```

lib/resdata/rd_grid_pybind.cpp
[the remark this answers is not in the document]
ajaust: I think the copilot comment is more about how the code reads rather than that it is broken. The anonymous namespace indicates to the reader of the code that everything here should have internal linkage. However, the pybind module should be visible outside and therefore placing it inside the namespace might confuse people reading the code. The pybind module is visible after all due to how pybind11 works (it sets some visibility attribute explicitly).

lib/resdata/rd_grid_pybind.cpp
```diff
@@ -6,6 +6,16 @@
...
 
 namespace {
+template <typename T> T *from_cwrap(py::handle obj) {
+    if (obj.is_none())
+        return nullptr;
+
+    py::int_ address = obj.attr("_BaseCClass__c_pointer");
+    void *pointer = PyLong_AsVoidPtr(address.ptr());
+
+    return reinterpret_cast<T *>(pointer);
```
ajaust, line 22: Same pre-existing pattern from before that we could guard against
```suggestion
    if (!grid)
        throw std::runtime_error(
            "Failed to create grid from GRDECL keywords");
    return reinterpret_cast<std::uintptr_t>(grid);
```

lib/resdata/rd_grid_pybind.cpp
```diff
@@ -37,5 +37,18 @@ PYBIND11_MODULE(_grid, m) {
...
+        "_alloc_rectangular",
+        [](int nx, int ny, int nz, double dx, double dy, double dz,
+           std::optional<std::vector<int>> actnum) {
+            if (actnum.has_value())
+                return reinterpret_cast<std::uintptr_t>(
+                    rd_grid_alloc_rectangular(nx, ny, nz, dx, dy, dz,
+                                              actnum->data()));
+            else
+                return reinterpret_cast<std::uintptr_t>(
+                    rd_grid_alloc_rectangular(nx, ny, nz, dx, dy, dz, nullptr));
```
ajaust, line 56: `rd_grid_alloc_rectangular` may return a null pointer. We could guard against it

```suggestion
            auto *grid = rd_grid_alloc_rectangular(
                nx, ny, nz, dx, dy, dz,
                actnum.has_value() ? actnum->data() : nullptr);
            if (!grid)
                throw std::runtime_error(
                    "Failed to allocate rectangular grid");
            return reinterpret_cast<std::uintptr_t>(grid);
```

The diff looks horrible for me. Not sure what I did wrong.

python/resdata/grid/rd_grid.py
```diff
@@ -262,39 +259,28 @@ def create(cls, specgrid, zcorn, coord, actnum, mapaxes=None):
...
-                dims[0], dims[1], dims[2], dV[0], dV[1], dV[2], actnum.getDataPtr()
+            rd_grid = _grid._alloc_rectangular(
+                dims[0], dims[1], dims[2], dV[0], dV[1], dV[2], list(actnum)
             )
-
-        # If we have not succeeded in creatin the grid we *assume* the
-        # error is due to a failed malloc.
-        if rd_grid is None:
+        if not rd_grid:
             raise MemoryError("Failed to allocated regualar grid")
```
ajaust, line 157: Preexisting, but I cannot unsee it

```suggestion
            raise MemoryError("Failed to allocated regular grid")
```

python/resdata/grid/rd_grid.py
ajaust, line 873: This might not needed anymore. Doesn't `_reference_from_ptr(` set the parent by internally calling `self.createCReference(ptr, parent=self)` with `parent=self`?

lib/resdata/rd_grid_pybind.cpp
```diff
@@ -50,5 +50,12 @@ PYBIND11_MODULE(_grid, m) {
                     rd_grid_alloc_rectangular(nx, ny, nz, dx, dy, dz, nullptr));
         },
         py::return_value_policy::reference);
+    m.def(
+        "_get_numbered_lgr",
+        [](py::handle self, int lgr_nr) {
+            return reinterpret_cast<std::uintptr_t>(rd_grid_get_lgr_from_lgr_nr(
+                from_cwrap<rd_grid_type>(self), lgr_nr));
```
ajaust, line 63:
```suggestion
         auto *lgr = rd_grid_get_lgr_from_lgr_nr(
             from_cwrap<rd_grid_type>(self), lgr_nr);
         if (!lgr)
             throw std::out_of_range(
                 "No LGR with number: " + std::to_string(lgr_nr));
```

lib/resdata/rd_grid_pybind.cpp
```diff
@@ -78,5 +78,8 @@ PYBIND11_MODULE(_grid, m) {
         return rd_grid_cell_in_coarse_group1(from_cwrap<rd_grid_type>(self),
                                              index);
     });
+    m.def("_free", [](py::handle self) {
+        return rd_grid_free(from_cwrap<rd_grid_type>(self));
```
ajaust: Are we returning here on purpose? `rd_grid_free` is marked as void.

```suggestion
        rd_grid_free(from_cwrap<rd_grid_type>(self));
```

lib/resdata/rd_grid_pybind.cpp
```diff
@@ -169,5 +169,10 @@ PYBIND11_MODULE(_grid, m) {
                 "Could not find the point:({:g},{:g}) in layer:{:d}", x, y, k));
         return std::make_tuple(i, j);
     });
+    m.def("_get_ijk_xyz",
+          [](py::handle self, double x, double y, double z, int start_index) {
+              return rd_grid_get_global_index_from_xyz(
+                  from_cwrap<rd_grid_type>(self), x, y, z, start_index);
```
ajaust, line 177: This one follows a different pattern than the previous commit, but it looks like it is a very similar method. In the previous commit, an exception is thrown if the point is not found. Here, we return `None` if the global index is not found.

Should we have something like this:

```c++
        int global_index =  rd_grid_get_global_index_from_xyz(
                  from_cwrap<rd_grid_type>(self), x, y, z, start_index);
        if (global_index == -1)
            throw std::domain_error(fmt::format(
                "Could not find global index of point:({:g},{:g},{:g})", x, y, z));
        return global_index;
```
Formatting/indentation in my suggestion might be off.

lib/resdata/rd_grid_pybind.cpp
```diff
@@ -213,5 +213,17 @@ PYBIND11_MODULE(_grid, m) {
...
+                               py::handle file, double default_value) {
+        if (header.has_value())
+            return rd_grid_grdecl_fprintf_kw(
+                from_cwrap<rd_grid_type>(self), from_cwrap<rd_kw_type>(kw),
+                header->c_str(), from_cwrap<FILE>(file), default_value);
+        else
+            return rd_grid_grdecl_fprintf_kw(
+                from_cwrap<rd_grid_type>(self), from_cwrap<rd_kw_type>(kw),
+                nullptr, from_cwrap<FILE>(file), default_value);
+    });
```
ajaust, line 229: `rd_grid_grdecl_fprintf_kw` is defined as `void` method. Could we remove the return statements or does this fulfill an purpose?

ajaust, line 229: `rd_grid_grdecl_fprintf_kw` is defined as `void` method. Could we remove the return statements or does this fulfill an purpose?

```suggestion
        const char *h = header.has_value() ? header->c_str() : nullptr;
        rd_grid_grdecl_fprintf_kw(
             from_cwrap<rd_grid_type>(self), from_cwrap<rd_kw_type>(kw),
             h, from_cwrap<FILE>(file), default_value);
```

lib/resdata/rd_grid_pybind.cpp
```diff
@@ -225,5 +225,20 @@ PYBIND11_MODULE(_grid, m) {
                 from_cwrap<rd_grid_type>(self), from_cwrap<rd_kw_type>(kw),
                 nullptr, from_cwrap<FILE>(file), default_value);
     });
+    m.def("_load_column",
+          [](py::handle self, py::handle kw, int i, int j, py::handle column) {
+              return rd_grid_get_column_property(
+                  from_cwrap<rd_grid_type>(self), from_cwrap<rd_kw_type>(kw), i,
+                  j, from_cwrap<double_vector_type>(column));
```
ajaust, line 234: This is returning from a `void` method as well.
```suggestion
             rd_grid_get_column_property(
                  from_cwrap<rd_grid_type>(self), from_cwrap<rd_kw_type>(kw), i,
                  j, from_cwrap<double_vector_type>(column));
```

lib/resdata/rd_grid_pybind.cpp
```diff
@@ -258,5 +258,11 @@ PYBIND11_MODULE(_grid, m) {
                              &dx, &dy, &dz);
         return std::make_tuple(dx, dy, dz);
     });
+    m.def("_fprintf_grdecl2", [](py::handle self, py::handle file,
+                                 int rd_unit) {
+        return rd_grid_fprintf_grdecl2(from_cwrap<rd_grid_type>(self),
+                                       from_cwrap<FILE>(file),
+                                       static_cast<ert_rd_unit_enum>(rd_unit));
```
ajaust: Returns a `void` method.

lib/resdata/rd_grid_pybind.cpp
```diff
@@ -264,5 +264,17 @@ PYBIND11_MODULE(_grid, m) {
...
+                              int rd_unit) {
+        return rd_grid_fwrite_GRID2(from_cwrap<rd_grid_type>(self),
+                                    filename.c_str(),
+                                    static_cast<ert_rd_unit_enum>(rd_unit));
+    });
+    m.def("_fwrite_EGRID2", [](py::handle self, std::string filename,
+                               int rd_unit) {
+        return rd_grid_fwrite_EGRID2(from_cwrap<rd_grid_type>(self),
+                                     filename.c_str(),
+                                     static_cast<ert_rd_unit_enum>(rd_unit));
```
ajaust: Both bindings return `void` methods.

lib/resdata/rd_grid_pybind.cpp
```diff
@@ -285,5 +285,27 @@ PYBIND11_MODULE(_grid, m) {
     m.def("_dual_grid", [](py::handle self) {
         return rd_grid_dual_grid(from_cwrap<rd_grid_type>(self));
     });
+    m.def(
+        "_init_actnum",
+        [](py::handle self) {
+            auto rd_grid = from_cwrap<rd_grid_type>(self);
+            int size = rd_grid_get_global_size(rd_grid);
+            int_vector_type *actnum = int_vector_alloc(size, 0);
```
ajaust, line 293: `int_vector_alloc` could return a `nullptr` so maybe we should check for this. I am not sure what the other methods do if the receive a nullptr instead of a pointer to a valid vector.

lib/resdata/rd_grid_pybind.cpp
```diff
@@ -285,5 +285,27 @@ PYBIND11_MODULE(_grid, m) {
...
+
+            return reinterpret_cast<std::uintptr_t>(actnum);
+        },
+        py::return_value_policy::reference);
+    m.def(
+        "_init_actnum_kw",
+        [](py::handle self) {
+            auto rd_grid = from_cwrap<rd_grid_type>(self);
+            int size = rd_grid_get_global_size(rd_grid);
+            rd_kw_type *actnum = rd_kw_alloc("ACTNUM", size, RD_INT);
```
ajaust, line 304: `rd_kw_alloc` could return a `nullptr` on allocation failure. Maybe we need a `nullptr` check  here.

lib/resdata/rd_grid_pybind.cpp
```diff
@@ -319,5 +319,15 @@ PYBIND11_MODULE(_grid, m) {
                                      from_cwrap<rd_kw_type>(kw_copy),
                                      from_cwrap<rd_kw_type>(kw));
           });
+    m.def(
+        "_create_volume_keyword",
+        [](py::handle self, bool active_size) {
+            return reinterpret_cast<std::uintptr_t>(rd_grid_alloc_volume_kw(
```
ajaust, line 328: `rd_grid_alloc_volume_kw` might return a `nullptr` if the internal allocation fails. Maybe we want a `nullptr` check before castin the pointer.

lib/resdata/rd_grid_pybind.cpp
```diff
@@ -329,5 +329,33 @@ PYBIND11_MODULE(_grid, m) {
...
+                rd_grid_alloc_actnum_kw(from_cwrap<rd_grid_type>(self)));
+        },
+        py::return_value_policy::reference);
+    m.def(
+        "_export_mapaxes",
+        [](py::handle self) {
+            return reinterpret_cast<std::uintptr_t>(
+                rd_grid_alloc_mapaxes_kw(from_cwrap<rd_grid_type>(self)));
+        },
+        py::return_value_policy::reference);
```
ajaust, line 362: All the allocations may return `nullptr` so we probably should check the returned pointer before casting it.

python/resdata/grid/rd_grid.py
```diff
@@ -1196,13 +1193,8 @@ def export_index(self, active_only=False):
             size = self.get_global_size()
         indx = np.zeros(size, dtype=np.int32)
         data = np.zeros([size, 4], dtype=np.int32)
```
ajaust: This preallocation can be removed. It is never used.

pyproject.toml
ajaust, line 32: Do you think that we need to add more reasoning justifying the change to the commit message such that GitHub can generate a good release note? How do we normally do this?

ajaust (commented):
-  I did not verify the code in commit 11ccbe0483362e5b0da84519d0552b24d27df665 `Translate cpp grid tests to python` in detail. If I should check it thoroughly, let me know

-  If I remember correctly, the purpose of this change is that exceptions in C++ end up as Python exceptions. Should we add any tests verifying that this is the case?

-  Some potentially missing nullptr checks when allocations are to be done. This is a preexisting issue, but we could address in the PR.

-  We quite often `return` from a method with return type `void`. I think we could remove the `return` statement in this case, but I may miss something.

pyproject.toml
eivindjahren: I think its fine as it is in the PR title.

conanfile.py
eivindjahren: 3.0.1 was the newest in conan so I added that

lib/resdata/rd_grid_pybind.cpp
eivindjahren: So I wanted to avoid changing the interface. This now returns `Noneq as it did before. Granted we are changing aborts to throws but I consider that admissible as it was not possible to recover from the abort signal.

lib/CMakeLists.txt
eivindjahren: We decided some time ago that the cpp interface of resdata is not public and I think we should stick with that for simplicity.

lib/resdata/rd_grid_pybind.cpp
eivindjahren: Added

lib/resdata/rd_grid_pybind.cpp
eivindjahren: I think we should keep the behavior the same as this turned into a None on the python side before.

lib/resdata/rd_grid_pybind.cpp
eivindjahren: As mentioned, I think sticking to the return of None that already existed is best.

python/resdata/grid/rd_grid.py
eivindjahren: Done

pyproject.toml
eivindjahren: Usually we include this in the pr title. As we can't build for win32, we don't have any choice.

conanfile.py
ajaust: It still shows as `self.requires("pybind11/2.13.6")` in my diff. Did you the changes already?

ajaust (commented):
The fixups look good. There are still a few open comments where I wonder if we to keep things as they are for consistency with old behavior or if they were missed (mostly return of `void`).

ajaust (approved):
I see that this one comment about using a pointer that might be a `nullptr` is still open: https://github.com/equinor/resdata/pull/1147/changes#r3180949754

Everything else was addressed.

PR #1148, opened 2026-04-30
Change util_abort to throw in rd_grid.cpp

eivindjahren:


lib/resdata/rd_grid.cpp
```diff
@@ -5173,38 +5192,44 @@ static void rd_grid_fwrite_self_nnc(const rd_grid_type *grid,
...
                 int nnc_index = nnc_vector_iget_nnc_index(nnc_vector, i);
-                int_vector_iset(g1, nnc_index, 1 + g);
-                int_vector_iset(g2, nnc_index,
-                                1 + nnc_vector_iget_grid_index(nnc_vector, i));
+                if (static_cast<size_t>(nnc_index) >= g1.size()) {
+                    g1.resize(nnc_index + 1, default_index);
+                    g2.resize(nnc_index + 1, default_index);
+                }
+                g1[nnc_index] = 1 + g;
+                g2[nnc_index] = 1 + nnc_vector_iget_grid_index(nnc_vector, i);
```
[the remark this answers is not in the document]

lib/resdata/rd_grid.cpp
```diff
@@ -5161,10 +5184,10 @@ void rd_grid_reset_actnum(rd_grid_type *grid, const int *actnum) {
 }
 
 static void rd_grid_fwrite_self_nnc(const rd_grid_type *grid,
-                                    fortio_type *fortio) {
+                                    const ERT::FortIO &fortio) {
     const int default_index = 1;
-    int_vector_type *g1 = int_vector_alloc(0, default_index);
-    int_vector_type *g2 = int_vector_alloc(0, default_index);
+    std::vector<int> g1(0, default_index);
+    std::vector<int> g2(0, default_index);
```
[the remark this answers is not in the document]
ajaust: Agree, we can drop the explicit initialisation if we initialise with size `0` anyway.

lib/resdata/rd_grid.cpp
```diff
@@ -5173,38 +5177,50 @@ static void rd_grid_fwrite_self_nnc(const rd_grid_type *grid,
...
-        rd_kw_type *nnc1_kw = rd_kw_alloc_new_shared(NNC1_KW, num_nnc, RD_INT,
-                                                     int_vector_get_ptr(g1));
-        rd_kw_type *nnc2_kw = rd_kw_alloc_new_shared(NNC2_KW, num_nnc, RD_INT,
-                                                     int_vector_get_ptr(g2));
-        rd_kw_type *nnchead_kw = rd_kw_alloc(NNCHEAD_KW, NNCHEAD_SIZE, RD_INT);
+    int num_nnc = 0;
+    if (g1.size() > static_cast<size_t>(std::numeric_limits<int>::max()))
+        throw std::overflow_error(
+            "rd_grid_fwrite_self_nnc: NNC count exceeds INT_MAX");
+    num_nnc = static_cast<int>(g1.size());
```
ajaust:
```suggestion
    if (g1.size() > static_cast<size_t>(std::numeric_limits<int>::max()))
        throw std::overflow_error(
            "rd_grid_fwrite_self_nnc: NNC count exceeds INT_MAX");
    int num_nnc = static_cast<int>(g1.size());
```

lib/resdata/rd_grid.cpp
```diff
@@ -2219,9 +2220,10 @@ static rd_grid_ptr rd_grid_alloc_GRDECL_kw__(
   */
```
ajaust, line 2221: 1. Does this comment still apply?
2. If it still applies, should we somehow mention that `util_abort` is deprecated?

lib/resdata/rd_grid.cpp
```diff
@@ -1852,12 +1857,14 @@ static void rd_grid_init_mapaxes(rd_grid_type *rd_grid, bool apply_mapaxes,
...
-    main_grid->LGR_list.emplace_back(lgr_grid, &rd_grid_free);
+static rd_grid_type *rd_grid_add_lgr(rd_grid_ptr &main_grid,
+                                     rd_grid_ptr &lgr_grid) {
     if (lgr_grid->lgr_nr >= static_cast<int>(main_grid->lgr_index_map.size()))
         main_grid->lgr_index_map.resize(lgr_grid->lgr_nr + 1, 0);
-    main_grid->lgr_index_map[lgr_grid->lgr_nr] = main_grid->LGR_list.size() - 1;
-    main_grid->LGR_hash[lgr_grid->name] = lgr_grid;
+    main_grid->lgr_index_map[lgr_grid->lgr_nr] = main_grid->LGR_list.size();
+    main_grid->LGR_hash[lgr_grid->name] = lgr_grid.get();
+    main_grid->LGR_list.push_back(std::move(lgr_grid));
```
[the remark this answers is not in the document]
ajaust: As discussed, it should be safe since we destroy the `main_grid` if `std::vector` throws. Nevertheless, I think that is a reasonable suggestion to change the order here to make it clearer.

lib/resdata/rd_grid.cpp
```diff
@@ -2534,89 +2544,93 @@ static rd_grid_type *rd_grid_alloc_EGRID__(rd_grid_type *main_grid,
...
             "%s: %s wrong file type - expected .EGRID file - aborting \n",
             __func__, grid_file);
     {
-        rd_file_type *rd_file = rd_file_open(grid_file, 0);
-        if (rd_file) {
-            int num_grid = rd_file_get_num_named_kw(rd_file, GRIDHEAD_KW);
-            rd_grid_type *main_grid = rd_grid_alloc_EGRID__(
-                NULL, rd_file, 0, apply_mapaxes, ext_actnum);
+        auto rd_file = rd_file_ptr(rd_file_open(grid_file, 0), &rd_file_close);
+        if (rd_file.get()) {
```
ajaust: I think you can keep this as
```suggestion
        if (rd_file) {
```
This type of "does it point to a valid object" check should also work with smart pointers.

lib/resdata/rd_grid.cpp
```diff
@@ -2534,89 +2544,93 @@ static rd_grid_type *rd_grid_alloc_EGRID__(rd_grid_type *main_grid,
...
         util_abort(
             "%s: %s wrong file type - expected .EGRID file - aborting \n",
             __func__, grid_file);
     {
-        rd_file_type *rd_file = rd_file_open(grid_file, 0);
-        if (rd_file) {
-            int num_grid = rd_file_get_num_named_kw(rd_file, GRIDHEAD_KW);
-            rd_grid_type *main_grid = rd_grid_alloc_EGRID__(
-                NULL, rd_file, 0, apply_mapaxes, ext_actnum);
+        auto rd_file = rd_file_ptr(rd_file_open(grid_file, 0), &rd_file_close);
```
ajaust, line 2570: I think the new typedef (next commit) in `[lib/include/resdata/rd_file.hpp‎](https://github.com/equinor/resdata/pull/1148/changes/0da16adaf82cd3ad50e95369603c53c6d0db73ce#diff-1253e33076ff5d2af5dc02705486506e1afa1f0ad4694e429a523abd7761f6d9)` must move into this commit. At the moment I get the following error message when trying to compile this commit:

```
lib/CMakeFiles/resdata.dir/resdata/rd_grid.cpp.o
/Users/AEJ/projects/github/resdata/lib/resdata/rd_grid.cpp:2568:24: fatal error: use of undeclared identifier 'rd_file_ptr'; did you mean 'rd_file_open'?
 2568 |         auto rd_file = rd_file_ptr(rd_file_open(grid_file, 0), &rd_file_close);
      |                        ^~~~~~~~~~~
      |                        rd_file_open
/Users/AEJ/projects/github/resdata/lib/include/resdata/rd_file.hpp:24:15: note: 'rd_file_open' declared here
   24 | rd_file_type *rd_file_open(const char *filename, int flags);
      |               ^
1 error generated.
make[2]: *** [lib/CMakeFiles/resdata.dir/resdata/rd_grid.cpp.o] Error 1
make[1]: *** [lib/CMakeFiles/resdata.dir/all] Error 2
make: *** [all] Error 2
```

lib/resdata/rd_grid.cpp
```diff
@@ -4380,9 +4393,10 @@ double rd_grid_get_property(const rd_grid_type *rd_grid,
...
 
     } else {
-        util_abort("%s: sorry - can not lookup type:%s with %s.\n", __func__,
-                   rd_type_alloc_name(data_type), __func__);
-        return -1;
+        throw std::invalid_argument(fmt::format(
+            "rd_grid_get_property: can not lookup type:{} - keyword type must "
+            "be numeric.",
+            rd_type_alloc_name(data_type)));
     }
```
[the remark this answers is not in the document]
eivindjahren: fixed

lib/resdata/rd_grid.cpp
```diff
@@ -4430,8 +4444,10 @@ void rd_grid_get_column_property(const rd_grid_type *rd_grid,
             }
         }
     } else
-        util_abort("%s: sorry - can not lookup type:%s with %s.\n", __func__,
-                   rd_type_alloc_name(data_type), __func__);
+        throw std::invalid_argument(fmt::format(
+            "rd_grid_get_column_property: can not lookup type:{} - keyword "
+            "type must be numeric.",
+            rd_type_alloc_name(data_type)));
```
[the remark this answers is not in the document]
eivindjahren: fixed

lib/resdata/rd_grid.cpp
```diff
@@ -4463,14 +4479,16 @@ void rd_grid_grdecl_fprintf_kw(const rd_grid_type *rd_grid,
...
+                    "Only 0 and 1 are allowed for bool interpolation, got {}",
+                    tmp));
             default_ptr = &bool_default;
         }
 
         if (default_ptr == NULL)
-            util_abort("%s: invalid type \n", __func__);
+            throw std::invalid_argument(
+                fmt::format("rd_grid_grdecl_fprintf_kw: invalid type {}",
+                            rd_type_alloc_name(rd_kw_get_data_type(rd_kw))));
```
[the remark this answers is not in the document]
eivindjahren: fixed

lib/resdata/rd_grid.cpp
```diff
@@ -2534,89 +2544,93 @@ static rd_grid_type *rd_grid_alloc_EGRID__(rd_grid_type *main_grid,
...
-            rd_grid_init_nnc(main_grid, rd_file);
-            rd_grid_init_nnc_amalgamated(main_grid, rd_file);
-
-            rd_file_close(rd_file);
+            rd_grid_init_nnc(main_grid.get(), rd_file.get());
+            rd_grid_init_nnc_amalgamated(main_grid.get(), rd_file.get());
             return main_grid;
         } else
             return NULL;
     }
```
ajaust, line 2605: This would probably be a compile error as well. I don't think a `rd_grid_ptr` can be implicitly constructed from `NULL`.

This compile error vanishes in the next commit due to the changes introduced there.

lib/resdata/rd_grid.cpp
```diff
@@ -1385,12 +1385,12 @@ static int rd_grid_get_global_index__(const rd_grid_type *rd_grid, int i, int j,
     return i + j * rd_grid->nx + k * rd_grid->nx * rd_grid->ny;
 }
 
-static void rd_grid_set_cell_EGRID(rd_grid_type *rd_grid, int i, int j, int k,
+static void rd_grid_set_cell_EGRID(rd_grid_ptr &rd_grid, int i, int j, int k,
```
ajaust: I will just put this here since it applies, but it applies to the full commit 0da16adaf82cd3ad50e95369603c53c6d0db73ce.

I think your idea of using some `std::move` is good. I think it also makes sense to keep some of the raw pointers where methods only borrow the object. I think we can also eliminate some of the lambdas.

I created a suggestion here: https://github.com/ajaust/resdata/commit/ecf8f4401164a42ed3eb2389d743560ea857af07

This is a fixup against this commit (0da16adaf82cd3ad50e95369603c53c6d0db73ce).

- Diff shown against this commit: https://github.com/ajaust/resdata/commit/ecf8f4401164a42ed3eb2389d743560ea857af07
- Diff shown against `main`: https://github.com/ajaust/resdata/compare/main..ecf8f4401164a42ed3eb2389d743560ea857af07

ajaust (commented):
I think the changes are very good. I have some minor comments and I tend to agree with copilot's review.

Regarding the third commit. I think this would work, but I also made a suggestion how we can make it look like. There is a comment linking to my fork with suggested code in the commit. Please check it out and see how you like it.

lib/resdata/rd_grid.cpp
eivindjahren: I am not sure, I think we can keep it as is though as where you can or cannot put util_abort is not easy to summarize.

lib/resdata/rd_grid.cpp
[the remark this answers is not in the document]
eivindjahren: I fixed it to do it in that order.

lib/resdata/rd_grid.cpp

lib/resdata/rd_grid.cpp
eivindjahren: I took in your changes. Thank you!

lib/resdata/rd_grid.cpp

lib/resdata/rd_grid.cpp
eivindjahren: Ah, I got them in the wrong order. I just changed the order of these commits.

lib/resdata/rd_grid.cpp
eivindjahren: Fixed by changing the order of the commits.

lib/resdata/rd_grid.cpp
ajaust: We could also use RAII here, I think. The diff does not show the way I want. The idea is to wrap the grid into a `rd_grid_ptr` for automatic memory management in case something goes wrong in the loop. For example, `grid->cells.at(global_index)` might throw. 

```suggestion
    auto grid = rd_grid_ptr(
        rd_grid_alloc_empty(NULL, unit_system, FILEHEAD_SINGLE_POROSITY, 
                            nx, ny, nz, 0, true),
        &rd_grid_free);
    if (grid) {
        // Same loop as before
        rd_grid_update_index(grid.get());
    }

    return grid.release();
```

lib/resdata/rd_grid.cpp
```diff
@@ -5169,38 +5177,49 @@ static void rd_grid_fwrite_self_nnc(const rd_grid_type *grid,
...
-    }
+    rd_kw_scalar_set_int(nnchead_kw.get(), 0);
+    rd_kw_iset_int(nnchead_kw.get(), NNCHEAD_NUMNNC_INDEX, num_nnc);
+    rd_kw_iset_int(nnchead_kw.get(), NNCHEAD_LGR_INDEX, grid->lgr_nr);
 
-    int_vector_free(g1);
-    int_vector_free(g2);
+    rd_kw_fwrite(nnchead_kw.get(), fortio.get());
+    rd_kw_fwrite(nnc1_kw.get(), fortio.get());
+    rd_kw_fwrite(nnc2_kw.get(), fortio.get());
```
ajaust, line 5244: Could we rewrite this to avoid the (artificial?) allocation of memory if `g1` is empty (which means `num_nnc == 0`).

```suggestion
    int num_nnc = static_cast<int>(g1.size());
 
    auto nnchead_kw = make_rd_kw(NNCHEAD_KW, NNCHEAD_SIZE, RD_INT);
    rd_kw_scalar_set_int(nnchead_kw.get(), 0);
    rd_kw_iset_int(nnchead_kw.get(), NNCHEAD_NUMNNC_INDEX, num_nnc);
    rd_kw_iset_int(nnchead_kw.get(), NNCHEAD_LGR_INDEX, grid->lgr_nr);
    rd_kw_fwrite(nnchead_kw.get(), fortio.get());
 
    if (num_nnc > 0) {
        auto nnc1_kw = rd_kw_ptr(
            rd_kw_alloc_new_shared(NNC1_KW, num_nnc, RD_INT, g1.data()),
            &rd_kw_free);
        auto nnc2_kw = rd_kw_ptr(
            rd_kw_alloc_new_shared(NNC2_KW, num_nnc, RD_INT, g2.data()),
            &rd_kw_free);
        rd_kw_fwrite(nnc1_kw.get(), fortio.get());
        rd_kw_fwrite(nnc2_kw.get(), fortio.get());
    }
```

lib/resdata/rd_grid.cpp
ajaust: I think you swapped the wrong commits. `rd_file_ptr` is still not defined in the third commit but only in the fourth commit.

lib/resdata/rd_grid.cpp
ajaust: I think this is still an issue since the order of this and the next (last) commit is still the same.

ajaust (commented):
There are still some issues with commit 3 w.r.t. undefined type `rd_file_ptr` and implicit construction of a `unique_ptr` from `NULL`. 

I also added two more comments for potential changes. You have to see if they make sense and fit into this PR.
1. One more opportunity to use RAII for grid management
2. On potential simplification of code to avoid having to allocate memory in some vectors (`g1` and `g2`).

lib/resdata/rd_grid.cpp
eivindjahren: Then I swapped the third and fourth commits :D

lib/resdata/rd_grid.cpp
eivindjahren: Added.

lib/resdata/rd_grid.cpp
eivindjahren: From what I understand this would change what is written to the file so that we don't  nnc1 and nnc2 to the file if the data is empty. Unfortunately the header is still written saying that the size is 0. This is probably fine, but I don't want to bet on it.

lib/resdata/rd_grid.cpp
ajaust: There is still something off. This still won't compile since rd_file_ptr is not defined here.

The current commit "Make it possible to throw while constructing the grid" uses `rd_file_ptr` which is not defined before the next commit "Change util_abort in rd_grid.cpp to exception".

I would recommend squashing it with the next commit or simply move the change in `rd_grid.hpp` (1 line change in the next commit) into the "Make it possible to throw while constructing the grid" commit.

lib/resdata/rd_grid.cpp
```diff
@@ -2530,89 +2537,91 @@ static rd_grid_type *rd_grid_alloc_EGRID__(rd_grid_type *main_grid,
...
             main_grid->name = grid_file;
-            rd_grid_init_nnc(main_grid, rd_file);
-            rd_grid_init_nnc_amalgamated(main_grid, rd_file);
-
-            rd_file_close(rd_file);
+            rd_grid_init_nnc(main_grid.get(), rd_file.get());
+            rd_grid_init_nnc_amalgamated(main_grid.get(), rd_file.get());
             return main_grid;
         } else
             return NULL;
```
ajaust, line 2594: Maybe we should just make it into just to make the code compile. Then replace it by an exception in the next commit.

```suggestion
            return rd_grid_ptr{nullptr, &rd_grid_free}
```

or

```suggestion
             util_abort("%s: failed to open file: %s\n", __func__, grid_file);
             return rd_grid_ptr{nullptr, &rd_grid_free};  // unreachable, satisfies return type
```

or just throw an exception immediately like in the next commit.

lib/resdata/rd_grid.cpp
```diff
@@ -1852,12 +1852,14 @@ static void rd_grid_init_mapaxes(rd_grid_type *rd_grid, bool apply_mapaxes,
...
+static rd_grid_type *rd_grid_add_lgr(rd_grid_type *main_grid,
+                                     rd_grid_ptr lgr_grid) {
     if (lgr_grid->lgr_nr >= static_cast<int>(main_grid->lgr_index_map.size()))
         main_grid->lgr_index_map.resize(lgr_grid->lgr_nr + 1, 0);
-    main_grid->lgr_index_map[lgr_grid->lgr_nr] = main_grid->LGR_list.size() - 1;
-    main_grid->LGR_hash[lgr_grid->name] = lgr_grid;
+    main_grid->lgr_index_map[lgr_grid->lgr_nr] = main_grid->LGR_list.size();
+    main_grid->LGR_hash[lgr_grid->name] = lgr_grid.get();
+    main_grid->LGR_list.push_back(std::move(lgr_grid));
+    return main_grid->LGR_list.back().get();
```
ajaust, line 1862: I think the reordering of commits introduced the unsafer order of commands here. I think we should immediately use the safe ordering that we introduce in commit 4

```suggestion
    auto ptr = lgr_grid.get();
    std::string lgr_name = lgr_grid->name;
    int lgr_nr = lgr_grid->lgr_nr;

    main_grid->LGR_list.push_back(std::move(lgr_grid));
    if (lgr_nr >= static_cast<int>(main_grid->lgr_index_map.size()))
        main_grid->lgr_index_map.resize(lgr_nr + 1, 0);
    main_grid->lgr_index_map[lgr_nr] = main_grid->LGR_list.size() - 1;
    main_grid->LGR_hash[lgr_name] = ptr;
    return ptr;
```

ajaust (commented):
I am technically happy with the exception that it is not possible to compile all commits (and run the tests). Additionally, reshuffling of the commits also has reintroduced an older of commands in some intermediate commits. 

I think the best is to scroll through "Make it possible to throw while constructing the grid" https://github.com/equinor/resdata/pull/1148/changes/69ad649810cb13468b0de2fc9203cc6b0698111d which introduces the offending changes. My comments should appear at the corresponding lines.

ajaust (approved):

PR #1152, opened 2026-05-05
Clean up functions defined only for use with pandas

eivindjahren:


lib/resdata/rd_grid_pybind.cpp
```diff
@@ -457,12 +457,18 @@ PYBIND11_MODULE(_grid, m) {
...
+        int32_t *index_ptr = static_cast<int32_t *>(idx_buffer.ptr);
+
+        std::fill(data_ptr, data_ptr + data_buffer.size, 0.0);
         double x[8], y[8], z[8];
-        int pos = 0;
-        for (int i = 0; i < index.size(); i++) {
-            int g = index.mutable_data()[i];
+        size_t pos = 0;
+
+        for (size_t i = 0; i < idx_buffer.size; i++) {
```
ajaust: Python uses a signed size type. We should either use the same type or cast explicitly to `size_t`. This applies to all the loops of this style.


```suggestion
        auto n = static_cast<size_t>(idx_buffer.size);
        for (size_t i = 0; i < n; i++) {
```

or

```suggestion
        for (ssize_t i = 0; i < idx_buffer.size; i++) {
```

or 

```suggestion
        for (size_t i = 0; i < static_cast<size_t>(idx_buffer.size); i++) {
```

lib/resdata/rd_grid_pybind.cpp
```diff
@@ -457,12 +457,18 @@ PYBIND11_MODULE(_grid, m) {
...
         py::array_t<double> data(shape);
-        auto data_ptr = data.mutable_data();
-        std::fill(data_ptr, data_ptr + data.size(), 0.0);
+        auto data_buffer = data.request();
+        double *data_ptr = static_cast<double *>(data_buffer.ptr);
+
+        auto idx_buffer = index.request();
+        int32_t *index_ptr = static_cast<int32_t *>(idx_buffer.ptr);
+
+        std::fill(data_ptr, data_ptr + data_buffer.size, 0.0);
```
ajaust, line 461: I technically like this, but we could skip this since the loop sets a values anyway this might not be needed anymore.

lib/resdata/rd_grid_pybind.cpp
```diff
@@ -434,13 +434,16 @@ PYBIND11_MODULE(_grid, m) {
...
+
         std::vector<ptrdiff_t> shape = {index.size(), 3};
         py::array_t<double> data(shape);
         auto data_ptr = data.mutable_data();
-        std::fill(data_ptr, data_ptr + data.size(), 0.0);
-        for (int i = 0; i < index.size(); i++) {
-            int g = index.mutable_data()[i];
-            int j = 3 * i;
+
+        for (size_t i = 0; i < idx_buffer.size; i++) {
```
ajaust: ssize_t vs size_t

lib/resdata/rd_grid_pybind.cpp
```diff
@@ -416,11 +416,14 @@ PYBIND11_MODULE(_grid, m) {
...
+        auto idx_buffer = index.request();
+        int32_t *index_ptr = static_cast<int32_t *>(idx_buffer.ptr);
+
+        py::array_t<double> data(idx_buffer.size);
         auto data_ptr = data.mutable_data();
-        std::fill(data_ptr, data_ptr + data.size(), 0.0);
-        for (int i = 0; i < index.size(); i++) {
-            int g = index.mutable_data()[i];
+
+        for (size_t i = 0; i < idx_buffer.size; i++) {
```
ajaust: ssize_t vs size_t

lib/resdata/rd_grid_pybind.cpp
```diff
@@ -395,17 +395,22 @@ PYBIND11_MODULE(_grid, m) {
...
+              int32_t *idx_ptr = static_cast<int32_t *>(idx_buffer.ptr);
 
-              for (int i = 0; i < size; i++) {
-                  int di = indx_ptr[i];
+              py::array_t<double> data(idx_buffer.size);
+              auto data_buffer = data.request();
+              double *data_ptr = static_cast<double *>(data_buffer.ptr);
+              std::fill(data_ptr, data_ptr + data_buffer.size, fill_value);
+
+              for (size_t i = 0; i < idx_buffer.size; i++) {
```
ajaust: ssize_t vs size_t

lib/resdata/rd_grid_pybind.cpp
```diff
@@ -376,19 +376,25 @@ PYBIND11_MODULE(_grid, m) {
...
-                output[i] = input[di];
-        }
-    });
+              py::array_t<int32_t> data(idx_buffer.size);
+              auto data_buffer = data.request();
+              int32_t *data_ptr = static_cast<int32_t *>(data_buffer.ptr);
+              std::fill(data_ptr, data_ptr + data_buffer.size, fill_value);
+
+              int *input = rd_kw_get_int_ptr(rd_kw);
+              for (size_t i = 0; i < idx_buffer.size; i++) {
```
ajaust: ssize_t vs size_t

lib/resdata/rd_grid_pybind.cpp
```diff
@@ -1,4 +1,5 @@
 #include <tuple>
+#include <algorithm>
```
ajaust, line 2: I guess this will be squashed? Otherwise it should be moved to the commit where fill is being introduced.

lib/resdata/rd_grid_pybind.cpp
```diff
@@ -390,9 +378,16 @@ PYBIND11_MODULE(_grid, m) {
...
     m.def("_export_data_as_int", [](ptrdiff_t size, py::array_t<int32_t> indx,
                                     py::handle kw, py::array_t<int32_t> data) {
-        rd_grid_export_data_as_int(size, indx.mutable_data(),
-                                   from_cwrap<rd_kw_type>(kw),
-                                   data.mutable_data());
+        auto rd_kw = from_cwrap<rd_kw_type>(kw);
+        int32_t *idx_ptr = indx.mutable_data();
+        int32_t *data_ptr = data.mutable_data();
+
+        int *input = rd_kw_get_int_ptr(kw);
```
ajaust: Only relevant if we do **not** squash.

This commit does not compile with the following error

```text
/Users/AEJ/projects/github/resdata/lib/resdata/rd_grid_pybind.cpp:385:22: fatal error: no matching function for call to 'rd_kw_get_int_ptr'
  385 |         int *input = rd_kw_get_int_ptr(kw);
      |                      ^~~~~~~~~~~~~~~~~
/Users/AEJ/projects/github/resdata/lib/include/resdata/rd_kw.hpp:201:1: note: candidate function not viable: no known conversion from 'py::handle' to 'const rd_kw_type *'
      (aka 'const rd_kw_struct *') for 1st argument
  201 | RD_KW_GET_TYPED_PTR_HEADER(int);
      | ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/AEJ/projects/github/resdata/lib/include/resdata/rd_kw.hpp:198:11: note: expanded from macro 'RD_KW_GET_TYPED_PTR_HEADER'
  198 |     type *rd_kw_get_##type##_ptr(const rd_kw_type *)
      |           ^                      ~~~~~~~~~~~~~~~~~~
<scratch space>:170:1: note: expanded from here
  170 | rd_kw_get_int_ptr
      | ^
```

All other commits compile fine.

ajaust (commented):
Overall good. I just have some minor comments w.r.t. compiler warnings (signed vs unsigned type) and some issues that only matter if we **not** squash the commits.

lib/resdata/rd_grid_pybind.cpp
eivindjahren: ssize_t would be preferred, but it is not available on windows. I would prefer to not cast, so are you ok with using ptrdiff_t?

lib/resdata/rd_grid_pybind.cpp
eivindjahren: Yes, lets squash.

lib/resdata/rd_grid_pybind.cpp
eivindjahren: Lets squash then :)

lib/resdata/rd_grid_pybind.cpp
ajaust: I checked and there is a `py::ssize_t` provided by pybind11. Give it a try and see if this works on all platforms.

ajaust (approved):

PR #1153, opened 2026-05-06
Fix format specifiers in test_util.cpp

eivindjahren:


ajaust (approved):
Looks good. 

I think the the findings of Copilot are noteworthy, especially if Windows support is relevant.

ajaust (approved):

PR #1157, opened 2026-05-07
Fix all instances of multiplication before convert to larger type

eivindjahren:


lib/resdata/rd_kw.cpp
```diff
@@ -392,8 +398,12 @@ static bool rd_kw_header_eq(const rd_kw_type *rd_kw1,
 
 static bool rd_kw_data_equal__(const rd_kw_type *rd_kw, const void *data,
                                int cmp_elements) {
+    if (cmp_elements < 0)
+        util_abort("%s : cmp_elements was negative: %d\n", __func__,
+                   rd_kw->size);
```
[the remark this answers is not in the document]
eivindjahren: Fixd

lib/resdata/layer.cpp
```diff
@@ -394,8 +397,10 @@ class BlockTracer {
     BlockTracer(layer_type *layer, bool erase, int value,
                 int_vector_type *i_list, int_vector_type *j_list)
         : layer(layer), erase(erase), value(value),
-          visited((layer->nx + 1) * (layer->ny + 1), false), i_list(i_list),
-          j_list(j_list) {}
+          visited(static_cast<size_t>(layer->nx + 1) *
+                      static_cast<size_t>(layer->ny + 1),
```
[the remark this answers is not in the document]
eivindjahren: Fixd

lib/resdata/rd_kw.cpp
```diff
@@ -2513,7 +2537,11 @@ static bool rd_kw_elm_equal_numeric__(const rd_kw_type *rd_kw1,
 
 static bool rd_kw_elm_equal__(const rd_kw_type *rd_kw1,
                               const rd_kw_type *rd_kw2, int offset) {
-    size_t data_offset = rd_type_get_sizeof_ctype(rd_kw1->data_type) * offset;
+    if (offset < 0)
+        util_abort("%s : offset was negative: %d\n", __func__, offset);
+    size_t data_offset =
+        static_cast<size_t>(rd_type_get_sizeof_ctype(rd_kw1->data_type)) *
+        offset;
```
ajaust: We could make this more consistent with the pattern used in other methods:
```suggestion
        static_cast<size_t>(offset)) *
        rd_type_get_sizeof_ctype(rd_kw1->data_type;
```

lib/resdata/layer.cpp
```diff
@@ -394,8 +397,10 @@ class BlockTracer {
     BlockTracer(layer_type *layer, bool erase, int value,
                 int_vector_type *i_list, int_vector_type *j_list)
         : layer(layer), erase(erase), value(value),
-          visited((layer->nx + 1) * (layer->ny + 1), false), i_list(i_list),
-          j_list(j_list) {}
+          visited((static_cast<size_t>(layer->nx) + 1) *
+                      (static_cast<size_t>(layer->ny) + 1),
```
ajaust, line 401: We could check if `nx` and `ny` are negative here, but not sure if this is needed. `layer_alloc` does this check and is called before a `BlockTracer` is created.

ajaust (approved):
Only two minor things.

lib/resdata/layer.cpp
eivindjahren: Yes, I think because it is allocated with non-negative now I feel fine with doing it this way.

PR #1160, opened 2026-05-08
Add more ruff rules

berland:
**Issue**
Tightens up ruff setup

**Approach**
Home-made script for detecting which ruff rules are already passing.


eivindjahren (approved):
🇹🇩

eivindjahren:
Could also probably  autofix I001 with (almost) no problems.

berland:
> Could also probably autofix I001 with (almost) no problems.

mañana mañana 😄

PR #1161, opened 2026-05-08
Ruff furb

berland:
**Issue**
Resolves ruff exceptions in pyproject.toml

**Approach**
`ruff check --select <rule-id> --fix --unsafe-fixes`


andreas-el (approved):
📦

PR #1162, opened 2026-05-11
Solve and enforce ruff UP rules

berland:
All these fixes have been done automatically using `ruff --fix`


python/resdata/grid/rd_grid_generator.py
```diff
@@ -716,7 +716,7 @@ def assert_ijk_bounds(cls, dims, ijk_bounds):
             if not (isinstance(bound[0], int) and isinstance(bound[1], int)):
                 raise TypeError(
                     "Expected bound to consist of two integers, ",
-                    "was %s (%s)" % (str(bound), str((map(type, bound)))),
+                    "was %s (%s)" % (str(bound), str(map(type, bound))),
```
eivindjahren: This seems like a bug. Its not the problem of the style change being made here, but `map(type, bound)` doesn't return a list but an iterator so the TypeError message becomes useless. Should probably just add a quick `pytest.raises` test for this and change it to `", ".join(map(type, bound))`.

python/resdata/summary/rd_smspec_node.py
```diff
@@ -34,7 +34,7 @@ class ResdataSMSPECNode(BaseCClass):
     _get_default = ResdataPrototype("float smspec_node_get_default(rd_smspec_node)")
 
     def __init__(self):
-        super(ResdataSMSPECNode, self).__init__(0)  # null pointer
+        super().__init__(0)  # null pointer
```
eivindjahren: This also seems suspicious. Calling this constructor could throw in `__init__` rather than raising the `NotImplementedError`.

eivindjahren (approved):
This seems fine, although some touched lines seem to have easily fixed unrelated problems that we probably should address now to avoid making the situation worse.

eivindjahren (approved):

PR #1163, opened 2026-05-11
Make ResdataKW.read_grdecl use pybind

eivindjahren:
Also fixes a segfault for zero-length min_max.

With regards to returning None or not, I changed the documentation of "If the keyword can not be found the method will return None." This is because we consider this legacy code so its more likely that someone depends on behavior rather than documentation.

### DIFF
--- a/lib/include/resdata/rd_kw_grdecl.hpp
+++ b/lib/include/resdata/rd_kw_grdecl.hpp
@@ -8,11 +8,9 @@
 
 bool rd_kw_grdecl_fseek_kw(const char *, bool, FILE *);
 
-rd_kw_type *rd_kw_fscanf_alloc_grdecl_dynamic(FILE *stream, const char *kw,
-                                              bool strict, rd_data_type);
-
-rd_kw_type *rd_kw_fscanf_alloc_grdecl(FILE *stream, const char *kw, int size,
-                                      rd_data_type data_type);
+rd_kw_type *rd_kw_fscanf_alloc_grdecl(FILE *stream, const char *kw,
+                                      rd_data_type data_type, int size = 0,
+                                      bool strict = true);
 
 void rd_kw_fprintf_grdecl(const rd_kw_type *rd_kw, FILE *stream,
                           const char *special_header = nullptr);
--- a/lib/include/resdata/rd_kw_grdecl.hpp
+++ b/lib/include/resdata/rd_kw_grdecl.hpp
@@ -18,9 +18,6 @@ rd_kw_type *rd_kw_fscanf_alloc_grdecl_dynamic(FILE *stream, const char *kw,
 rd_kw_type *rd_kw_fscanf_alloc_grdecl(FILE *stream, const char *kw, int size,
                                       rd_data_type data_type);
 
-rd_kw_type *rd_kw_fscanf_alloc_current_grdecl(FILE *stream,
-                                              rd_data_type data_type);
-
 void rd_kw_fprintf_grdecl(const rd_kw_type *rd_kw, FILE *stream);
 void rd_kw_fprintf_grdecl__(const rd_kw_type *rd_kw, const char *special_header,
                             FILE *stream);
--- a/lib/include/resdata/rd_kw_grdecl.hpp
+++ b/lib/include/resdata/rd_kw_grdecl.hpp
@@ -1,14 +1,8 @@
-/*
- This header does not define datatypes; just a couple of functions. It should
- be included from the rd_kw.h header, so applications do not need to include this
- header explicitly.
-*/
+#pragma once
 
-#ifndef ERT_RD_KW_GRDECL_H
-#define ERT_RD_KW_GRDECL_H
-#ifdef __cplusplus
-extern "C" {
-#endif
+#include <cstdio>
+
+#include <resdata/rd_kw.hpp>
 
 bool rd_kw_grdecl_fseek_kw(const char *, bool, FILE *);
 
@@ -21,8 +15,3 @@ rd_kw_type *rd_kw_fscanf_alloc_grdecl(FILE *stream, const char *kw, int size,
 void rd_kw_fprintf_grdecl(const rd_kw_type *rd_kw, FILE *stream);
 void rd_kw_fprintf_grdecl__(const rd_kw_type *rd_kw, const char *special_header,
                             FILE *stream);
-
-#ifdef __cplusplus
-}
-#endif
-#endif
--- a/lib/include/resdata/rd_kw_grdecl.hpp
+++ b/lib/include/resdata/rd_kw_grdecl.hpp
@@ -12,10 +12,8 @@ extern "C" {
 
 bool rd_kw_grdecl_fseek_kw(const char *, bool, FILE *);
 
-rd_kw_type *rd_kw_fscanf_alloc_grdecl_dynamic__(FILE *stream, const char *kw,
-                                                bool strict, rd_data_type);
 rd_kw_type *rd_kw_fscanf_alloc_grdecl_dynamic(FILE *stream, const char *kw,
-                                              rd_data_type);
+                                              bool strict, rd_data_type);
 
 rd_kw_type *rd_kw_fscanf_alloc_grdecl(FILE *stream, const char *kw, int size,
                                       rd_data_type data_type);
--- a/lib/include/resdata/rd_kw_grdecl.hpp
+++ b/lib/include/resdata/rd_kw_grdecl.hpp
@@ -1,6 +1,8 @@
 #pragma once
 
 #include <cstdio>
+#include <optional>
+#include <string>
 
 #include <resdata/rd_kw.hpp>
 
@@ -12,6 +14,5 @@ rd_kw_type *rd_kw_fscanf_alloc_grdecl_dynamic(FILE *stream, const char *kw,
 rd_kw_type *rd_kw_fscanf_alloc_grdecl(FILE *stream, const char *kw, int size,
                                       rd_data_type data_type);
 
-void rd_kw_fprintf_grdecl(const rd_kw_type *rd_kw, FILE *stream);
-void rd_kw_fprintf_grdecl__(const rd_kw_type *rd_kw, const char *special_header,
-                            FILE *stream);
+void rd_kw_fprintf_grdecl(const rd_kw_type *rd_kw, FILE *stream,
+                          const char *special_header = nullptr);
--- a/lib/resdata/rd_kw.cpp
+++ b/lib/resdata/rd_kw.cpp
@@ -2360,6 +2360,9 @@ void rd_kw_max_min(const rd_kw_type *rd_kw, void *_max, void *_min) {
 #define RD_KW_MAX_MIN(ctype)                                                   \
     void rd_kw_max_min_##ctype(const rd_kw_type *rd_kw, ctype *_max,           \
                                ctype *_min) {                                  \
+        if (rd_kw->size < 1)                                                   \
+            util_abort("%s: size of zero length array is undefined \n",        \
+                       __func__);                                              \
         KW_MAX_MIN(ctype);                                                     \
     }
 
--- a/lib/resdata/rd_kw_grdecl.cpp
+++ b/lib/resdata/rd_kw_grdecl.cpp
@@ -452,16 +452,10 @@ static rd_kw_type *__rd_kw_fscanf_alloc_grdecl__(FILE *stream,
    the whole keyword is loaded, and then return.
 */
 
-rd_kw_type *rd_kw_fscanf_alloc_grdecl_dynamic__(FILE *stream, const char *kw,
-                                                bool strict,
-                                                rd_data_type data_type) {
-    return __rd_kw_fscanf_alloc_grdecl__(stream, kw, strict, 0, data_type);
-}
-
 rd_kw_type *rd_kw_fscanf_alloc_grdecl_dynamic(FILE *stream, const char *kw,
+                                              bool strict,
                                               rd_data_type data_type) {
-    bool strict = true;
-    return rd_kw_fscanf_alloc_grdecl_dynamic__(stream, kw, strict, data_type);
+    return __rd_kw_fscanf_alloc_grdecl__(stream, kw, strict, 0, data_type);
 }
 
 /*
--- a/lib/resdata/rd_kw_grdecl.cpp
+++ b/lib/resdata/rd_kw_grdecl.cpp
@@ -231,9 +231,9 @@ void checked_realloc(std::unique_ptr<T[], void (*)(void *)> &ptr,
 
    Observe that no-spaces-are-allowed-around-the-*
 */
-static char *fscanf_alloc_grdecl_data(const char *header, bool strict,
-                                      rd_data_type data_type, int *kw_size,
-                                      FILE *stream) {
+static std::unique_ptr<char[], void (*)(void *)>
+fscanf_alloc_grdecl_data(const char *header, bool strict,
+                         rd_data_type data_type, int *kw_size, FILE *stream) {
     char newline = '\n';
     bool atEOF = false;
     size_t init_size = 32;
@@ -355,7 +355,7 @@ static char *fscanf_alloc_grdecl_data(const char *header, bool strict,
     }
     *kw_size = data_index;
     checked_realloc<char>(data, sizeof_ctype * data_index * sizeof *data.get());
-    return data.release();
+    return data;
 }
 
 /*
@@ -420,13 +420,12 @@ rd_kw_type *rd_kw_fscanf_alloc_grdecl(FILE *stream, const char *kw,
         char file_header[MAX_GRDECL_HEADER_SIZE] = {0};
         if (fscanf(stream, MAX_GRDECL_HEADER_SCANF_FMT, file_header) == 1) {
             int kw_size;
-            char *data = fscanf_alloc_grdecl_data(file_header, strict,
-                                                  data_type, &kw_size, stream);
+            auto data = fscanf_alloc_grdecl_data(file_header, strict, data_type,
+                                                 &kw_size, stream);
 
             // Verify size
             if (size > 0)
                 if (size != kw_size) {
-                    free(data);
                     throw std::invalid_argument(
                         fmt::format("size mismatch when loading:{}. File:{} "
                                     "elements. Requested:{} elements",
@@ -436,7 +435,7 @@ rd_kw_type *rd_kw_fscanf_alloc_grdecl(FILE *stream, const char *kw,
             {
                 rd_kw_type *rd_kw =
                     rd_kw_alloc_new(file_header, kw_size, data_type, NULL);
-                rd_kw_set_data_ptr(rd_kw, data);
+                rd_kw_set_data_ptr(rd_kw, data.release());
                 return rd_kw;
             }
 
--- a/lib/resdata/rd_kw_grdecl.cpp
+++ b/lib/resdata/rd_kw_grdecl.cpp
@@ -175,6 +175,33 @@ static void iset_range(char *data, int data_index, int sizeof_ctype,
     }
 }
 
+template <typename T>
+std::unique_ptr<T[], void (*)(void *)> checked_calloc(size_t num) {
+    T *ptr = static_cast<T *>(std::calloc(num, sizeof(T)));
+
+    if (ptr == nullptr) {
+        throw std::bad_alloc{};
+    }
+    return std::unique_ptr<T[], void (*)(void *)>(
+        ptr, [](void *p) { std::free(p); });
+}
+
+template <typename T>
+void checked_realloc(std::unique_ptr<T[], void (*)(void *)> &ptr,
+                     size_t new_element_count) {
+    if (new_element_count == 0) {
+        ptr.reset();
+        return;
+    }
+    T *raw_ptr = ptr.release();
+    void *new_raw_ptr = std::realloc(raw_ptr, new_element_count * sizeof(T));
+    if (new_raw_ptr == nullptr) {
+        ptr.reset(raw_ptr);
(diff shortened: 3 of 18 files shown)


python/resdata/resfile/rd_kw.py
```diff
@@ -953,6 +953,8 @@ def get_min_max(self):
         Will raise TypeError exception if the keyword is not of
         numerical type.
         """
+        if len(self) == 0:
+            return (np.nan, np.nan)
```
ajaust, line 959: I am wondering if we would want to throw some exception instead of creating an invalid min-max pair.

lib/resdata/rd_kw.cpp
```diff
@@ -2360,6 +2360,9 @@ void rd_kw_max_min(const rd_kw_type *rd_kw, void *_max, void *_min) {
 #define RD_KW_MAX_MIN(ctype)                                                   \
     void rd_kw_max_min_##ctype(const rd_kw_type *rd_kw, ctype *_max,           \
                                ctype *_min) {                                  \
+        if (rd_kw->size < 1)                                                   \
+            util_abort("%s: size of zero length array is undefined \n",        \
+                       __func__);                                              \
```
ajaust, line 2365: It seems like we want to get rid of `util_abort` so adding a new `util_abort` here is surprising. Should/Could we replace it by throwing an exception?

tests/rd_tests/test_rd_kw.py
```diff
@@ -853,6 +859,13 @@ def nice_from_dtype(dtype):
 """)
 
 
+def order_of(values):
+    if len(values):
+        return np.max(np.abs(values))
```
ajaust: This is the infinity/max norm. It might be clearer to call the corresponding method
```suggestion
        return np.linalg.norm(values, ord=np.inf)
```

tests/rd_tests/test_rd_kw.py
```diff
@@ -1130,15 +1146,14 @@ def add_squared(self, kw1, kw2):
         akw1.add_squared(akw2)
         model_values1 += model_values2 * model_values2
 
-        size_order = max(
-            1.0, np.max(np.abs(model_values2)), np.max(np.abs(model_values1))
-        )
+        size_order = max(1.0, order_of(model_values2), order_of(model_values1))
```
ajaust, line 1166: What is the `size_order` of model values?

tests/rd_tests/test_rd_kw.py
```diff
@@ -853,6 +859,13 @@ def nice_from_dtype(dtype):
 """)
 
 
+def order_of(values):
+    if len(values):
+        return np.max(np.abs(values))
+    else:
+        return 1.0
```
ajaust: Shouldn't and empty vector have max norm `0`?  `np.linalg.norm(values, ord=np.inf)` would automatically handle this. 

All locations where `order_of` is being used appear to wrap the return value into a `max(1, order_of(...), ...)` so it should be save to remove the `if`-`else` clause.

tests/rd_tests/test_rd_kw.py
```diff
@@ -1243,7 +1260,7 @@ def isub(self, kw1, kw2):
         akw1 -= akw2
         model_values1 -= model_values2
 
-        size_order = max(1.0, np.max(model_values1), np.max(abs(model_values2)))
+        size_order = max(1.0, order_of(model_values1), order_of(model_values2))
```
ajaust, line 1280: This changes `np.max(model_values1)` to something like `np.max(np.abs(model_values1))`, an expression that uses the absolute value. It therefore chages the notion of this test/value. Is this intended?

lib/include/resdata/rd_kw_grdecl.hpp
```diff
@@ -12,10 +12,8 @@ extern "C" {
 
 bool rd_kw_grdecl_fseek_kw(const char *, bool, FILE *);
 
-rd_kw_type *rd_kw_fscanf_alloc_grdecl_dynamic__(FILE *stream, const char *kw,
-                                                bool strict, rd_data_type);
```
ajaust, line 16: The commit message says that `rd_kw_fscanf_alloc_grdecl_dynamic` is removed, but we keep `rd_kw_fscanf_alloc_grdecl_dynamic`. Instead, `rd_kw_fscanf_alloc_grdecl_dynamic__` is removed.

lib/resdata/rd_kw_grdecl.cpp
```diff
@@ -483,26 +483,6 @@ rd_kw_type *rd_kw_fscanf_alloc_grdecl(FILE *stream, const char *kw, int size,
...
-   as input, and has virtually zero possibilities to check what it is
-   doing. The function should only be used when you are certain
-   that the input file is well formatted.
-*/
-
-static rd_kw_type *rd_kw_fscanf_alloc_current_grdecl__(FILE *stream,
-                                                       bool strict,
-                                                       rd_data_type data_type) {
-    return __rd_kw_fscanf_alloc_grdecl__(stream, NULL, strict, 0, data_type);
-}
```
ajaust, line 504: Commit message mentions that `rd_kw_fscanf_alloc_current_grdecl` is removed, but both `rd_kw_fscanf_alloc_current_grdecl` and `rd_kw_fscanf_alloc_current_grdecl__` are being removed.

lib/resdata/rd_kw_pybind.cpp
```diff
@@ -44,5 +44,10 @@ PYBIND11_MODULE(_kw, m) {
                                                       *rd_data_type));
         },
         py::return_value_policy::reference);
+    m.def("_fprintf_grdecl", [](py::handle self, py::handle file) {
```
ajaust, line 47: Typo in commit message: `_pfrintf_grdecl -> `_fprintf_grdecl`

lib/resdata/rd_kw_grdecl.cpp
```diff
@@ -175,6 +175,33 @@ static void iset_range(char *data, int data_index, int sizeof_ctype,
...
+    if (ptr == nullptr) {
+        throw std::bad_alloc{};
+    }
+    return std::unique_ptr<T[], void (*)(void *)>(
+        ptr, [](void *p) { std::free(p); });
+}
+
+template <typename T>
+void checked_realloc(std::unique_ptr<T[], void (*)(void *)> &ptr,
+                     size_t new_element_count) {
```
ajaust, line 177: Is this the element or byte count? The method is called as `checked_realloc<char>(data, byte_size);` and with `sizeof *data.get()`.

lib/resdata/rd_kw_grdecl.cpp
```diff
@@ -175,6 +175,33 @@ static void iset_range(char *data, int data_index, int sizeof_ctype,
...
+    if (new_element_count == 0) {
+        ptr.reset();
+        return;
+    }
+    T *raw_ptr = ptr.release();
+    void *new_raw_ptr = std::realloc(raw_ptr, new_element_count * sizeof(T));
```
ajaust, line 183: Depending on whether the element count should be the element count or the byte count, see comment above, this will break depending on the type of `T`. At the moment, `T=char` so this is fine `sizeof(char) == 1`, but `sizeof(T) > 1` could be an issue.

lib/include/resdata/rd_kw_grdecl.hpp
```diff
@@ -12,6 +14,6 @@ rd_kw_type *rd_kw_fscanf_alloc_grdecl_dynamic(FILE *stream, const char *kw,
 rd_kw_type *rd_kw_fscanf_alloc_grdecl(FILE *stream, const char *kw, int size,
                                       rd_data_type data_type);
 
-void rd_kw_fprintf_grdecl(const rd_kw_type *rd_kw, FILE *stream);
-void rd_kw_fprintf_grdecl__(const rd_kw_type *rd_kw, const char *special_header,
-                            FILE *stream);
+void rd_kw_fprintf_grdecl(
+    const rd_kw_type *rd_kw, FILE *stream,
+    std::optional<std::string> special_header = std::nullopt);
```
ajaust: Nothing wrong with the commit itself, but a bit of nitpicking:
As much as I like the C++ style and `std::optional`, I am not sure if using the `std::optional` in the best way since we cannot consistently use the `std::optional`. We still need to convert between std::optional and pointers after the change.


## Use C++ default argument for a pointer type

- Swap order of arguments the way you did such that `special_header` is the last one, but keep it as a pointer and make it `nullptr` by default
```suggestion
    const char *special_header = nullptr);
```
In this case we do not need to include `string` and `optional`. We can also get rid of the new `if-else` constructions further below since 

## Convert methods to consistently use `std::optional`

Change `rd_grid_grdecl_fprintf_kw` to accept `std::optional` instead of a pointer for `special_header`. This allows to pass through the optional to `rd_kw_fprintf_grdecl` as

```c++
rd_kw_fprintf_grdecl(rd_kw, stream, special_header); 
```

Would avoid branching, but still some more changes in the tests and Python bindings, I think.

## Add some helper to convert pointers to `std::optional<string>`

Some kind of in between option.

Define some helper like this

```c++
static std::optional<std::string> to_optional(const char *s) {
   return s ? std::optional<std::string>(s) : std::nullopt;
}
```

and then replace `if-else` constructs by something like

```c++
rd_kw_fprintf_grdecl(rd_kw, stream, to_optional(special_header));
```

This will create some `std::strings` and still has internal branching.

lib/resdata/rd_kw_grdecl.cpp
```diff
@@ -291,13 +319,14 @@ static char *fscanf_alloc_grdecl_data(const char *header, bool strict,
...
-                            size_t byte_size = sizeof_ctype * sizeof *data;
+                            size_t byte_size =
+                                sizeof_ctype * sizeof *data.get();
 
                             data_size = util_size_t_min(
                                 RD_KW_MAX_SIZE, 2 * (data_index + multiplier));
                             byte_size *= data_size;
 
-                            data = (char *)util_realloc(data, byte_size);
+                            checked_realloc<char>(data, byte_size);
```
ajaust: Checked realloc will allocate space for `new_element_count` of the given type. Therefore, we need to pass the element count instead of the `byte_size`. We use `char` here so this issue does not lead to any leaks, but I think the current usage is dangerous and wrong.

lib/resdata/rd_kw_grdecl.cpp
```diff
@@ -317,10 +346,9 @@ static char *fscanf_alloc_grdecl_data(const char *header, bool strict,
         } else
             break;
     }
-    free(buffer);
     *kw_size = data_index;
-    data = (char *)util_realloc(data, sizeof_ctype * data_index * sizeof *data);
-    return data;
+    checked_realloc<char>(data, sizeof_ctype * data_index * sizeof *data.get());
```
ajaust: Same as above. This passes the byte size to allocate instead of the element count.

lib/resdata/rd_kw_grdecl.cpp
```diff
@@ -175,6 +175,33 @@ static void iset_range(char *data, int data_index, int sizeof_ctype,
...
+        return;
+    }
+    T *raw_ptr = ptr.release();
+    void *new_raw_ptr = std::realloc(raw_ptr, new_element_count * sizeof(T));
+    if (new_raw_ptr == nullptr) {
+        ptr.reset(raw_ptr);
+        throw std::bad_alloc{};
+    }
+    ptr.reset(static_cast<T *>(new_raw_ptr));
+}
```
ajaust, line 189: This method is often called in confusing way passing the byte count instead of the element cound. The call sites all use `T = char` so there is no overallocation, but it is confusing. 

We should with
- Fix the call sites
- Update the method to expect the byte count instead of the element count.

lib/resdata/rd_kw_grdecl.cpp
```diff
@@ -213,12 +240,12 @@ static char *fscanf_alloc_grdecl_data(const char *header, bool strict,
     size_t data_index = 0;
     int sizeof_ctype = rd_type_get_sizeof_ctype(data_type);
     size_t data_size = init_size;
-    char *buffer = (char *)util_calloc((buffer_size + 1), sizeof *buffer);
-    char *data = (char *)util_calloc(sizeof_ctype * data_size, sizeof *data);
+    auto buffer = checked_calloc<char>(buffer_size + 1);
+    auto data = checked_calloc<char>(sizeof_ctype * data_size);
```
ajaust: Passes byte counts (?) instead of element counts. See comment below.

ajaust (commented):
I have mostly minor points. I have some concerns how `checked_realloc` is defined vs. how it is used (using element count vs byte size for allocation).

I don't know if all my comments appear at the correct location nor whether they are still all relevent. Some commits changed while I was preparing the review. Therefore, there might be the some comment duplications.

python/resdata/resfile/rd_kw.py
eivindjahren: I agree, but unfortunately, the min-max pair is used in the `__repr__` function so we should probably give some number. Alternatively, I could just do a check in `__repr__`.

lib/resdata/rd_kw.cpp
eivindjahren: Since this condition eventually ends up in an abort and we cannot yet replace it with an exception that does not leak and can be caught on the python side, I just wanted to stay consistent with rest.

lib/include/resdata/rd_kw_grdecl.hpp
eivindjahren: The idea of the using std::optional because it was in the path of the pybind interface, but there seems to have been no reason for it. I changed it to nullptr.

lib/resdata/rd_kw_grdecl.cpp
eivindjahren: So yes, this is wrong unfortunately. I will fix it to be correct here, but this will be a separate commit as we are changing the behavior to cast to char * when we pass to the rd_kw constructor.

tests/rd_tests/test_rd_kw.py
eivindjahren: Changed to linalg.norm :)

tests/rd_tests/test_rd_kw.py
eivindjahren: Changed to linalg.norm :)

tests/rd_tests/test_rd_kw.py
eivindjahren: The `size_order` is the magnitude of the input values for the calculation. I thought it would make sense to use the model values for it to avoid using the system under test for defining the criteria.

lib/resdata/rd_kw_grdecl.cpp
eivindjahren: I changed it so that that we use check_realloc<T> for whatever the type we are reading is and then only later cast to char*.

lib/resdata/rd_kw_grdecl.cpp
eivindjahren: I think I fixed the callsites :)

lib/resdata/rd_kw_grdecl.cpp
eivindjahren: Should be fixed now.

tests/rd_tests/test_rd_kw.py
eivindjahren: Yes, this is intentional, it was incorrect as it was.

lib/resdata/rd_kw_grdecl.cpp
eivindjahren: Added to body of message.

lib/include/resdata/rd_kw_grdecl.hpp
eivindjahren: Fixed

lib/resdata/rd_kw_pybind.cpp
eivindjahren: Fixed

lib/resdata/rd_kw_grdecl.cpp
eivindjahren: Fixed by changing the callsite.

tests/rd_tests/test_rd_kw.py
```diff
@@ -880,10 +880,7 @@ def nice_from_dtype(dtype):
 
 
 def order_of(values):
-    if len(values):
-        return np.max(np.abs(values))
-    else:
-        return 1.0
+    return np.linalg.norm(values, ord=np.inf)
```
ajaust, line 883: Just to be super sure that this is what we want. This will return different values than before

### Now

- `order_of([0])` returns `0` 
- `order_of([])` returns `0`.
 
### Before 
- `order_of([0])` returns `0` 
- `order_of([])` returns `1`.

tests/rd_tests/test_rd_kw.py
ajaust: Thanks for the explanation. :)

lib/resdata/rd_kw_grdecl.cpp
```diff
@@ -420,8 +374,22 @@ rd_kw_type *rd_kw_fscanf_alloc_grdecl(FILE *stream, const char *kw,
...
+                                                            kw_size, stream)
+                                        .release()));
+            } else if (rd_type_is_float(data_type)) {
+                data.reset((char *)(fscanf_grdecl_data<float>(
+                                        file_header, strict, kw_size, stream)
+                                        .release()));
+            } else if (rd_type_is_double(data_type)) {
+                data.reset((char *)(fscanf_grdecl_data<double>(
+                                        file_header, strict, kw_size, stream)
+                                        .release()));
```
ajaust, line 393: Optional:

We could replace the `(char *)` casts with `reinterprete_cast` if we want to stay closer to the C++ style.

lib/resdata/rd_kw_grdecl.cpp
```diff
@@ -202,6 +187,27 @@ void checked_realloc(std::unique_ptr<T[], void (*)(void *)> &ptr,
     }
     ptr.reset(static_cast<T *>(new_raw_ptr));
 }
+template <typename T> const inline std::string value_format();
```
ajaust: Nitpick:
Empty line missing above definition of `value_format`.

lib/resdata/rd_kw_grdecl.cpp
```diff
@@ -202,6 +187,27 @@ void checked_realloc(std::unique_ptr<T[], void (*)(void *)> &ptr,
...
+template <> const inline std::string value_format<int>() { return "%d"; }
+template <> const inline std::string value_format<float>() { return "%128g"; }
+template <> const inline std::string value_format<double>() { return "%128lg"; }
+
+template <typename T> const inline std::string multiplier_format() {
+    return "%d*" + value_format<T>();
+}
```
ajaust: Optional:

There might be a performance optimisation here depending on what the compiler does. `multiplier_format` may create a new `std::string` whenever it is called by `scan_data`. `scan_data` is called in a loop so we might call `multiplier_format` often. We could avoid this by using more compile time constants, e.g.,

```suggestion
template <typename T> constexpr const char *value_format();
template <> constexpr const char *value_format<int>() { return "%d"; }
template <> constexpr const char *value_format<float>() { return "%128g"; }
template <> constexpr const char *value_format<double>() { return "%128lg"; }
 
template <typename T> constexpr const char *multiplier_format();
template <> constexpr const char *multiplier_format<int>() { return "%d*%d"; }
template <> constexpr const char *multiplier_format<float>() { return "%d*%128g"; }
template <> constexpr const char *multiplier_format<double>() { return "%d*%128lg"; }
```

`std::string` isn't `constexpr` compatible in C++17 yet. Thus, the `const char *`.

lib/resdata/rd_kw_grdecl.cpp
```diff
@@ -202,6 +187,27 @@ void checked_realloc(std::unique_ptr<T[], void (*)(void *)> &ptr,
...
+                           int &multiplier) {
+    T value;
+    if (sscanf(buffer.get(), multiplier_format<T>().c_str(), &multiplier,
+               &value) == 2)
+        return value;
+    multiplier = 1;
+    if (sscanf(buffer.get(), value_format<T>().c_str(), &value) == 1)
+        return value;
+    return std::nullopt;
+}
```
ajaust, line 214: I like this change and how the code looks afterwards.

lib/resdata/rd_kw_grdecl.cpp
```diff
@@ -420,8 +374,22 @@ rd_kw_type *rd_kw_fscanf_alloc_grdecl(FILE *stream, const char *kw,
         char file_header[MAX_GRDECL_HEADER_SIZE] = {0};
         if (fscanf(stream, MAX_GRDECL_HEADER_SCANF_FMT, file_header) == 1) {
             int kw_size;
-            auto data = fscanf_grdecl_data(file_header, strict, data_type,
-                                           &kw_size, stream);
+            std::unique_ptr<char, void (*)(void *)> data{nullptr, &std::free};
```
ajaust: We could use

```suggestion
            std::unique_ptr<char[], void (*)(void *)> data{nullptr, &std::free};
```

if we want to express that `data` may contain more than one char/byte.

ajaust (approved):
Looks good overall. One verification questions and otherwise nitpicks. Check if you want to address these, but feel free to ignore them. :)

tests/rd_tests/test_rd_kw.py
eivindjahren: This is fine :)

lib/resdata/rd_kw_grdecl.cpp
eivindjahren: True, I just kept the c style for now.

lib/resdata/rd_kw_grdecl.cpp
eivindjahren: Fixed

PR #1165, opened 2026-05-12
Use pre-commit in CI, check whitespace and use actionlint

berland:
**Issue**
Resolves trailing whitespaces, one bug in testing.yml

**Approach**
`pre-commit`


andreas-el (approved):
🔨 🔨 🔨

PR #1168, opened 2026-05-18
Clean up rd_alloc_filename

eivindjahren:
After the removal of the c++ RFT code there was a cpp version of rd_alloc_filename that was unused. Also the c style rd_alloc_filename was used clunkily after changing rd_grid.cpp to c++ style. This PR cleans this up by changing all callers of rd_alloc_filename and rd_alloc_exfilename to rd::filename.


lib/resdata/rd_sum.cpp
```diff
@@ -24,6 +25,8 @@
 
 #include "detail/util/path.hpp"
 
+namespace fs = std::filesystem;
```
ajaust, line 31: `filesystem` is included and gets a convenient namespace name, but I don't see it being used anywhere.

Maybe it will be used in a later commits.

lib/tests/test_rd_sum.cpp
```diff
@@ -0,0 +1,670 @@
...
+ * Tests for rd_sum that exercise the public summary API.
+ *
+ * Summary files are produced on-the-fly with rd_sum_alloc_writer and
+ * rd_sum_fwrite, and read back with the read_summary RAII utility. Resources
+ * are owned by RAII wrappers (rd_sum_ptr, rd_grid_ptr, rd_file_ptr, rd_kw_ptr,
+ * ERT::FortIO) rather than being freed by hand.
+ */
+
+#include <catch2/catch.hpp>
+#include <cctype>
```
ajaust: I am not sure if this header is actually being used. I haven't found any character conversion yet.

lib/tests/test_rd_sum.cpp
```diff
@@ -0,0 +1,670 @@
...
+#include <memory>
+#include <string>
+#include <vector>
+
+#include <ert/util/stringlist.hpp>
+#include <ert/util/time_t_vector.hpp>
+#include <ert/util/util.hpp>
+
+#include <resdata/rd_file.hpp>
+#include <resdata/rd_grid.hpp>
```
ajaust: Is this header needed? I did not find any `rd_grid_` function calls or types in the tests.

lib/tests/test_rd_sum.cpp
```diff
@@ -0,0 +1,670 @@
+/**
+ * test_rd_sum.cpp
+ * ---------------
+ * Tests for rd_sum that exercise the public summary API.
+ *
+ * Summary files are produced on-the-fly with rd_sum_alloc_writer and
+ * rd_sum_fwrite, and read back with the read_summary RAII utility. Resources
+ * are owned by RAII wrappers (rd_sum_ptr, rd_grid_ptr, rd_file_ptr, rd_kw_ptr,
```
ajaust: `rd_grid_ptr` is mentioned in comment, but it is never used.

lib/tests/test_rd_grid_regular.cpp
```diff
@@ -730,4 +730,159 @@ TEST_CASE_METHOD(Tmpdir, "rd_grid_load_case", "[unittest]") {
...
+        // becomes "CASE" with extension "HIST". The resolver therefore
+        // looks for CASE.EGRID, not CASE.HIST.EGRID.
+        auto hist_egrid = dirname / "CASE.HIST.EGRID";
+        rd_grid_fwrite_EGRID2(grid_3x3x3.get(), hist_egrid.c_str(),
+                              RD_METRIC_UNITS);
+        auto basename_with_dot = dirname / "CASE.HIST";
+        THEN("the resolver strips the trailing component and finds CASE.EGRID "
+             "(written in the fixture)") {
+            load_and_compare(basename_with_dot, grid_5x5x5.get());
+        }
```
ajaust: I am slightly confused. We write a `grid_3x3x3` (as egrid), but  we expect that the `5x5x5` grid is loaded. Is this correct? From the description, I would expect that the file that we have written inside this test will be read since it looks like we explicitly add a file extension to the `3x3x3` grid file that should be stripped/ignored. 

I think we should either skip writing the `3x3x3` grid or make it clear what the purpose of writing the `3x3x3` grid is.

tests/test_convert.py
```diff
@@ -0,0 +1,151 @@
...
+import pytest
+import resfo
+
+CONVERT_X = shutil.which("convert.x")
+
+
+pytestmark = pytest.mark.skipif(
+    CONVERT_X is None,
+    reason="convert.x binary is not on PATH (only built on Linux)",
+)
```
ajaust, line 22: This will silently skip the tests if we are on Linux, but `covnert.x` is missing (for whatever reason). Is this ok or should the test fail if we are on Linux, but `convert.x` is absent?

tests/test_convert.py
```diff
@@ -0,0 +1,151 @@
...
+                np.testing.assert_array_equal(arr_a, arr_b)
+        else:
+            assert arr_a == arr_b
+
+
+@pytest.fixture
+def simple_unsmry(use_tmpdir):
+    for name in ("SIMPLE.UNSMRY", "SIMPLE.SMSPEC"):
+        shutil.copy(TEST_DATA / name, name)
+    yield Path("SIMPLE.UNSMRY")
```
ajaust, line 76: Do we need to copy `SIMPLE.SMSPEC`? The fixture's name only menstions the `unsmry` file.

tests/test_convert.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+def test_that_converting_unrecognized_file_produces_error_message(simple_unsmry):
+    Path("garbage.txt").write_text("not a res file\n")
+    result = run_convert("SIMPLE.UNSMRY", "garbage.txt")
+    assert result.returncode != 0
+    assert "must be recognizable" in result.stderr
+    # The first recognizable is still converted
+    assert Path("SIMPLE.FUNSMRY").exists()
+    assert not Path("garbage.txt.out").exists()
```
ajaust, line 135: Do we need the valid input file `SIMPLE.UNSMRY`? From the name of the test, it sounds like it would be good enough to only test with the invalid file `garbage.txt`. I can understand if we want to verify the current behavior (convert all files until an invalid file is encountered), but I want to be sure that this is done on purpose.

tests/test_convert.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+def test_that_restart_file_is_converted_to_formatted(simple_unsmry):
+    shutil.copy("SIMPLE.UNSMRY", "SIMPLE.X0042")
```
ajaust, line 139: Why do we `shutil.copy` here, but many other places we used `os.rename`?

tests/test_convert.py
```diff
@@ -0,0 +1,151 @@
...
+    assert "Converting SIMPLE.X0042 -> SIMPLE.F0042" in result.stdout
+    assert Path("SIMPLE.F0042").exists()
+    assert_equal_fortran_files("SIMPLE.X0042", "SIMPLE.F0042")
+
+
+def test_that_convert_does_not_overwrite_when_target_matches_source(simple_unsmry):
+    before = Path("SIMPLE.UNSMRY").read_bytes()
+    assert run_convert("SIMPLE.UNSMRY").returncode == 0
+    after = Path("SIMPLE.UNSMRY").read_bytes()
+    assert before == after
```
ajaust, line 151: I understood the convert script the following way: Running `run_convert("SIMPLE.UNSMRY")` will create a file called `SIMPLE.FUNSMRY`.

From this the test does the following according to my understanding:
1. Read source file `SIMPLE.UNSMRY`. 
2. Convert source file `SIMPLE.UNSMRY` to `SIMPLE.FUNSMRY`
3. Reads the source file `SIMPLE.UNSMRY` again.

If my understanding is correct, then I don't understand the purpose of the test because it only reads source files, but not target files. At what point is it checked that nothing is overwritten? I also do not see how it is implied that the target file is the source file.

tests/test_rd_pack.py
```diff
@@ -0,0 +1,196 @@
...
+
+import os
+import shutil
+import subprocess
+from pathlib import Path
+
+import numpy as np
+import pytest
+import resfo
+from .test_rd_unpack import run_unpack
```
ajaust, line 19: This might this lead to an issue if `rd_unpack.x` is missing on the system, but `rd_pack.x` is available since the tests are not able to run `rd_unpack.x` in this case. Should we add a guard for the tests that depend on `rd_unpack.x`?

tests/test_rd_pack.py
```diff
@@ -0,0 +1,196 @@
...
+        if arr_a.dtype.kind == "f" and arr_b.dtype.kind == "f":
+            np.testing.assert_allclose(arr_a, arr_b, rtol=1e-5, atol=1e-7)
+        elif arr_a.dtype.kind in ("S", "U") or arr_b.dtype.kind in ("S", "U"):
+            assert [_normalize_kw(x) for x in arr_a.tolist()] == [
+                _normalize_kw(x) for x in arr_b.tolist()
+            ]
+        else:
+            np.testing.assert_array_equal(arr_a, arr_b)
+    else:
+        assert arr_a == arr_b
```
ajaust, line 71: I think this is the third commit defining these methods in a local test file, We could consider moving them into a shared module.

tests/test_rd_pack.py
```diff
@@ -0,0 +1,196 @@
...
+        shutil.copy(TEST_DATA / name, name)
+    yield Path(".")
+
+
+@pytest.fixture
+def split_restart(simple_unified):
+    """Unpack SIMPLE.UNRST into Xnnnn files and remove the unified source."""
+    assert run_unpack("SIMPLE.UNRST").returncode == 0
+    # Keep a pristine copy for round-trip comparisons.
+    shutil.copy("SIMPLE.UNRST", "ORIG.UNRST")
```
ajaust, line 94: - Why do we make the copy after running `run_unpack`? Wouldn't the file be "most" pristine if we copy it before we run `run_unpack` on it to rule out that the unpacking did not change the file?
- Why do we do a `copy` + `remove` instead of `move`?

tests/test_rd_pack.py
```diff
@@ -0,0 +1,196 @@
...
+    os.remove("SIMPLE.UNRST")
+    return sorted(str(p) for p in Path(".").glob("SIMPLE.X[0-9]*"))
+
+
+@pytest.fixture
+def split_summary(simple_unified):
+    """Unpack SIMPLE.UNSMRY into Snnnn files and remove the unified source."""
+    assert run_unpack("SIMPLE.UNSMRY").returncode == 0
+    shutil.copy("SIMPLE.UNSMRY", "ORIG.UNSMRY")
+    os.remove("SIMPLE.UNSMRY")
```
ajaust, line 104: Same question about order of copying and need for removal as [above](https://github.com/equinor/resdata/pull/1168/changes/19c033ee608c80a767385f915d839b3704c32b73#r3266811903).

lib/include/resdata/rd_util.hpp
```diff
@@ -107,8 +108,10 @@ extern "C" rd_file_enum rd_get_file_type(const char *filename, bool *fmt_file,
 char *rd_alloc_filename(const char * /* path */, const char * /* base */,
                         rd_file_enum, bool /* fmt_file */, int /*report_nr*/);
 
-std::filesystem::path rd_alloc_filename(std::filesystem::path path,
-                                        rd_file_enum, bool /* fmt_file */);
+namespace rd {
+std::filesystem::path filename(std::filesystem::path path, rd_file_enum,
+                               bool fmt_file, int report_nr = -1);
+}; // namespace rd
```
ajaust: The trailing `;` is not needed for namespaces.

lib/include/resdata/rd_util.hpp
```diff
@@ -128,3 +131,8 @@ int rd_select_filelist(const char *path, const char *base,
 void rd_set_datetime_values(time_t t, int *sec, int *min, int *hour, int *mday,
                             int *month, int *year);
 bool rd_path_access(const char *rd_case);
+
+inline bool accessible(std::filesystem::path p) noexcept {
+    std::error_code ec;
+    return std::filesystem::exists(p, ec);
+}
```
ajaust, line 136: - Should this also be in the `rd` namespace?
- You mentioned that you were uncertain about the name. I think the suggested name of `try_exists` would actually be nicer. `accessible` can be misleading (existence of file vs. having access rights to the file).

lib/resdata/tests/rd_filenames.cpp
```diff
@@ -64,39 +61,32 @@ void test_file_list() {
...
+        FILE *f3 = fopen(restart_file1.c_str(), "w");
         fclose(f3);
-
-        free(summary_file1);
-        free(summary_file2);
-        free(restart_file1);
     }
-    printf(
-        "-----------------------------------------------------------------\n");
+    printf("---------------------------------------------------------------\n");
```
ajaust, line 89: The number of dashes changed. Not sure if this was on purpose. The line had 65 dashes, but now there are 63.

applications/resdata/convert.cpp
```diff
@@ -26,43 +26,42 @@
...
+    ERT::FortIO src(src_file, std::ios_base::in, formatted_src);
 
     while (true) {
-        if (fortio_read_at_eof(src))
+        if (fortio_read_at_eof(src.get()))
             break;
 
         {
-            rd_kw_type *rd_kw = rd_kw_fread_alloc(src);
+            rd_kw_type *rd_kw = rd_kw_fread_alloc(src.get());
```
ajaust, line 55: This probably could be a smart pointer making the `free` obsolete, but I am not sure if this line will disappear in a future commit anyway.

lib/resdata/tests/rd_grid_volume.cpp
```diff
@@ -61,13 +64,9 @@ int main(int argc, char **argv) {
     printf("Total volume difference: %g %% \n",
            100 * total_diff / total_volume);
 
-    rd_grid_free(grid);
-    rd_file_close(init);
-    free(grid_file);
-    free(init_file);
     if (error_found) {
         printf("Error_count: %d / %d \n", error_count,
-               rd_grid_get_nactive(grid));
```
ajaust, line 70: Nice fix! This was a bug before (accessing `grid` after freeing it in the error case.

lib/resdata/rd_smspec.cpp
```diff
@@ -32,6 +31,8 @@
 #include <fnmatch.h>
 #endif
 
+namespace fs = std::filesystem;
+
```
ajaust, line 35: The `fs` namespace appears to be unused in this commit.

lib/include/resdata/rd_util.hpp
```diff
@@ -133,3 +136,30 @@ inline bool accessible(std::filesystem::path p) noexcept {
...
+        return;
+    }
+    T *raw_ptr = ptr.release();
+    void *new_raw_ptr = std::realloc(raw_ptr, new_element_count * sizeof(T));
+    if (new_raw_ptr == nullptr) {
+        ptr.reset(raw_ptr);
+        throw std::bad_alloc{};
+    }
+    ptr.reset(static_cast<T *>(new_raw_ptr));
+}
```
ajaust, line 162: This is in the global namespace now. Is this fine or should we place it in the `rd` namespace?

lib/include/resdata/rd_util.hpp
```diff
@@ -1,12 +1,15 @@
...
 #include <ctime>
+#include <cstdlib>
 
 #include <ert/util/stringlist.hpp>
 #include <ert/util/time_t_vector.hpp>
 #include <ert/util/parser.hpp>
 #include <resdata/rd_type.hpp>
 #include <filesystem>
 #include <system_error>
+#include <optional>
```
ajaust: `std::optional` is unused in the current commit.

lib/include/resdata/rd_util.hpp
```diff
@@ -163,3 +165,11 @@ void checked_realloc(std::unique_ptr<T[], void (*)(void *)> &ptr,
...
     ptr.reset(static_cast<T *>(new_raw_ptr));
 }
+
+inline std::string strip_spaces(std::string_view s) {
+    auto first = s.find_first_not_of(' ');
+    if (first == std::string_view::npos)
+        return {};
+    auto last = s.find_last_not_of(' ');
+    return std::string(s.substr(first, last - first + 1));
+}
```
ajaust, line 169: We should consider moving it into the `rd` namespace.

lib/resdata/rd_sum.cpp
```diff
@@ -393,46 +375,43 @@ static bool rd_alloc_summary_files(const char *path, const char *_base,
...
-            return false; /* If you insist on e.g. unformatted and only fsmspec exists - no results for you. */
+    if (fmt_file) {
+        header_file = fsmspec_file;
+    } else {
+        header_file = smspec_file;
     }
 
+    if (!fs::exists(header_file))
+        return std::
+            nullopt; /* If you insist on e.g. unformatted and only fsmspec exists - no results for you. */
```
ajaust: Is this autoformat trying to have lines that are as short as possible leading to a line break betweeh `std::` and `nullopt`?

ajaust (approved):
I have mostly minor comments. You have to check which of these are relevant/applicable.

lib/tests/test_rd_sum.cpp
eivindjahren: Yes, sorry about that. There was a test involving a grid that I was looking into, ended up omitting it but forgot to remove the header

lib/tests/test_rd_sum.cpp
eivindjahren: I just removed this comment as there wasn't really anything of use here.

lib/resdata/rd_sum.cpp
eivindjahren: Yes, I moved the comment to above.

lib/resdata/rd_sum.cpp
eivindjahren: Ah yes, I initially used it for read_summary, but that just created problems.

lib/include/resdata/rd_util.hpp
eivindjahren: Fixed

lib/tests/test_rd_sum.cpp
eivindjahren: Removed it

lib/tests/test_rd_grid_regular.cpp
eivindjahren: I will figure out how to make this more clear. The thing is we expect not to find the grid_3x3x3 when we look for `CASE.HIST` as the extension gets stripped and we find `CASE.EGRID`(grid_5x5x5) and not `CASE.HIST.EGRID`(grid_3x3x3). I think I can just make this clearer by renaming the variables and explicitly writing both grids.

lib/resdata/rd_sum.cpp
eivindjahren: Removed now

lib/tests/test_rd_grid_regular.cpp
eivindjahren: I updated the test to make it clearer.

lib/include/resdata/rd_util.hpp
eivindjahren: Fixed

PR #1169, opened 2026-05-18
Avoid overriding Python builtins

berland:
**Issue**
Resolves Thou shall not override python builtins 

**Approach**
`nvim` rename feature


andreas-el (approved):
💯

PR #1170, opened 2026-05-19
Sort imports (and enforce with ruff)

eivindjahren:
This ment sometimes making implict cwrap dependencies explicitly imported. These imports were then used as typehints.

Also resolved some circular imports that were created because the module was imported from (from my.submodule import Sibling). Changed to importing from defining file (from .sibling import Sibling).


berland (commented):
lgtm! ❤️

berland (approved):

PR #1171, opened 2026-05-20
Use std::shared_ptr for data_files

eivindjahren:


lib/resdata/rd_sum_data.cpp
```diff
@@ -214,7 +215,7 @@ class CaseIndex {
 
 struct rd_sum_data_struct {
     const rd_smspec_type *smspec;
-    std::vector<rd::rd_sum_file_data *>
+    std::vector<std::shared_ptr<rd::rd_sum_file_data>>
```
ajaust, line 218: Maybe add a short comment why we need a `shared_ptr` here instead of a `unique_ptr` as most of the other, modernized code parts.

lib/tests/test_rd_sum.cpp
```diff
@@ -66,7 +67,7 @@ time_t write_test_summary(const std::string &case_path, const WriteSpec &spec,
     }
     rd_sum_fwrite(rd_sum.get());
 
-    time_t end_time = spec.start_time;
+    time_t end_time = spec.start_seconds + spec.start_time;
     util_inplace_forward_seconds_utc(
         &end_time,
         (spec.num_report_steps * spec.num_ministep - 1) * spec.ministep_length);
```
ajaust, line 73: There is some mix of and `double` (`spec.start_seconds`) and `time_t` (`spec.start_time`) here. Not sure if this could be an issue due to implicit conversion.

ajaust (approved):

lib/resdata/rd_sum_data.cpp
eivindjahren: Done

lib/tests/test_rd_sum.cpp
eivindjahren: Not great, but we depend on it in `util_inplace_forward_seconds_utc` anyways. Probably worth revisiting though.

PR #1172, opened 2026-05-20
Convert most rtype and type comments to typehints

eivindjahren:
Keep resdata.geometry as is for now as it is not straightforward to convert due to sometimes using 3d points and sometimes 2d points.


berland (approved):
❤️

PR #1173, opened 2026-05-20
Avoid manually calling rd_grid_free in all tests

eivindjahren:
Also removes some tests that were not compiled and for code that no longer exists.


lib/resdata/tests/rd_nnc_index_list.cpp
ajaust, line 1: There is a typo in the commit message: `non-existant` -> `non-existent`

lib/include/resdata/rd_grid.hpp
```diff
@@ -206,5 +206,8 @@ using rd_grid_ptr = std::unique_ptr<rd_grid_type, decltype(&rd_grid_free)>;
 rd_grid_ptr make_rectangular_grid(int nx, int ny, int nz, double dx, double dy,
                                   double dz, const int *actnum);
 rd_grid_ptr read_grid(const std::filesystem::path &filename);
+inline rd_grid_ptr copy_grid(const rd_grid_ptr &src_grid) {
+    return {rd_grid_alloc_copy(src_grid.get()), &rd_grid_free};
+}
```
ajaust: If you want to fully live by the C++ core guidelines, we should rewrite this to accept a raw pointer instead of a reference to a smart pointer, see [F.7: For general use, take T* or T& arguments rather than smart pointers](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f7-for-general-use-take-t-or-t-arguments-rather-than-smart-pointers), because `copy_grid` does not take ownership (indicated due to by-reference signature). However, this would give uglier signatures when using `copy_grid` so you can also leave it as is.

ajaust (approved):
Looks good. Found one type and one minor comment that you also may ignore. 🙂

lib/include/resdata/rd_grid.hpp
eivindjahren: I think we leave it as is so that we can turn `rd_grid_ptr` into `Grid` in the long term.

lib/resdata/tests/rd_nnc_index_list.cpp
eivindjahren: Fixed

lib/include/resdata/rd_grid.hpp
eivindjahren: I ended up following the core guidelines.

PR #1175, opened 2026-05-22
Replace old summary tests with catch2 equivalents

eivindjahren:
### Translation

* `rd_sum_alloc_resampled_test` becomes the scenario "rd_sum_alloc_resample over a time vector"

* `rd_unsmry_loader_test` becomes the scenario "rd::unsmry_loader reads back values from a UNSMRY"

* `rd_sum_data_intermediate_test` and run_sum_restart becomes the scenario "Loading restarts"

For `test_rd_sum`, `test_copy` was never actually called. The test used internal test data, which we have not been using for c++ tests since https://github.com/equinor/resdata/commit/8e711e26b8899d6be7bc19fec0379a9e4c507247 . So its better to use test data generated on the fly here.

### Changes to the tmpdir fixture

First the Tmpdir fixture is made to create shorter directory names as resdata limits the size of the RESTART file to 8*8 characters and we reach this limit on mac. This limit is there for interoperability with older versions of simulators and we should reconsider it, but for now it would be best to be able to test the existing behavior well instead.

Then two old c style tests are converted to catch to make them more like executable specification.

I used the following to check that the tmdir fixture is threadsafe:
```
int main() {
    const int NUM_THREADS = 8;
    const int DIRS_PER_THREAD = 10000;
    const int TOTAL = NUM_THREADS * DIRS_PER_THREAD;

    auto parentdir = fs::temp_directory_path() / "rdtest";
    fs::remove_all(parentdir);

    std::atomic<int> success_count{0};
    std::atomic<int> fail_count{0};

    // Each claim gets a monotonic ID and a timestamp
    struct Claim {
        fs::path path;
        int claim_id;
        std::chrono::steady_clock::time_point acquired;
        std::chrono::steady_clock::time_point released;
    };
    std::vector<Claim> all_claims(TOTAL);
    std::atomic<int> claim_idx{0};
    std::atomic<int> next_claim_id{0};

    auto worker = [&](int thread_id) {
        try {
            for (int i = 0; i < DIRS_PER_THREAD; ++i) {
                Tmpdir t;
                t.delete_temporary_files = true;

                if (!fs::is_directory(t.dirname)) {
                    std::cerr << "Thread " << thread_id
                              << ": directory not found: " << t.dirname << "\n";
                    fail_count++;
                    continue;
                }

                auto name = t.dirname.filename().string();
                int num = std::stoi(name);
                if (num < 0 || num >= MAX_SUBDIRS) {
                    std::cerr << "Thread " << thread_id
                              << ": unexpected dirname: " << t.dirname << "\n";
                    fail_count++;
                    continue;
                }

                // Record the claim while we still hold it
                int idx = claim_idx.fetch_add(1);
                all_claims[idx] = {
                    t.dirname,
                    next_claim_id.fetch_add(1),
                    std::chrono::steady_clock::now(),
                    {}  // released filled in destructor-like scope
                };

                success_count++;

                // Write a file to verify ownership
                auto testfile = t.dirname / ("owned_by_" + std::to_string(thread_id));
                std::ofstream(testfile) << "hello from thread " << thread_id;

                std::this_thread::sleep_for(std::chrono::milliseconds(1));

                // Record release time just before destruction
                all_claims[idx].released = std::chrono::steady_clock::now();
            }
        } catch (const std::exception& e) {
            std::cerr << "Thread " << thread_id << " exception: " << e.what() << "\n";
            fail_count++;
        }
    };

    std::vector<std::thread> threads;
    for (int i = 0; i < NUM_THREADS; ++i) {
        threads.emplace_back(worker, i);
    }
    for (auto& t : threads) {
        t.join();
    }

    // Check for overlapping claims of the same path
    int n = claim_idx.load();
    std::sort(all_claims.begin(), all_claims.end(),
        [](const Claim& a, const Claim& b) {
        if (a.path != b.path) return a.path < b.path;
        return a.acquired < b.acquired;
    });

    bool has_overlap = false;
    for (int i = 1; i < n; ++i) {
        if (all_claims[i].path == all_claims[i - 1].path) {
            auto& a = all_claims[i - 1];
            auto& b = all_claims[i];
            // Overlap = A hasn't been released yet when B is acquired
            if (a.released > b.acquired) {
                std::cerr << "OVERLAP: " << a.path
                          << " claim " << a.claim_id
                          << " and claim " << b.claim_id << "\n";
                has_overlap = true;
            }
        }
    }

    int remaining = 0;
    if (fs::exists(parentdir)) {
        for (auto& entry : fs::directory_iterator(parentdir)) {
            remaining++;
            std::cerr << "Left over: " << entry.path() << "\n";
        }
    }

    std::cout << "Successes:  " << success_count << "/" << TOTAL << "\n";
    std::cout << "Failures:   " << fail_count << "\n";
    std::cout << "Overlaps:   " << (has_overlap ? "YES" : "NO") << "\n";
    std::cout << "Remaining:  " << remaining << "\n";

    return (success_count == TOTAL && !has_overlap && remaining == 0) ? 0 : 1;
}
```


lib/tests/test_rd_sum.cpp
```diff
@@ -696,3 +708,300 @@ TEST_CASE_METHOD(Tmpdir, "write then read produces matching case key") {
...
+
+                // 3 (CASE1) + 2 (CASE2) + 4 (CASE3) = 9 ministeps; the first
+                // 5 are filled with the default (the restart chain doesn't
+                // contain BPR:8), the last 4 are the BPR:3 values copied into
+                // PARAMS[4] when CASE4 was synthesised.
+                check_vector(sum.get(), "BPR:8",
+                             {
+                                 missing,
+                                 missing,
+                                 missing,
```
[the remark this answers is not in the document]
eivindjahren: Fixed

lib/tests/tmpdir.hpp
```diff
@@ -1,31 +1,67 @@
...
 #include <string>
 
 #include <filesystem>
+#include <system_error>
+
 namespace fs = std::filesystem;
 
-class Tmpdir {
-public:
+// We do not expect more than single-digit subdirs,
```
ajaust, line 12: What ise meant by "single-digit subdirs". Does it mean we expect <10 subdirectories to be created? Couldn't we set `MAX_SUBDIRS` to a much lower number in this case?

lib/tests/test_rd_sum.cpp
```diff
@@ -696,3 +709,473 @@ TEST_CASE_METHOD(Tmpdir, "write then read produces matching case key") {
...
+                // BPR at sim_day 1: between (day 0, 0) and (day 3, 2592000)
+                REQUIRE_THAT(
+                    rd_sum_get_from_sim_time(
+                        resampled.get(), util_make_date_utc(2, 1, 2010), &bpr),
+                    WithinAbs(864000.0, 1e-3));
+                // WWCT exactly at sim_day 3
+                REQUIRE_THAT(
+                    rd_sum_get_from_sim_time(
+                        resampled.get(), util_make_date_utc(4, 1, 2010), &wwct),
+                    WithinAbs(25920000.0, 1e-3));
```
ajaust, line 1209: I don't fully understand what is going on, but why were the numbers changed fomr `3.33333`, `3.33333`, `10.00000` to `432000.0`, `864000.0` and `25920000.0`. Are the tests still fully equivalent as mentioned by the commit message?

lib/tests/test_rd_sum.cpp
```diff
@@ -696,3 +709,473 @@ TEST_CASE_METHOD(Tmpdir, "write then read produces matching case key") {
...
+                REQUIRE_THAT(fopt[3], WithinAbs(259200.0, 1e-6));
+                REQUIRE_THAT(bpr[2], WithinAbs(1728000.0, 1e-6));
+                REQUIRE_THAT(wwct[1], WithinAbs(8640000.0, 1e-6));
+            }
+        }
+    }
+}
+
+SCENARIO_METHOD(Tmpdir, "rd_sum_alloc_resample over a time vector") {
+    GIVEN("A summary case sampled at sim_days 0, 3, 6, 9") {
```
ajaust: It looks like the sampling has changed from the previous implementation to larger time intervals (more days). Is this correct? If yes, why did we do this?

lib/tests/test_rd_sum.cpp
```diff
@@ -696,3 +709,473 @@ TEST_CASE_METHOD(Tmpdir, "write then read produces matching case key") {
...
+    GIVEN("A chain CASE1 -> CASE2 -> CASE3 of summary cases with BPR:1-3") {
+        const time_t start_time = util_make_date_utc(1, 1, 2010);
+        const double ministep_length = 86400.0;            // 1 day in seconds
+        const double step_spacing = 3.0 * ministep_length; // 3 days per step
+        const int nx = 10, ny = 10, nz = 10;
+
+        const std::vector<std::vector<double>> case1_bpr = {
+            {100.0, 200.0, 300.0},
+            {110.0, 210.0, 310.0},
+            {120.0, 220.0, 320.0},
```
ajaust, line 803: This case data seems to have the same row order as in the old test file, but case 2 and 3 seem to have reversed row order. I am wondering if this is done on purpose.

lib/resdata/tests/rd_sum_restart.cpp
ajaust, line 1: There is a typo in the commit message. It mentions `run_sum_restart` instead of `rd_sum_restart`.

lib/tests/test_rd_sum.cpp
```diff
@@ -696,3 +709,473 @@ TEST_CASE_METHOD(Tmpdir, "write then read produces matching case key") {
...
+                rd_file_iget_named_kw(smspec_in.get(), "KEYWORDS", 0);
+            rd_kw_resize(keywords, 5);
+            rd_kw_iset_char_ptr(keywords, 3, "WTPRWI1");
+            rd_kw_iset_char_ptr(keywords, 4, "BPR");
+
+            rd_kw_type *nums =
+                rd_file_iget_named_kw(smspec_in.get(), "NUMS", 0);
+            rd_kw_resize(nums, 5);
+            auto *nums_ptr =
+                static_cast<unsigned int *>(rd_kw_get_void_ptr(nums));
```
ajaust: Why did this change to `rd_kw_get_void_ptr`? It was `rd_kw_get_int_ptr` before? Could we do this without casting, for example as,

```suggestion
            int *nums_ptr = rd_kw_get_int_ptr(nums);
```

lib/resdata/tests/rd_smspec.cpp
ajaust, line 1: The commit message mentions that `test_rd_sum` is being replaced, but it seems like that `rd_smspec.cpp` is being replaced. It is intergrated into `test_rd_sum`.

lib/CMakeLists.txt
```diff
@@ -542,10 +541,6 @@ add_test(NAME rd_rsthead
          COMMAND rd_rsthead ${_resdatapath}/Gurbat/ECLIPSE.UNRST
                  ${_resdatapath}/DualPoro/DUALPORO.X0005)
 
-add_test(NAME rd_smspec
-         COMMAND rd_smspec ${_resdatapath}/Gurbat/ECLIPSE.SMSPEC
-                 ${_resdatapath}/Heidrun/Summary/FF12_2013B3_CLEAN_RS.SMSPEC)
-
```
ajaust, line 552: Is it good that we remove this file from the test input files? Will we also remove the testfile from whatever drive it is living on?

lib/tests/test_rd_sum.cpp
```diff
@@ -292,31 +310,98 @@ TEST_CASE_METHOD(Tmpdir, "Loading a case with data before its parent in time") {
...
+    }
+}
+
+TEST_CASE_METHOD(Tmpdir, "Restart names >64 characters are ignored") {
+    const time_t start_time = util_make_date_utc(1, 1, 2010);
+    const std::string too_long(72, 'A');
+    rd_smspec_type *smspec = rd_smspec_alloc_restart_writer(
+        ":", too_long.c_str(), 10, start_time, true, 3, 3, 3);
+    REQUIRE(rd_smspec_get_restart_case(smspec) == nullptr);
+    rd_smspec_free(smspec);
```
ajaust: Most memory is managed by smart pointers, but this still appears to be a raw pointer. Not too relevant for a test though.

lib/tests/test_rd_sum.cpp
```diff
@@ -38,6 +39,19 @@
...
+struct Chdir {
+    fs::path prev;
+    explicit Chdir(const fs::path &target) : prev(fs::current_path()) {
+        fs::current_path(target);
+    }
+    ~Chdir() {
+        std::error_code ec;
+        fs::current_path(prev, ec);
+    }
+};
```
ajaust, line 52: Not sure if this matter for your thread-safety concerns, but this is probably not thread-safe.

ajaust (approved):
Looks good overall. Some minor comments. You might have to go through it commit-by-commit to see all comments since some comments are attached to deleted  files.

lib/tests/tmpdir.hpp
eivindjahren: We could, but it was already lightening fast to go all the way up to the limit on ext3.

lib/tests/test_rd_sum.cpp
eivindjahren: Yes, that is because the original test used different numbers. I wanted to reuse the numbers in the already created file for simplicity.

lib/resdata/tests/rd_smspec.cpp
eivindjahren: Fixed

lib/resdata/tests/rd_sum_restart.cpp
eivindjahren: Fixed

lib/CMakeLists.txt
eivindjahren: Yes, it is unfortunately not running now anyways. I think we can keep the file around for a while at least.

lib/tests/test_rd_sum.cpp
eivindjahren: Yes, indeed. My thread safety concerns was just about ensuring that two instances of running the tests did not have a race condition. That could happen on eg. self-hosted runners.

lib/tests/test_rd_sum.cpp
eivindjahren: Fixed

lib/tests/test_rd_sum.cpp
eivindjahren: This was done to separate the test into interpolate, exact match, and extrapolate cases. The original test would use days 1,2,3,4 and resample to 2,4,6,8 and then check in one go. This new setup would check rate vs non-rate also.

lib/tests/test_rd_sum.cpp
eivindjahren: Oops, that was a mistake. Fixed now.

lib/tests/test_rd_sum.cpp
eivindjahren: I changed the tests so that it more obviously covers all the cases that the old assertions did.

lib/tests/test_rd_sum.cpp
eivindjahren: I think you are referring to this comment:

```
CASE1
---------
1 : BPR:1     100      200      300      400     500   600   700   800   900   ....
2 : BPR:2     110      210      310      410     510   610   710   810   910   ....
3 : BRP:3     120      220      320      420     520   620   720   820   920   ....



CASE2
---------
1 : BRP:2                            1100      2100     3100    4100
2 : BPR:1                            1000      2000     3000    4000



Total CASE2:
------------
1 : BPR:2   110  210  310  1100   2100   3100  4100
2 : BRP:1   100  200  300  1000   2000   3000  4000



CASE3
-----
1 : BPR:3                                            12000   22000   32000   42000
2 : BRP:2                                            11000   21000   31000   41000
3 : BPR:1                                            10000   20000   30000   40000
```

And in CASE1 the order of the keys are `BPR:1`, `BPR:2`, `BPR:3` but in all the other cases that order is for some reason reversed. So the new one has the order the correct way round.

PR #1176, opened 2026-05-29
Avoid manual memory management in rd_sum.cpp

eivindjahren:


python/resdata/summary/rd_sum.py
```diff
@@ -330,9 +206,20 @@ def writer(
...
+            time_in_days,
+            nx,
+            ny,
+            nz,
+            None,
+            0,
         )
+        smry = Summary._python_object_from_ptr(ptr)
         smry._load_case = "writer"
         return smry
```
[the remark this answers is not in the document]
eivindjahren: This is the case, and was the case before. We should probably create an issue for it.

python/resdata/summary/rd_sum.py
```diff
@@ -356,19 +243,20 @@ def restart_writer(
...
             time_in_days,
             nx,
             ny,
             nz,
+            restart_case,
+            restart_step,
         )
+        smry = Summary._python_object_from_ptr(ptr)
         smry._load_case = "restart_writer"
         return smry
```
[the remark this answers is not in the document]
eivindjahren: This is the case, and was the case before. We should probably create an issue for it.

lib/include/resdata/rd_sum.hpp
```diff
@@ -121,23 +108,7 @@ rd_smspec_type *rd_sum_get_smspec(const rd_sum_type *rd_sum);
...
-                                         const char *key_join_string,
-                                         time_t sim_start, bool time_in_days,
-                                         int nx, int ny, int nz);
-rd_sum_type *rd_sum_alloc_writer(const char *rd_case, bool fmt_output,
-                                 bool unified, const char *key_join_string,
-                                 time_t sim_start, bool time_in_days, int nx,
-                                 int ny, int nz);
+void rd_sum_set_case(rd_sum_type *rd_sum, const std::string &input_arg);
 void rd_sum_fwrite(const rd_sum_type *rd_sum);
 bool rd_sum_can_write(const rd_sum_type *rd_sum);
```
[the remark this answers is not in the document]
eivindjahren: The C interface is private.

python/resdata/summary/rd_sum.py
```diff
@@ -280,10 +156,10 @@ def load(cls, smspec_file, unsmry_file, key_join_string=":", include_restart=Tru
 
         data_files = StringList()
         data_files.append(unsmry_file)
-        c_ptr = cls._fread_alloc(
-            smspec_file, data_files, key_join_string, include_restart
+        c_ptr = _rd_sum._fread_alloc(
+            smspec_file, data_files, key_join_string, include_restart, False, 0
         )
```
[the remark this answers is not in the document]
eivindjahren: Actually, this fixes a bug where not all arguments to fread_alloc was given, causing the variables to not be initialized.

lib/tests/test_rd_sum.cpp
```diff
@@ -76,13 +77,12 @@ time_t write_test_summary(const std::string &case_path, const WriteSpec &spec,
...
-    rd_smspec_type *smspec = rd_sum_get_smspec(rd_sum.get());
     const rd::smspec_node *fopt =
-        rd_smspec_add_node(smspec, "FOPT", "SM3", 99.0f);
+        rd_sum_add_var(rd_sum.get(), "FOPT", nullptr, 0, "SM3", 99.0f);
     const rd::smspec_node *bpr =
-        rd_smspec_add_node(smspec, "BPR", 567, "BARS", 0.0f);
+        rd_sum_add_var(rd_sum.get(), "BPR", nullptr, 567, "BARS", 0.0f);
     const rd::smspec_node *wwct =
-        rd_smspec_add_node(smspec, "WWCT", "OP-1", "(1)", 0.0f);
+        rd_sum_add_var(rd_sum.get(), "WWCT", "OP-1", 0, "(1)", 0.0f);
```
ajaust, line 85: I think we don't have any test anymore that explicitly calls `rd_smspec_add_node`. Not sure if this is relevant w.r.t. test coverage. We still test it implicitly by other method calls that call `rd_smspec_add_node` (e.g., via `rd_sum_add_var`).

lib/include/resdata/rd_smspec.hpp
```diff
@@ -27,12 +26,6 @@ rd_smspec_iget_node_w_node_index(const rd_smspec_type *smspec, int node_index);
 const rd::smspec_node &
 rd_smspec_iget_node_w_params_index(const rd_smspec_type *smspec,
                                    int params_index);
-#endif
-
-#ifdef __cplusplus
-extern "C" {
```
ajaust, line 33: This changes the symbols/linkage. Is this an issue? I think the file will be installed in `CMAKE_INSTALL_PREFIX/include` and thus could be considered part of the public interface.

We probably change the interface the whole time with the current refactoring and bug fixing, but I didn't think about it before or I didn't realise the we seem to install all headers into the public `include/` directory.

lib/include/resdata/rd_sum_data.hpp
```diff
@@ -1,4 +1,5 @@
 #pragma once
+#include <cstdio>
```
ajaust, line 2: Nice. Might want to squash this commit into the previous one. 🔨

lib/include/resdata/rd_sum_data.hpp
```diff
@@ -8,6 +8,7 @@
 #include <ert/util/double_vector.hpp>
 #include <ert/util/stringlist.hpp>
 
+#include <memory>
```
ajaust, line 11: This include might be more suited in the section of standard library headers above (line 2-5).

lib/resdata/rd_sum.cpp
```diff
@@ -25,6 +25,7 @@
 #include <resdata/rd_smspec.hpp>
 #include <resdata/rd_sum_data.hpp>
 #include <resdata/smspec_node.hpp>
+#include <utility>
```
ajaust: I think this include should move to the first section of include where standard library includes are located.

lib/resdata/rd_sum.cpp
```diff
@@ -25,9 +30,12 @@
...
 #include <resdata/rd_sum_data.hpp>
 #include <resdata/smspec_node.hpp>
-#include <utility>
 
-#include "detail/util/path.hpp"
+#include <detail/util/path.hpp>
+#include <ert/util/double_vector.hpp>
+#include <ert/util/type_macros.hpp>
+#include <resdata/rd_sum_tstep.hpp>
+#include <resdata/rd_sum_vector.hpp>
```
ajaust: I would consider keeping the includes somewhat ordered by type of include. Since the previous ordering is C/C++ standard library, `<ert/xxx>` and `<resdata/xxx>`, we should probably stick to it. Not sure where `<detail/xxx>` fits in or if this is its own include section.

lib/tests/test_rd_sum.cpp
```diff
@@ -333,6 +333,18 @@ TEST_CASE_METHOD(Tmpdir, "Read summary written by writer") {
...
+            time_t before = spec.start_time;
+            util_inplace_forward_seconds_utc(&before, -spec.ministep_length);
+            REQUIRE_THROWS_AS(rd_sum_get_general_var_from_sim_time(
+                                  rd_sum.get(), before, "FOPT"),
+                              std::out_of_range);
+            time_t after = end_time;
+            util_inplace_forward_seconds_utc(&after, spec.ministep_length);
+            REQUIRE_THROWS_AS(rd_sum_get_general_var_from_sim_time(
+                                  rd_sum.get(), after, "FOPT"),
+                              std::out_of_range);
```
ajaust, line 346: I think there are 5 new exceptions added that replace `util_abort`, but only two new tests expecting a method to throw. Is this correct or should we add more tests that expect an exception to be thrown, i.e., one `REQUIRE_THROWS_AS` test per newly added exception?

tests/rd_tests/test_rd_sum.py
```diff
@@ -579,25 +592,163 @@ def test_summary_writer_basic():
...
+    np.testing.assert_allclose(
+        loaded.numpy_vector(rsft.get_key1()), [3000.0 * i for i in range(1, 5)]
+    )
+
+
+@pytest.mark.usefixtures("use_tmpdir")
+def test_summary_writer_local_variables():
+    start_date = datetime.datetime(2005, 5, 10)
+    # writer does not actually support writing local
+    # variables but will add them
```
ajaust, line 741: Does it mean we cannot test the full round trip (writing file and then reading it again) since the local variables are not written?

lib/include/resdata/rd_sum.hpp
```diff
@@ -1,20 +1,23 @@
...
+#include <memory>
+
 #include <ert/util/stringlist.hpp>
 #include <ert/util/time_t_vector.hpp>
 #include <ert/util/double_vector.hpp>
 
 #include <resdata/rd_smspec.hpp>
 #include <resdata/rd_sum_tstep.hpp>
 #include <resdata/smspec_node.hpp>
+#include <string>
```
ajaust: Ordering of includes is mixed up. It was standard library, `<ert/xxx>`, `<resdata/xxx`> before.

lib/include/resdata/rd_sum.hpp
```diff
@@ -150,8 +150,6 @@ void rd_sum_init_double_vector_interp(const rd_sum_type *rd_sum,
                                       double *data);
 void rd_sum_init_double_vector(const rd_sum_type *rd_sum, const char *gen_key,
                                double *data);
-void rd_sum_init_double_frame(const rd_sum_type *rd_sum,
```
ajaust, line 182: The commit message says that this commit inlines `init_pandas_frame`, but it seems to inline `rd_sum_init_double_frame` instead. `rd_sum_init_double_frame` is inlined within `init_pandas_frame`.

lib/include/resdata/rd_sum.hpp
```diff
@@ -142,8 +142,6 @@ double rd_sum_get_last_value_gen_key(const rd_sum_type *rd_sum,
                                      const char *gen_key);
 double rd_sum_get_first_value_gen_key(const rd_sum_type *rd_sum,
                                       const char *gen_key);
-void rd_sum_init_datetime64_vector(const rd_sum_type *rd_sum, int64_t *data,
```
ajaust, line 174: The commit message mentions that `init_numpy_datetime64 ` is inlined, but we actually inline `rd_sum_init_datetime64_vector` into `init_numpy_datetime64`.

ajaust (commented):
Looks great. I have some minor things that should be simple to fix (ordering of includes, some commit messages are ambiguous).

lib/tests/test_rd_sum.cpp
eivindjahren: I am thinking that we want to inline rd_smspec.cpp and rd_sum_data.cpp into rd_sum.cpp. The interface in rd_sum.hpp is really the one that the python api is using so I think we should try to stick to only that.

lib/include/resdata/rd_smspec.hpp
eivindjahren: The c api is considered private now.

lib/tests/test_rd_sum.cpp
eivindjahren: Tried to reach them, but it seems like these are not really reachable, but just checks bounds.

tests/rd_tests/test_rd_sum.py
eivindjahren: Correct, thats why I added the additional test which adds local variables but doesn't do the roundtrip.

lib/include/resdata/rd_sum.hpp
eivindjahren: Ah right, that is what I ment to say, but got it backwards, fixed it.

lib/include/resdata/rd_sum.hpp
eivindjahren: Fixed.

ajaust (approved):

PR #1184, opened 2026-06-09
Use pybind for FortIO

eivindjahren:


lib/include/ert/util/util.hpp
ajaust, line 1: The last commit `Add missing cmakedefine` should be moved to the second position or squashed appropriately to ensure that the history is bisectable with working tests. At the moment, the Python tests will fail starting from the second commit when `extern C` is removed.

lib/resdata/fortio_pybind.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+    m.def("_guess_fortran", [](const std::string &filename, bool endian_flip) {
+        return fortio_looks_like_fortran_file(filename.c_str(), endian_flip);
+    });
+
+    m.def("_get_position", [](py::handle self) {
+        return fortio_ftell(from_cwrap<fortio_type>(self));
+    });
+
+    m.def("_seek", [](py::handle self, long pos, int whence) {
```
ajaust: Is it safe to use `long pos` here or should it be `offset_type pos` to be consistent with the definition of `fortio_fseek`?

lib/resdata/fortio_pybind.cpp
```diff
@@ -0,0 +1,81 @@
...
+        return fortio_fseek(from_cwrap<fortio_type>(self), pos, whence);
+    });
+
+    m.def("_close", [](py::handle self) {
+        return fortio_fclose(from_cwrap<fortio_type>(self));
+    });
+
+    m.def("_truncate", [](py::handle self, long size) {
+        return fortio_ftruncate(from_cwrap<fortio_type>(self), size);
+    });
```
ajaust: Is it safe to use `long pos` here or should it be `offset_type pos` to be consistent with the definition of `fortio_ftruncate`?

lib/include/resdata/FortIO.hpp
```diff
@@ -1,15 +1,13 @@
...
-#include <fstream>
+#include <ios>
 #include <string>
-#include <memory>
-#include <stdlib.h>
-#include <stdio.h>
-#include <stdbool.h>
+#include <cstdlib>
+#include <cstdio>
+#include <cstdbool>
```
ajaust: We can probably remove `cstdbool` if we compile this as C++ since `bool`, `true`, and `false` are part of the C++ language.

https://en.cppreference.com/cpp/header/cstdbool

lib/resdata/FortIO.cpp
```diff
@@ -437,11 +442,11 @@ bool fortio_data_fskip(fortio_type *fortio, const int element_size,
...
                        const int element_count, const int block_size) {
-    if (element_count < 0 ||
-        data_element >= static_cast<size_t>(element_count)) {
-        util_abort("%s: Element index is out of range: 0 <= %zu < %d \n",
-                   __func__, data_element, element_count);
-    }
+    if (element_count < 0 || data_element >= static_cast<size_t>(element_count))
+        throw std::invalid_argument(
+            fmt::format("Element index is out of range: 0 <= {} < {}",
+                        data_element, element_count));
```
ajaust, line 448: The commit message says:

```
The added logic_error should not be reachable.

Note that we keep the somewhat out of place exception
in fseek as opposed to returning false to not change to
a silent failure on the python side.
```

I don't see any `logic_error` being added. I am wondering which code parts should be unreachable now. Please align the commit message with the content of the commit.

ajaust (approved):
Looks great overall. I have some smaller things that I commented on.

lib/include/ert/util/util.hpp
eivindjahren: Fixed!

lib/resdata/fortio_pybind.cpp
eivindjahren: Ah, you are correct. It should be offset_type. Fixed.

lib/resdata/fortio_pybind.cpp
eivindjahren: Ah, you are correct. It should be offset_type. Fixed.

lib/include/resdata/FortIO.hpp
eivindjahren: Ah, didn't notice what I was converting here :D

lib/resdata/FortIO.cpp
eivindjahren: Yes you are correct. There was a logic_error added before rebasing. I removed it.

PR #1185, opened 2026-06-10
Fix validation of record_size in fortio_fread_buffer

eivindjahren:


lib/resdata/FortIO.cpp
```diff
@@ -494,60 +489,33 @@ bool fortio_complete_read(fortio_type *fortio, int record_size) {
...
+        if (end - itr < static_cast<ptrdiff_t>(record_size))
+            return false;
+        size_t items_read = 0;
+        if (record_size > 0)
+            items_read = fread(itr, 1, record_size, fortio->stream);
+        if (items_read != static_cast<size_t>(record_size) ||
+            !fortio_complete_read(fortio, record_size))
+            return false;
+    }
+    return itr == end;
```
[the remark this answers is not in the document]
eivindjahren: This is not the case as encountering a zero-length record still consumes  the four bytes from the stream. If the stream is not progressed, then fortio_init_read returns -1 and you get an early exit.

lib/resdata/FortIO.cpp
```diff
@@ -1,12 +1,14 @@
...
+#include <cerrno>
+
+#include <ios>
 #include <stdexcept>
-#include <stdbool.h>
-#include <stdlib.h>
-#include <stdio.h>
-#include <string.h>
-#include <errno.h>
+#include <string>
```
ajaust, line 9: Not sure if these include are used yet.

lib/resdata/FortIO.cpp
```diff
@@ -494,60 +489,33 @@ bool fortio_complete_read(fortio_type *fortio, int record_size) {
...
-            if (total_bytes_read >= buffer_size)
-                break;
-        }
-    }
-
-    if (total_bytes_read == buffer_size)
-        return true;
-
-    if (total_bytes_read < buffer_size)
+    if (buffer == nullptr && buffer_size != 0)
```
ajaust, line 499: Just for understanding. If we have no buffer (`buffer == nullptr`) and we pass a buffer size of `0`, then we go to the loop, do nothing and return `true`. Is this intended? What is the meaning of "successfully read 0 bytes into `nullptr`" in the context of the library?

lib/resdata/FortIO.cpp
```diff
@@ -494,60 +489,33 @@ bool fortio_complete_read(fortio_type *fortio, int record_size) {
...
-
-    util_abort("%s: internal inconsistency: buffer_size:%d  read %d bytes \n",
-               __func__, buffer_size, total_bytes_read);
-    return false;
+    if (buffer_size < 0)
+        return false;
+    char *end = buffer + buffer_size;
+    char *itr = buffer;
+    for (int record_size = 0; itr < end; itr += record_size) {
+        record_size = fortio_init_read(fortio);
```
ajaust: In general correct and might be consistent with the rest of the code (I did not check), but it might be unexpected for a `for`-loop that we mix loop variable, aborting condition etc. in a somewhat rare way. Loop variable `record_size` is not the one being increased or checked.

If we consider this a risk that could lead to misunderstandings, we could use a while loop 

```c++
while (itr < end) {
  int record_size = fortio_init_read(fortio);
  //code
  itr += record_size;
}
```

lib/resdata/FortIO.cpp
eivindjahren: Not quite, we still progress the stream past the markers. So we successfully read the b"\x00\x00\x00\x00\x00\x00\x00\x00" markers and read 0 bytes into nullptr.

lib/resdata/FortIO.cpp
eivindjahren: Yeah, its somewhat weird. I think the while loop is indeed better.

eivindjahren: I orginally had 

```cpp
int record_size = 0;
for (;itr < end; itr += record_size) {
...
}
```

but the `record_size` is available outside the scope.

lib/resdata/FortIO.cpp
eivindjahren: Reported as used by include-what-you-use

ajaust: I see. These are used in parts of the code that were not change so they did not appear in the diff.

ajaust (approved):

Issue #1186, opened 2026-06-11
openFortIO swallows all exceptions


eivindjahren:
The following will not raise a ValueError as openFortIO swallows it.

```python
    with openFortIO(str(filepath), fmt_file=False) as f:
        raise ValueError()
```

This is very surprising behavior and should be changed for version 7.

Code here: https://github.com/equinor/resdata/blob/main/python/resdata/resfile/fortio.py#L38-L63

eivindjahren:
Problem is the implementation of

```python
class FortIOContextManager:
    def __init__(self, fortio):
        self.__fortio = fortio

    def __enter__(self):
        return self.__fortio

    def __exit__(self, exc_type, exc_val, exc_tb):
        self.__fortio.close()
        return exc_type is not None
```

What should happen is that close is called on `__exit__` only, however, exc_type & exc_val, exc_tb is mostly ignored here which then ends up with swallowing the exception. By using `contextlib.contextmanager` to ensure that close is called on `self.__fortio` instead, then the exception should be correctly handled.

PR #1187, opened 2026-06-11
Ensure that tests raise

eivindjahren:
The tests for ResdataKW.fread were raising assertion errors, but this was swallowed by the context manager openFortIO due to https://github.com/equinor/resdata/issues/1186 . This PR fixes the tests and ensures that they can raise by avoiding the context manager.

### DIFF
--- a/tests/rd_tests/test_rd_kw.py
+++ b/tests/rd_tests/test_rd_kw.py
@@ -759,59 +759,76 @@ def test_read_grdecl_float_strict_raises_on_malformed(tmp_path):
             ResdataKW.read_grdecl(fh, "FLTKW", rd_type=ResDataType.RD_FLOAT)
 
 
-def test_that_short_record_size_returns_none(tmp_path):
+def read_kw_from_bytes(tmp_path, kw_bytes: bytes) -> ResdataKW:
     filepath = tmp_path / "test.bin"
-    filepath.write_bytes(
+    filepath.write_bytes(kw_bytes)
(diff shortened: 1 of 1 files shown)


tests/rd_tests/test_rd_kw.py
```diff
@@ -759,59 +759,76 @@ def test_read_grdecl_float_strict_raises_on_malformed(tmp_path):
...
 
-def test_that_oversized_record_size_returns_none(tmp_path):
-    filepath = tmp_path / "test.bin"
-    filepath.write_bytes(
+def test_that_single_element_keyword_can_be_read(tmp_path):
+    kw = read_kw_from_bytes(
+        tmp_path,
         b"\x00\x00\x00\x10KEYWORD1\x00\x00\x00\x01INTE\x00\x00\x00\x10"
-        b"\x00\x00\x00\x0f" + b"\x00" * 1000
+        b"\x00\x00\x00\x04\x00\x00\x00\x01\x00\x00\x00\x04",
```
ajaust, line 781: Is there an explanation why the data is changed as well?

ajaust (approved):

tests/rd_tests/test_rd_kw.py
eivindjahren: Yes. I assumed that a zero array would have a zero record, but there is at least an assumption in the rest of the code that zero length arrays have no zero records.

PR #1188, opened 2026-06-11
Fix an unintended change to fortio_fread_buffer

eivindjahren:
In eb8e61a6aa26e571c6633e186c794aa4eea28e3f, fortio_fread_buffer was changed to not read zero record markers when buffer_size=0. This assumption is consistent with its use in rd_kw_fread_data, but not consistent with the existing behavior.

This commit introduces more c tests for fortio and undos that change in behavior.

### DIFF
--- a/lib/resdata/FortIO.cpp
+++ b/lib/resdata/FortIO.cpp
@@ -502,7 +502,7 @@ bool fortio_fread_buffer(fortio_type *fortio, char *buffer, int buffer_size) {
         return false;
     char *end = buffer + buffer_size;
     char *itr = buffer;
-    while (itr < end) {
+    do {
         int record_size = fortio_init_read(fortio);
         if (record_size < 0)
             return false;
@@ -515,7 +515,7 @@ bool fortio_fread_buffer(fortio_type *fortio, char *buffer, int buffer_size) {
             !fortio_complete_read(fortio, record_size))
             return false;
         itr += record_size;
-    }
+    } while (itr < end);
     return itr == end;
 }
 
--- a/lib/tests/test_fortio.cpp
+++ b/lib/tests/test_fortio.cpp
@@ -1,0 +1,190 @@
+#include <array>
+#include <catch2/catch_test_macros.hpp>
+#include <catch2/generators/catch_generators.hpp>
+
+#include <cstddef>
+#include <cstdint>
+#include <filesystem>
+#include <fstream>
+#include <ios>
+#include <stdexcept>
+#include <vector>
+
+#include <resdata/FortIO.hpp>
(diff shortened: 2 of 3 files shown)


lib/tests/test_fortio.cpp
```diff
@@ -0,0 +1,191 @@
...
+#include <resdata/FortIO.hpp>
+#include <resdata/rd_endian_flip.hpp>
+#include <ert/util/util.hpp>
+#include <ert/util/vector.hpp>
+#include <vector>
+
+#include "tmpdir.hpp"
+
+TEST_CASE_METHOD(Tmpdir, "Basic FortIO operations") {
+    GIVEN("An out fortio instance") {
```
ajaust: Is this `GIVEN` statement misleading? It says "An **out** fortio instance"., but from what I understand it is creating instances for pure `out`, `appending` and `in`?

```suggestion
    GIVEN("A fortio instance opened in various modes") {
```

lib/tests/test_fortio.cpp
```diff
@@ -0,0 +1,191 @@
...
+#include <cstdint>
+#include <filesystem>
+#include <fstream>
+#include <ios>
+#include <stdexcept>
+
+#include <resdata/FortIO.hpp>
+#include <resdata/rd_endian_flip.hpp>
+#include <ert/util/util.hpp>
+#include <ert/util/vector.hpp>
```
ajaust: Is this `ert/util/vector.hpp` used anywhere?

lib/tests/test_fortio.cpp
```diff
@@ -0,0 +1,191 @@
...
+            REQUIRE(buffer[1] == 'B');
+            REQUIRE(buffer[2] == 'B');
+        }
+
+        THEN("read_buffer will fail if not enough data in records") {
+            std::array<char, 4> buffer = {0, 0, 0, 0};
+            REQUIRE_FALSE(fortio_fread_buffer(fortio.get(), buffer.data(), 4));
+        }
+
+        THEN("read_buffer will fail if records overshoot") {
```
ajaust: Somebody who reads this test the next time might not immediately know what "overshoot" means in this context
```suggestion
        THEN("read_buffer will fail if record exceeds remaining buffer size") {
```

lib/resdata/FortIO.cpp
ajaust, line 1: The commit message has a typo

`undos` -> `undoes`

ajaust (approved):
Nice! I added some small suggestions.

lib/tests/test_fortio.cpp
eivindjahren: No, I removed it. It was used in a test that is now removed.

PR #1191, opened 2026-06-15
Check cwrap types when calling via pybind

eivindjahren:
Adds `TypeError` for calls to methods that would lead to undefined behavior.

In order for this to not be a breaking change, we have to be sure that the previous behavior was not defined, at least not in documentation.

The simplest case is that whenever `from_cwrap<T>(self)` is called, somehow inserting anything other than the instance as self (is that even possible?) would not be supported.

Here are some of the less simple cases:

In the call to _rd_sum._fread_alloc, `data_files` is a local variable, so this is unproblematic. Similarly, the following have also always been unproblematic:
* `_rd_sum._init_pandas_frame`, `keywords`
* `_rd_sum._init_pandas_frame_interp`, `keywords`
* `_rd_sum._init_pandas_frame_interp`, `time_points`
* `_rd_sum._dump_csv_line`, `key_words`
* `_rd_sum._export_csv`, `var_list`
* `_grid._fprintf_grdecl2`, `output_unit`
* `_grid._fprintf_GRID2`, `output_unit`
* `_grid._fprintf_EGRID2`, `output_unit`
* `_grid._compressed_kw_copy`, `kw_copy`
* `_grid._global_kw_copy`, `kw_copy`
* `_kw._fprintf_grdecl`, `file`
* `_grid._fwrite_grdecl`, `file`
* `_grid._fprintf_grdecl2`, `file`
* `_rd_sum._dump_csv_line`, `file`
* `_grid._equal`, `other`

For the following function, parameter combinations the following holds: In the call to the function, giving None or an invalid BaseCClass would result in UB. After this change all UB cases are `TypeError`.

* `_rd_sum._resample`, `time_points`
* `_grid._grdecl_create`, `zcorn`
* `_grid._grdecl_create`, `coord`
* `_grid._fwrite_grdecl`, `kw`
* `_grid._load_column`, `kw`
* `_grid._load_column`, `column`
* `_grid._fprintf_grdecl2`, `kw`
* `_kw._load_grdecl`, `cfile`
* `_kw._load_grdecl`, `kw`

The following would give a different exception if given None, but potentially a UB if a different BaseCClass. We have now introduced a `TypeError` instead.

* `_grid._compressed_kw_copy`, `kw`
* `_grid._global_kw_copy`, `kw`
* `_grid._export_data_as_int`, `kw`
* `_grid._export_data_as_double`, `kw`

For `_grid._grdecl_create, `actnum` and `_grid._grdecl_create, `mapaxes`, the only difference is that these could always be None. However, we have now introduced a `TypeError` if not the correct BaseCClass.

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -163,7 +163,7 @@ install(DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/include/
 # -----------------------------------------------------------------
 set(PYBIND11_FINDPYTHON ON)
 find_package(pybind11 CONFIG REQUIRED)
-pybind11_add_module(_grid resdata/rd_grid_pybind.cpp)
+pybind11_add_module(_grid resdata/rd_grid_pybind.cpp resdata/cwrap_pybind.cpp)
 target_link_libraries(_grid PRIVATE resdata fmt::fmt)
 target_include_directories(
   _grid
@@ -193,7 +193,7 @@ set_target_properties(
 # -----------------------------------------------------------------
 # Pybind11 extension module: resdata.resfile._kw
 # -----------------------------------------------------------------
-pybind11_add_module(_kw resdata/rd_kw_pybind.cpp)
+pybind11_add_module(_kw resdata/rd_kw_pybind.cpp resdata/cwrap_pybind.cpp)
 target_link_libraries(_kw PRIVATE resdata fmt::fmt)
 target_include_directories(
   _kw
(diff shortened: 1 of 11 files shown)


lib/CMakeLists.txt
```diff
@@ -192,8 +195,9 @@ set_target_properties(
 # -----------------------------------------------------------------
 pybind11_add_module(_kw resdata/rd_kw_pybind.cpp)
 target_link_libraries(_kw PRIVATE resdata fmt::fmt)
-target_include_directories(_kw PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/include
-                                       ${CMAKE_CURRENT_BINARY_DIR}/include)
+target_include_directories(
+  _kw PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/include
+              ${CMAKE_CURRENT_SOURCE_DIR}/private-include)
```
ajaust: Is the removal of `${CMAKE_CURRENT_BINARY_DIR}/include` from the include directories done on purpose?

It compiles so I am not too worried, but I am still surprised that it was removed for only one of the modules.

ajaust (approved):
Nice changes! I only have a single point that might need validation (about changing some `target_include_directories`).

lib/CMakeLists.txt
eivindjahren: Fixed!

PR #1192, opened 2026-06-16
Fix stateful rd_kw test

eivindjahren:


tests/rd_tests/test_rd_kw.py
```diff
@@ -1015,6 +1015,13 @@ def create_numeric_kw(self, res_data, file_format):
 
         npt.assert_allclose(kw.numpy_view(), res_data[1], rtol=1e-2, atol=1e-6)
 
+        # For formatted files, we get truncated numbers which
+        # will causes a difference which may eventually exceed tolerance.
```
ajaust:
```suggestion
        # will cause a difference which may eventually exceed tolerance.
```

tests/rd_tests/test_rd_kw.py
```diff
@@ -1015,6 +1015,13 @@ def create_numeric_kw(self, res_data, file_format):
...
+        # Therefore, the truncated values are assigned to the model values.
+        match file_format:
+            case resfo.Format.FORMATTED:
+                res_data[1][:] = kw.numpy_view()
```
ajaust, line 1022: I am wondering if we use the `match`-`case` syntax for a specific reason. We only have a single case so I expected something like `if file_format ==  resfo.Format.FORMATTED:`.

ajaust (approved):
Nice. Added two minor remarks.

tests/rd_tests/test_rd_kw.py
eivindjahren: True, I just find it nicer to use the match case, but it is equivalent.

PR #1193, opened 2026-06-16
Merge ERT::FortIO and fortio_type

eivindjahren:
Note that this changes how failure to open files are handled. It goes from returning a nullptr and then checking for it on the python side. Now that is raised from the c++ side.


lib/resdata/fortio_pybind.cpp
```diff
@@ -1,73 +1,143 @@
...
+        "       if kw.name() == 'PRESSURE':\n"
+        "          kw.fwrite(fortio)\n"
+        "\n"
+        "   fortio.close()\n"
+        "\n"
+        "See the documentation of openFortIO() for an alternative\n"
+        "method based on a context manager and the with statement.\n");
+
+    cls.def(py::init([](std::string &file_name, Mode mode, bool fmt_file,
+                        bool endian_flip_header) {
```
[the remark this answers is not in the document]
eivindjahren: This worked fine in tests.

lib/resdata/fortio_pybind.cpp
```diff
@@ -1,73 +1,143 @@
...
+                if (mode == Mode::READ_MODE)
+                    iosmode = std::ios_base::in;
+                else if (mode == Mode::WRITE_MODE)
+                    iosmode = std::ios_base::out;
+                else if (mode == Mode::READ_AND_WRITE_MODE)
+                    iosmode = std::ios_base::in | std::ios_base::out;
+                else if (mode == Mode::APPEND_MODE)
+                    iosmode = std::ios_base::app;
+                else
+                    throw py::value_error(fmt::format("Unknown mode {}", mode));
```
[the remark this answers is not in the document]
eivindjahren: Compiled just fine.

lib/include/resdata/FortIO.hpp
```diff
@@ -20,64 +18,97 @@ typedef enum {
...
+    static bool looks_like_fortran_file(const char *filename, bool endian_flip);
     void open(const std::string &filename, std::ios_base::openmode mode,
               bool fmt_file = false, bool endian_flip_header = RD_ENDIAN_FLIP);
-    void fflush() const;
-    bool ftruncate(offset_type new_size);
-
-    fortio_type *get() const;
     void close();
 
+    [[nodiscard]] FortIO *get() const;
```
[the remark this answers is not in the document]
eivindjahren: removed

lib/resdata/tests/rd_fortio.cpp
```diff
@@ -9,53 +14,46 @@
...
-        fortio_fclose(fortio);
-    } else
-        test_assert_NULL(fortio);
+    bool raised = false;
+    try {
+        ERT::FortIO fortio(filename, std::ios_base::out);
+    } catch (std::exception e) {
+        raised = true;
+    }
+    test_assert_true(raised || !path_exists);
```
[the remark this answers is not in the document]
eivindjahren: Fixed

lib/resdata/rd_kw_pybind.cpp
```diff
@@ -9,11 +13,18 @@
...
+
+#include <ert/util/int_vector.hpp>
 
 #include <detail/resdata/cwrap_pybind.hpp>
 
 namespace py = pybind11;
 
+extern "C" {
+rd_data_type *rd_type_alloc_copy_python(const rd_data_type *src_type);
+}
```
ajaust, line 26: Not a blocker:
This is somewhat fragile due to the declaration as `extern C`. We could define it in some header, e.g, `rd_type.hpp` or `cwrap_pybind.hpp`, and the implementation in an appropriate `cpp` file. This would allow the compiler to catch signature mismatches.

2It might even work to leave the headers untouched and to add the declaration in `cwrap_pybind.cpp` assuming that `cwrap_pybind` is compiled/linked into all pybind modules.

lib/resdata/rd_file_pybind.cpp
```diff
@@ -0,0 +1,96 @@
...
+namespace py = pybind11;
+
+namespace {
+PYBIND11_MODULE(_file, m) {
+    m.doc() = "pybind11 bindings between rd_file.py and rd_file.cpp";
+
+    m.def("_open", [](std::string filename, int flags) -> py::object {
+        auto *file = rd_file_open(filename.c_str(), flags);
+        if (file == nullptr)
+            return py::none();
```
ajaust, line 28: The PR description says that it's the goal to raise exceptions from the C++ side. This one still returns `None`. I assume that this is due to how `rd_file_open` works internally and throwing an exception would need (too much) extra effort to make it work. Is this correct?

lib/tests/test_rd_grid_nnc_lgr.cpp
```diff
@@ -2,6 +2,7 @@
 #include <catch2/catch_test_macros.hpp>
 #include <catch2/matchers/catch_matchers.hpp>
 #include <catch2/matchers/catch_matchers_floating_point.hpp>
+#include <ios>
```
ajaust, line 5: Good change, but the commit message is slightly misleading. The commit does more than changing the docstring since it also implements the RAII pattern in `lib/tests/test_rd_grid_nnc_lgr.cpp`.

lib/resdata/FortIO.cpp
```diff
@@ -1,26 +1,21 @@
...
 #include <cstddef>
 #include <cstdlib>
 #include <cstdio>
 #include <cstring>
 #include <cerrno>
-
-#include <ios>
-#include <stdexcept>
-#include <string>
+#include <cstdio>
```
ajaust: We have this include already in line 9.

lib/private-include/detail/resdata/cwrap_pybind.hpp
```diff
@@ -6,3 +6,15 @@
...
+inline void register_exceptions(pybind11::module &m) {
+    pybind11::register_exception_translator([](std::exception_ptr p) {
+        try {
+            if (p)
+                std::rethrow_exception(p);
+        } catch (const std::ios_base::failure &e) {
+            PyErr_SetString(PyExc_OSError, e.what());
+        }
+    });
+}
```
ajaust, line 20: I am curious here. Why do we register a our own exception handler?

lib/resdata/FortIO.cpp
```diff
@@ -167,43 +123,92 @@ FortIO::FortIO(const std::string &filename, std::ios_base::openmode mode,
...
+    m_read_size = 0;
+    strcpy(m_opts, endian_flip_header ? "c" : "ce");
+    m_stream = stream;
+}
+
+FortIO::~FortIO() { close(); }
+
+FortIO::FortIO(FortIO &&other) noexcept { *this = std::move(other); }
+
+FortIO &FortIO::operator=(FortIO &&other) noexcept {
```
ajaust, line 150: There might be some fragility in here. Not 100% sure though.

Everything is fine since members of `FortIO` are default initialized (which is the case now). However, things may become weird when we add a member without default initialization. In this case, `this` contains uninitialized members and depending on what `close()` does, things may got wrong. I just want to make sure that we have considered this.

We could initialize the members explicitly in the move constructor (line 140) to avoid this. This still needs us to extend  the initialiser list when we add new members.

```c++
FortIO::FortIO(FortIO &&other) noexcept
    : m_stream(nullptr)
    , m_stream_owner(false)
    , m_writable(false)
    , m_read_size(0)
    , m_endian_flip_header(false)
    , m_fmt_file(false)
    , m_fopen_mode(nullptr)
    , m_opts{0, 0, 0}
 {
    std::swap(m_stream, other.m_stream);
    std::swap(m_filename, other.m_filename);
    std::swap(m_endian_flip_header, other.m_endian_flip_header);
    std::swap(m_fmt_file, other.m_fmt_file);
    std::swap(m_fopen_mode, other.m_fopen_mode);
    std::swap(m_stream_owner, other.m_stream_owner);
    std::swap(m_writable, other.m_writable);
    std::swap(m_read_size, other.m_read_size);
    std::swap(m_opts, other.m_opts);  // or memcpy + zero

    // Or instead of all the swaps do
    // *this = std::move(other);
}
```

ajaust (approved):
Nice changes. I have some minor comments or found things that I find interesting to understand better. 🙂 There might be some fragility in the code, but not sure if this is an issue (more fragile than everything is anyway). 😆

lib/resdata/rd_file_pybind.cpp
eivindjahren: This is just to not cause any changes in behavior here. You can see that the None value is checked an causes an OSError:

https://github.com/equinor/resdata/blob/8737dd2e569e4e92f915a7530d8fa22b9c8745d2/python/resdata/resfile/rd_file.py#L188

I think further down the line we will raise an OSError directly.

lib/resdata/rd_kw_pybind.cpp
eivindjahren: True, but let's maybe do the simple fix for now and move it separately.

lib/tests/test_rd_grid_nnc_lgr.cpp
eivindjahren: Moved it to its own commit.

lib/private-include/detail/resdata/cwrap_pybind.hpp
eivindjahren: There is no default handler that translates to OSError which is what we want (and the existing API uses). So we set one up that translates std::ios_base::failure to OSError.

lib/resdata/FortIO.cpp
eivindjahren: After talking about this we ended up fixing this with std::exchange.

PR #1194, opened 2026-06-18
Translate fortio tests to catch2

eivindjahren:


lib/tests/test_fortio.cpp
```diff
@@ -85,15 +102,24 @@ std::vector<std::byte> concat(const Buffers &...buffers) {
     return result;
 }
 
+template <typename T> const char *as_char(const T *p) {
+    return reinterpret_cast<const char *>(p);
+};
```
ajaust: Trailing `;`
```suggestion
}
```

ajaust (approved):
Nice!

lib/tests/test_fortio.cpp
eivindjahren: Fixed

PR #1196, opened 2026-06-19
Fix rd_grid_alloc_mapaxes_kw not working correctly with missing mapaxes and fix lazy reading of formatted summary files

### LINKED ISSUE #901: EclSum does not handled formatted ecl files correctly
Using the following FSMSPEC and FUNSMRY files respectively
```
 'INTEHEAD'           2 'INTE'
           1         100
 'RESTART '           9 'CHAR'
 'AAAAAAAA' '        ' '        ' '        ' '        ' '        ' '        '
 '        ' '        '
 'DIMENS  '           6 'INTE'
           3           1           1           1           0           1
 'KEYWORDS'           3 'CHAR'
 'TIME    ' 'AAQR    ' 'AAQR    '
 'WGNAMES '           3 'CHAR'
 ':+:+:+:+' 'AAAAAAAA' 'AAAAAAAA'
 'NUMS    '           3 'INTE'
      -32676           0           0
 'UNITS   '           3 'CHAR'
 'DAYS    ' 'SM3/DAY ' 'SM3/DAY '
 'STARTDAT'           6 'INTE'
           1           1        1971           0           0           0

```
```
 'INTEHEAD'           2 'INTE'
           1         100
 'RESTART '           9 'CHAR'
 'AAAAAAAA' '        ' '        ' '        ' '        ' '        ' '        '
 '        ' '        '
 'DIMENS  '           6 'INTE'
           3           1           1           1           0           1
 'KEYWORDS'           3 'CHAR'
 'TIME    ' 'AAQR    ' 'AAQR    '
 'WGNAMES '           3 'CHAR'
 ':+:+:+:+' 'AAAAAAAA' 'AAAAAAAA'
 'NUMS    '           3 'INTE'
      -32676           0           0
 'UNITS   '           3 'CHAR'
 'DAYS    ' 'SM3/DAY ' 'SM3/DAY '
 'STARTDAT'           6 'INTE'
           1           1        1999           0           0           0

```
When loaded with EclSum I get
```python
eclsum = EclSum("TEST")
print([v.value() for v in eclsum.alloc_time_vector(False)])
```
I would expect `[915148800, 915235200, 915321600]` to be printed, but instead I get `[3784234291200, 3784234291200, 3784234291200]`

### LINKED ISSUE #1117: rd_grid_alloc_mapaxes_kw returns inconsistent rd_kw when mapaxes is not used

eivindjahren:
Resolves #1117
Resolves #901

### DIFF
--- a/lib/include/resdata/rd_grid.hpp
+++ b/lib/include/resdata/rd_grid.hpp
@@ -1,17 +1,18 @@
-#ifndef ERT_RD_GRID_H
-#define ERT_RD_GRID_H
+#pragma once
+
+#include <cstdio>
+#include <memory>
+#include <filesystem>
 
 #include <ert/util/double_vector.hpp>
 #include <ert/util/int_vector.hpp>
 #include <ert/util/stringlist.hpp>
+#include <ert/util/type_macros.hpp>
 
 #include <resdata/rd_coarse_cell.hpp>
 #include <resdata/rd_kw.hpp>
 #include <resdata/nnc_info.hpp>
-
-#ifdef __cplusplus
-extern "C" {
-#endif
+#include <resdata/rd_util.hpp>
 
 #define RD_GRID_COORD_SIZE(nx, ny) (((nx) + 1) * ((ny) + 1) * 6)
 #define RD_GRID_ZCORN_SIZE(nx, ny, nz) (((nx) * (ny) * (nz) * 8))
@@ -196,12 +197,6 @@ float rd_grid_output_scaling(const rd_grid_type *grid,
 UTIL_IS_INSTANCE_HEADER(rd_grid);
 UTIL_SAFE_CAST_HEADER(rd_grid);
 
-#ifdef __cplusplus
-}
-
-#include <memory>
-#include <filesystem>
-
 using rd_grid_ptr = std::unique_ptr<rd_grid_type, decltype(&rd_grid_free)>;
 rd_grid_ptr make_rectangular_grid(int nx, int ny, int nz, double dx, double dy,
                                   double dz, const int *actnum);
@@ -209,5 +204,3 @@ rd_grid_ptr read_grid(const std::filesystem::path &filename);
 inline rd_grid_ptr copy_grid(const rd_grid_type *src_grid) {
     return {rd_grid_alloc_copy(src_grid), &rd_grid_free};
 }
-#endif
-#endif
--- a/lib/include/resdata/rd_grid.hpp
+++ b/lib/include/resdata/rd_grid.hpp
@@ -3,6 +3,7 @@
 #include <cstdio>
 #include <memory>
 #include <filesystem>
+#include <optional>
 
 #include <ert/util/double_vector.hpp>
 #include <ert/util/int_vector.hpp>
@@ -88,7 +89,7 @@ rd_grid_type *rd_grid_alloc_rectangular(int nx, int ny, int nz, double dx,
                                         double dy, double dz,
                                         const int *actnum);
 rd_kw_type *rd_grid_alloc_volume_kw(const rd_grid_type *grid, bool active_size);
-rd_kw_type *rd_grid_alloc_mapaxes_kw(const rd_grid_type *grid);
+std::optional<rd_kw_ptr> rd_grid_alloc_mapaxes_kw(const rd_grid_type *grid);
 rd_kw_type *rd_grid_alloc_coord_kw(const rd_grid_type *grid);
 
 bool rd_grid_exists(const char *case_input);
--- a/lib/include/resdata/rd_kw.hpp
+++ b/lib/include/resdata/rd_kw.hpp
@@ -237,7 +237,16 @@ void rd_kw_fix_uninitialized(rd_kw_type *rd_kw, int nx, int ny, int nz,
 rd_type_enum rd_kw_get_type(const rd_kw_type *);
 
 using rd_kw_ptr = std::unique_ptr<rd_kw_type, decltype(&rd_kw_free)>;
-rd_kw_ptr make_rd_kw(const char *header, int size, rd_data_type data_type);
+inline rd_kw_ptr make_rd_kw(const char *header, int size,
+                            rd_data_type data_type) {
+    return {rd_kw_alloc(header, size, data_type), rd_kw_free};
+}
+
+inline rd_kw_ptr make_rd_kw(const char *header, int size,
+                            rd_data_type data_type, const void *data) {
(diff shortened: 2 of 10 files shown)


lib/include/resdata/rd_grid.hpp
```diff
@@ -88,7 +89,7 @@ rd_grid_type *rd_grid_alloc_rectangular(int nx, int ny, int nz, double dx,
                                         double dy, double dz,
                                         const int *actnum);
 rd_kw_type *rd_grid_alloc_volume_kw(const rd_grid_type *grid, bool active_size);
-rd_kw_type *rd_grid_alloc_mapaxes_kw(const rd_grid_type *grid);
+std::optional<rd_kw_ptr> rd_grid_alloc_mapaxes_kw(const rd_grid_type *grid);
```
ajaust, line 92: The commit message says 
```text
Fix rd_grid_alloc_mapaxes with for missing mapaxes
```
I think it was supposed to be
```text
Fix rd_grid_alloc_mapaxes for missing mapaxes
```

lib/include/resdata/rd_kw.hpp
```diff
@@ -237,7 +237,16 @@ void rd_kw_fix_uninitialized(rd_kw_type *rd_kw, int nx, int ny, int nz,
...
-rd_kw_ptr make_rd_kw(const char *header, int size, rd_data_type data_type);
+inline rd_kw_ptr make_rd_kw(const char *header, int size,
+                            rd_data_type data_type) {
+    return {rd_kw_alloc(header, size, data_type), rd_kw_free};
+}
+
+inline rd_kw_ptr make_rd_kw(const char *header, int size,
+                            rd_data_type data_type, const void *data) {
+    return {rd_kw_alloc_new(header, size, data_type, data), rd_kw_free};
+}
```
ajaust, line 248: This is for my understanding: 

I see that we call `rd_kw_alloc` and `rd_kw_alloc_new` which almost the same things, but `rd_allow_new` will also memcpy the data.

Why do we need the two different ways of making a `rd_kw`? As we are more leaning towards C++ now, I expected `make_rd_kw` to get a signature like this:

```c++
inline rd_kw_ptr make_rd_kw(const char *header, int size,
                            rd_data_type data_type, const void *data = nullptr) {
    return {rd_kw_alloc_new(header, size, data_type, data), rd_kw_free};
}
```
but I get that this would lead to very different behavior (memcpy_data vs non memcpy).

lib/include/resdata/rd_kw.hpp
```diff
@@ -55,7 +55,7 @@ void rd_kw_fskip(ERT::FortIO &);
 bool rd_kw_fread_realloc(rd_kw_type *, ERT::FortIO &);
 rd_kw_type *rd_kw_fread_alloc(ERT::FortIO &);
 rd_kw_type *rd_kw_alloc_actnum(const rd_kw_type *porv_kw, float porv_limit);
-void rd_kw_fread_indexed_data(ERT::FortIO &fortio, offset_type data_offset,
+void rd_kw_fread_indexed_data(ERT::FortIO &fortio, offset_type kw_offset,
```
ajaust, line 58: The commit says
```text
Fix a lazy-loading formatted summaries
```
but should maybe be
```text
Fix lazy-loading of formatted summaries
```

lib/resdata/rd_kw.cpp
```diff
@@ -1202,31 +1202,57 @@ static bool rd_kw_fread_data(rd_kw_type *rd_kw, ERT::FortIO &fortio) {
...
+    // on-disk size, so we can seek directly to each requested element. For
+    // formatted (ASCII) files the elements have a variable text width and
+    // direct seeking is not possible; in that case we read the whole
+    // keyword and extract the requested elements afterwards.
+    if (fortio.fmt_file()) {
+        fortio.fseek(kw_offset, SEEK_SET);
+        rd_kw_ptr rd_kw(rd_kw_fread_alloc(fortio), rd_kw_free);
+        if (rd_kw == NULL)
+            util_abort("%s: failed to load keyword at offset:%ld\n", __func__,
+                       (long)kw_offset);
```
ajaust, line 1227: Are we sure that we want to introduce new `util_abort` calls? I think this is a problem for users calling using the Python interface as they will not be able to catch this as exception.

Could we throw an exception
```suggestion
            throw std::ios_base::failure(
                fmt::format("failed to load keyword at offset:{}", kw_offset));
```
instead?

lib/resdata/rd_kw.cpp
```diff
@@ -1202,31 +1202,57 @@ static bool rd_kw_fread_data(rd_kw_type *rd_kw, ERT::FortIO &fortio) {
...
+            int element_index = int_vector_iget(index_map, index);
 
-        fortio.data_fseek(data_offset, element_index, sizeof_iotype,
-                          element_count, block_size);
-        util_fread(&io_buffer[index * sizeof_iotype], sizeof_iotype, 1, stream,
-                   __func__);
-    }
+            if (element_index < 0 || element_index >= element_count)
+                util_abort("%s: Element index is out of range 0 <= %d < %d\n",
+                           __func__, element_index, element_count);
```
ajaust, line 1244: I don't have too strong feelings about this `util_abort` since that one is preexisting.

ajaust (approved):
Looks good overall. 👍 

I have some remarks that should be straighforward to fix. I have some concerns about introducing new calls to `util_abort`, but I see that this happens in a part of the code where another `util_abort` exists anyway.

What do we do about this PR
https://github.com/equinor/resdata/pull/1134

Will say "thank you" and then close it after your PR is merged?

lib/include/resdata/rd_kw.hpp
eivindjahren: Yes, lets refactor this separately.

lib/resdata/rd_kw.cpp
eivindjahren: Unfortunately called through code that does not clean up if an exception is thrown so I would rather fix that first before introducing an exception.

PR #1200, opened 2026-06-22
Pybind well info

eivindjahren:


tests/test_well.py
```diff
@@ -0,0 +1,784 @@
+"""Tests for the ``resdata.well`` Python bindings.
+
+The sepecification for the restart file format can be found
```
ajaust:
```suggestion
The specification for the restart file format can be found
```

tests/test_well.py
```diff
@@ -0,0 +1,784 @@
...
+NICONZ = 25  # elements per connection in ICON
+NCWMAX = 5  # max connections per well
+NSCONZ = 25  # elements per connection in SCON
+NXCONZ = 60  # elements per connection in XCON
+NISEGZ = 22  # elements per segment in ISEG
+NRSEGZ = 140  # elements per segment in RSEG
+NSEGMX = 10  # max segments per well
+NSWLMX = 2  # max segmented wells
+NLBRMX = 2  # max lateral branches per well
+NILBRZ = 10  # elements per branch in ILBR
```
ajaust: Do we use the same nomenclature as OPM Flow? The OPM flow uses the word "values" instead of "elements". For example, for `IWELZ` the documentation of OPM Flow says

> Number of values per well in the IWEL array (default value is 97).

tests/test_well.py
```diff
@@ -0,0 +1,784 @@
...
+    well = Well(name="W1", well_type=iwel_type, connections=[Connection(1, 1, 1)])
+    path = str(tmp_path / "CASE.X0000")
+    write_restart(path, [well])
+
+    well_state = WellInfo(grid, path)["W1"][0]
+
+    assert well_state.wellType() == expected_type
+
+
+def test_that_a_closed_well_is_reported_as_shut(tmp_path, grid):
```
ajaust: How accurate do these test names need to be? 
In my opinion, the test verifies that the well is reported as "not open". That could mean many things to me as I don't know what states a well can have besides "open". We do not have an explicit test checking if a condition referring to "shut" is true.

tests/test_well.py
```diff
@@ -0,0 +1,784 @@
...
+    assert connections[1].connectionFactor() == 2.0
+
+
+def test_that_well_head_returns_the_global_connection_head(tmp_path, grid, producer):
+    path = str(tmp_path / "CASE.X0000")
+    write_restart(path, [producer])
+
+    well_head = WellInfo(grid, path)["OP1"][0].wellHead()
+
+    # IWEL stores the head as (2, 3, 1) using 1-based indices.
```
ajaust, line 523: It says that the head is stored as `(2, 3, 1)`, but we check for `(1, 2, 0)`. Are we using zero-based indexing?

lib/resdata/well_info.cpp
```diff
@@ -25,10 +25,10 @@
 #include <resdata/well/well_ts.hpp>
 
 /*
-  The library libwell contains functionality to read and interpret
+  This file contains functionality to read and interpret
   some of the well related keywords from a restart
   file. Roughly speaking the implementation is spread between three
-  six datatypes; users of the libwell library are mainly conserned
+  six datatypes; users of the interface are mainly conserned
```
ajaust: Not your fault, but now I cannot unsee it after reading the comment. 
```suggestion
  six datatypes; users of the interface are mainly concerned
```

lib/resdata/well_info.cpp
```diff
@@ -237,6 +237,24 @@ static void well_info_add_state(well_info_type *well_info,
    be discarded afterwards.
 */
 
+namespace {
+struct close_guard {
```
ajaust, line 241: Commit message suggestion:

"Use RAII to handled (un)set of RD_FILE_CLOSE_STREAM" -> "Use RAII to handle (un)set of RD_FILE_CLOSE_STREAM"

lib/resdata/well_info.cpp
```diff
@@ -237,6 +237,24 @@ static void well_info_add_state(well_info_type *well_info,
...
+    explicit close_guard(rd_file_view_type *file_view)
```
ajaust, line 242: I like this pattern. 🙂

ajaust (approved):
Looks nice. I left some (minor) comments.

tests/test_well.py
eivindjahren: Fixed

tests/test_well.py
eivindjahren: I went with `test_that_status_0_means_closed_well`

tests/test_well.py
eivindjahren: Yes, there is a conversion.

lib/resdata/well_info.cpp
eivindjahren: Fixed

PR #1202, opened 2026-06-24
Use pybind for rsthead

eivindjahren:


lib/include/resdata/rd_rsthead.hpp
```diff
@@ -59,25 +57,104 @@ typedef struct {
...
+                   const rd_kw_type *doubhead_kw, const rd_kw_type *logihead_kw)
+        : report_step(report_step),
+          sim_days(rd_kw_iget_double(doubhead_kw, DOUBHEAD_DAYS_INDEX)) {
+        const int *data = (const int *)rd_kw_get_void_ptr(intehead_kw);
+
+        this->day = data[INTEHEAD_DAY_INDEX];
+        this->month = data[INTEHEAD_MONTH_INDEX];
+        this->year = data[INTEHEAD_YEAR_INDEX];
+        this->version = data[INTEHEAD_IPROG_INDEX];
+        this->phase_sum = data[INTEHEAD_PHASE_INDEX];
```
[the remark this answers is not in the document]
eivindjahren: I ended up adding this.

lib/include/resdata/rd_rsthead.hpp
```diff
@@ -59,25 +57,104 @@ typedef struct {
...
+        this->nzwelz = data[INTEHEAD_NZWELZ_INDEX];
+        this->nsconz = data[INTEHEAD_NSCONZ_INDEX];
+        this->nxconz = data[INTEHEAD_NXCONZ_INDEX];
+        this->niconz = data[INTEHEAD_NICONZ_INDEX];
+        this->ncwmax = data[INTEHEAD_NCWMAX_INDEX];
+
+        this->nisegz = data[INTEHEAD_NISEGZ_INDEX];
+        this->nsegmx = data[INTEHEAD_NSEGMX_INDEX];
+        this->nswlmx = data[INTEHEAD_NSWLMX_INDEX];
+        this->nrsegz = data[INTEHEAD_NRSEGZ_INDEX];
```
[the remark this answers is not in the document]
eivindjahren: I ended up adding this as well.

tests/rd_tests/test_restart_head.py
```diff
@@ -57,6 +58,63 @@ def test_restart_headers():
...
+    doubhead = ResdataKW("DOUBHEAD", 100, ResDataType.RD_DOUBLE)
+    logihead = ResdataKW("LOGIHEAD", 100, ResDataType.RD_BOOL)
+    header = ResdataRestartHead(kw_arg=(1, intehead, doubhead, logihead))
+    return header.get_sim_date()
+
+
+@given(
+    year=st.integers(min_value=1970, max_value=2600),
+    month=st.integers(min_value=1, max_value=12),
+    day=st.integers(min_value=1, max_value=27),
```
ajaust, line 80: Why do we limit days to 27? Shouldn't 28 be a safe limit? In general, we should be sure that 31 is a valid input (for some months) as well.

tests/rd_tests/test_restart_head.py
```diff
@@ -57,6 +58,63 @@ def test_restart_headers():
...
+    intehead[65] = month
+    intehead[66] = year
+    doubhead = ResdataKW("DOUBHEAD", 100, ResDataType.RD_DOUBLE)
+    logihead = ResdataKW("LOGIHEAD", 100, ResDataType.RD_BOOL)
+    header = ResdataRestartHead(kw_arg=(1, intehead, doubhead, logihead))
+    return header.get_sim_date()
+
+
+@given(
+    year=st.integers(min_value=1970, max_value=2600),
```
ajaust, line 78: Is `1970` early enough as start date? How old could the "oldest" data we expect be?

lib/resdata/tests/well_segment_conn_load.cpp
ajaust, line 1: Is there now value in this test so removing makes more sense than adding it to the built tests?

lib/include/resdata/rd_rsthead.hpp
```diff
@@ -59,25 +57,118 @@ typedef struct {
...
+
+        this->nisegz = get(INTEHEAD_NISEGZ_INDEX);
+        this->nsegmx = get(INTEHEAD_NSEGMX_INDEX);
+        this->nswlmx = get(INTEHEAD_NSWLMX_INDEX);
+        this->nlbrmx = get(INTEHEAD_NLBRMX_INDEX);
+        this->nilbrz = get(INTEHEAD_NILBRZ_INDEX);
+        this->nrsegz = get(INTEHEAD_NRSEGZ_INDEX);
+
+        this->sim_time = rd_make_date(this->day, this->month, this->year);
+        this->sim_days = rd_kw_iget_double(doubhead_kw, DOUBHEAD_DAYS_INDEX);
```
ajaust: `sim_days` was already initialized in the initializer list.

lib/include/resdata/rd_rsthead.hpp
```diff
@@ -59,25 +57,118 @@ typedef struct {
...
+        if (rd_file_view_has_kw(rst_view, SEQNUM_KW)) {
+            const rd_kw_type *seqnum_kw =
+                rd_file_view_iget_named_kw(rst_view, SEQNUM_KW, 0);
+            report_step = rd_kw_iget_int(seqnum_kw, 0);
+        }
+
+        return {report_step, intehead_kw, doubhead_kw, logihead_kw};
+    }
+
+    inline bool operator==(const RSTHead &other) const {
```
ajaust, line 146: The parameters

- `report_step`
- `unit_system`
- `nxwelz`
- `nsconz`
- `nxconz`
- `nrsegz`

seem to be ignored for the equality comparison. This is preexisting behavior, but a bit surprising to me. I guess it makes sense.

lib/resdata/rd_file_view.cpp
```diff
@@ -470,6 +478,16 @@ bool rd_file_view_has_report_step(const rd_file_view_type *rd_file_view,
         return false;
 }
 
+static time_t rsthead_date(int day, int month, int year) {
+    return rd_make_date(day, month, year);
+}
+
+time_t rd_rsthead_date(const rd_kw_type *intehead_kw) {
```
ajaust: We could consider making this method `static` as well (same as `rsthead_date`) to indicate that this has internal linkage only (or use an anonymous namespace, but `static` is the predominant style).

ajaust (approved):
Looks good. 👍 
I added some minor comments and suggestions.

tests/rd_tests/test_restart_head.py
eivindjahren: It should be, but I wanted to avoid February on leap years.

tests/rd_tests/test_restart_head.py
eivindjahren: Turns out we get an exception on dates prior to 1970 on mac.

lib/resdata/tests/well_segment_conn_load.cpp
eivindjahren: The test did not compile and used internal test data so I didn't see any value in it.

lib/include/resdata/rd_rsthead.hpp
eivindjahren: Yes, surprised me as well, but I didn't want to change the behavior now, as the operator is not used on the python side and only in some tests.

lib/resdata/rd_file_view.cpp
eivindjahren: Done.

lib/include/resdata/rd_rsthead.hpp
eivindjahren: Fixed

PR #1208, opened 2026-06-24
Remove manual memory management in rd_subsidence

eivindjahren:


python/resdata/gravimetry/rd_subsidence.py
```diff
@@ -171,7 +169,7 @@ def eval(
 
         If supplied the optional argument @region should be an
         ResdataRegion() instance; this region will be used to limit the
-        part of the reserviour included in the subsidence calculations.
+        part of the reservoir included in the subsidence calculations.
 
         The argument @compressibility is the total reservoir compressibility.
         """
```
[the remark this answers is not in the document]
eivindjahren: This has been fixed.

lib/resdata/rd_subsidence.cpp
```diff
@@ -243,57 +215,43 @@ static double rd_subsidence_survey_eval_geertsma_rporv(
...
                           double poisson_ratio) {
-    rd_subsidence_survey_type *base_survey =
-        rd_subsidence_get_survey(subsidence, base);
-    rd_subsidence_survey_type *monitor_survey =
-        rd_subsidence_get_survey(subsidence, monitor);
+    rd_subsidence_survey_type *base_survey = subsidence->surveys.at(base).get();
+    rd_subsidence_survey_type *monitor_survey = nullptr;
+    if (monitor)
+        monitor_survey = subsidence->surveys.at(monitor).get();
     return rd_subsidence_survey_eval(base_survey, monitor_survey, region, utm_x,
```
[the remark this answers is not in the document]
eivindjahren: Since we already rely on this behavior I think it is unproblematic. In any case we are going to convert this to use pybind and remove the extern "C" shortly.

lib/resdata/rd_subsidence.cpp
```diff
@@ -27,110 +37,79 @@
...
-
-static rd_subsidence_survey_type *
+static std::unique_ptr<rd_subsidence_survey_type>
 rd_subsidence_survey_alloc_PRESSURE(rd_subsidence_type *rd_subsidence,
                                     const rd_file_view_type *restart_view,
                                     const char *name) {
 
-    rd_subsidence_survey_type *survey =
-        rd_subsidence_survey_alloc_empty(rd_subsidence, name);
+    std::unique_ptr<rd_subsidence_survey_type> survey(
```
ajaust: The current code has a small risk of leaking memory in case of an exception. Resizing the vectors or the allocation of the string may throw. In this case, the allocated survey leaks.

We could consider replacing `rd_subsidence_survey_alloc_empty` with a constructor in `struct rd_subsidence_survey_type`:

```c++
 rd_subsidence_survey_type(const rd_subsidence_type &sub, const char *name)
     : grid_cache(sub.grid_cache.get()),
       aquifer_cell(sub.aquifer_cell.get()),
       name(name),
       porv(sub.grid_cache->size()),
       pressure(sub.grid_cache->size()) {}
```

Then,
```c++
    std::unique_ptr<rd_subsidence_survey_type> survey(
        rd_subsidence_survey_alloc_empty(rd_subsidence, name));
```
would become
```c++
auto survey = std::make_unique<rd_subsidence_survey_type>(*rd_subsidence, name);
```

This would also remove the need for the bare `new`.

lib/include/resdata/rd_file.hpp
```diff
@@ -74,3 +78,10 @@ using rd_file_view_ptr =
     std::unique_ptr<rd_file_view_type, decltype(&rd_file_view_free)>;
 
 using rd_file_ptr = std::unique_ptr<rd_file_type, decltype(&rd_file_close)>;
+
+inline rd_file_ptr open_rd_file(std::string path, int flags) {
+    return {rd_file_open(path.c_str(), flags), &rd_file_close};
+}
+inline rd_file_ptr open_rd_file(std::filesystem::path path, int flags) {
+    return {rd_file_open(path.string().c_str(), flags), &rd_file_close};
+}
```
ajaust: We could take the path parameter by reference `const std::string&/std::filesystem::path&` if we are worried about the copy that may be made if taking the path by value?

lib/resdata/rd_subsidence.cpp
```diff
@@ -27,110 +37,79 @@
...
     const rd_file_type *
         init_file; /* The init file - a shared reference owned by calling scope. */
-    rd::rd_grid_cache *grid_cache;
-    bool *aquifer_cell;
-    hash_type *
-        surveys; /* A hash table containg rd_subsidence_survey_type instances; one instance
-                                          for each interesting time. */
-    double *compressibility; /*total compressibility*/
-    double *poisson_ratio;
+    std::unique_ptr<rd::rd_grid_cache> grid_cache{nullptr};
```
ajaust, line 43: Commit message:
The last sentence ends quite abrupt with "So we inline and only allow it to be nullptr;". Should this be something like "So we inline get_survey and only allow the monitor parameter to be nullptr"?

ajaust (approved):
Looks great! I added few, minor comments, but I think we also have a potential memory leak in case of exceptions being thrown.

lib/resdata/rd_subsidence.cpp
eivindjahren: Fixed!

lib/include/resdata/rd_file.hpp
eivindjahren: Done

lib/resdata/rd_subsidence.cpp
eivindjahren: Correct! I changed it to your suggestion.

PR #1210, opened 2026-06-25
Openfortio swallows all exceptions

### LINKED ISSUE #1186: openFortIO swallows all exceptions

kaamyashinde:
**Issue**
Resolves #1186 

**Approach**
Error raised inside the `with openFortIO(...)` block was being swallowed because of the old context manager.

This was rewrote using Python's `contextlib.contextmanager` helper. A test for this behaviour was also introduced.

### DIFF
--- a/python/resdata/resfile/fortio.py
+++ b/python/resdata/resfile/fortio.py
@@ -19,20 +19,18 @@ structure; mainly to support passing FortIO handles to the underlying cpp
 functions.
 """
 
+from contextlib import contextmanager
+
 from resdata.resfile._fortio import FortIO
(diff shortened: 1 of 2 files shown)


eivindjahren (approved):
👍  🎉  🌮

PR #1214, opened 2026-06-26
Use pybind for subsidence bindings

### LINKED ISSUE #1205: convert rd_subsidence.py to use pybind instead of ResdataPrototype

MagnusSletten:
**Issue**
Resolves #1205 

**Approach**
Adds a new pybind module for subsidence and updates `rd_subsidence.py` to use it instead of defining the native bindings through `ResdataPrototype`.


eivindjahren (approved):
🎉

PR #1216, opened 2026-06-29
Use pybind for gravimetry bindings

### LINKED ISSUE #1204: convert rd_grav.py to use pybind instead of ResdataPrototype

MagnusSletten:
Replace ResdataPrototype calls and let gravimetry errors propagate as Python exceptions.

**Issue**
Resolves #1204 

**Approach**
Added a pybind module for the gravimetry bindings and updated ResdataGrav to call this module instead of using ResdataPrototype.

Converted relevant aborting error paths to exceptions so Python callers can handle them normally.

### DIFF
--- a/lib/include/resdata/rd_grav.hpp
+++ b/lib/include/resdata/rd_grav.hpp
@@ -1,41 +1,47 @@
 #ifndef ERT_RD_GRAV_H
 #define ERT_RD_GRAV_H
 
+#include <optional>
+#include <string>
+
 #include <resdata/rd_file.hpp>
 #include <resdata/rd_file_view.hpp>
 #include <resdata/rd_grid.hpp>
 #include <resdata/rd_region.hpp>
 
-#ifdef __cplusplus
-extern "C" {
-#endif
 typedef struct rd_grav_struct rd_grav_type;
 typedef struct rd_grav_survey_struct rd_grav_survey_type;
 
 void rd_grav_free(rd_grav_type *rd_grav_config);
+
 rd_grav_type *rd_grav_alloc(rd_grid_type *rd_grid,
                             const rd_file_type *init_file);
+
 rd_grav_survey_type *
-rd_grav_add_survey_FIP(rd_grav_type *grav, const char *name,
+rd_grav_add_survey_FIP(rd_grav_type *grav, const std::string &name,
                        const rd_file_view_type *restart_file);
+
 rd_grav_survey_type *
-rd_grav_add_survey_PORMOD(rd_grav_type *grav, const char *name,
+rd_grav_add_survey_PORMOD(rd_grav_type *grav, const std::string &name,
                           const rd_file_view_type *restart_file);
+
 rd_grav_survey_type *
-rd_grav_add_survey_RPORV(rd_grav_type *grav, const char *name,
+rd_grav_add_survey_RPORV(rd_grav_type *grav, const std::string &name,
                          const rd_file_view_type *restart_file);
+
 rd_grav_survey_type *
-rd_grav_add_survey_RFIP(rd_grav_type *grav, const char *name,
+rd_grav_add_survey_RFIP(rd_grav_type *grav, const std::string &name,
                         const rd_file_view_type *restart_file);
-double rd_grav_eval(const rd_grav_type *grav, const char *base,
-                    const char *monitor, rd_region_type *region, double utm_x,
-                    double utm_y, double depth, int phase_mask);
+
+double rd_grav_eval(const rd_grav_type *grav, const std::string &base,
+                    const std::optional<std::string> &monitor,
+                    rd_region_type *region, double utm_x, double utm_y,
+                    double depth, int phase_mask);
+
 void rd_grav_new_std_density(rd_grav_type *grav, rd_phase_enum phase,
                              double default_density);
+
 void rd_grav_add_std_density(rd_grav_type *grav, rd_phase_enum phase,
                              int pvtnum, double density);
 
-#ifdef __cplusplus
-}
-#endif
 #endif
--- a/lib/resdata/cwrap_pybind.cpp
+++ b/lib/resdata/cwrap_pybind.cpp
@@ -6,6 +6,7 @@
 
 #include <resdata/rd_kw.hpp>
 #include <resdata/rd_grid.hpp>
+#include <resdata/rd_grav.hpp>
 #include <resdata/rd_sum.hpp>
 #include <resdata/rd_file.hpp>
 #include <resdata/well/well_info.hpp>
@@ -277,6 +278,14 @@ py::object WellState() {
     return cls;
 }
 
+py::object ResdataGrav() {
+    static py::object cls;
+    if (!cls) {
+        cls = py::module_::import("resdata.gravimetry").attr("ResdataGrav");
+    }
+    return cls;
+}
+
 template <> well_state_type *from_cwrap<well_state_type>(py::handle obj) {
     if (!py::isinstance(obj, WellState()))
         throw py::type_error("Expected WellState, got " +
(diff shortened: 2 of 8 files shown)


lib/resdata/rd_grav_pybind.cpp
```diff
@@ -0,0 +1,114 @@
...
+        auto *init_file_ptr = from_cwrap<rd_file_type>(init_file);
+
+        auto *grav = rd_grav_alloc(grid_ptr, init_file_ptr);
+
+        if (grav == nullptr) {
+            return py::none();
+        }
+
+        return py::int_(reinterpret_cast<uintptr_t>(grav));
+    });
```
eivindjahren, line 37: So this does `py::int_` here and `py::cast` in `_add_survey_RFIP`. Likely we should choose one style and stick with it :)

lib/resdata/rd_grav.cpp
```diff
@@ -465,8 +466,9 @@ rd_grav_survey_alloc_RPORV(rd_grav_type *rd_grav,
         for (iactive = 0; iactive < rd_kw_get_size(rporv_kw); iactive++)
             survey->porv[iactive] = rd_kw_iget_as_double(rporv_kw, iactive);
     } else
-        util_abort("%s: restart file did not contain %s keyword??\n", __func__,
-                   RPORV_KW);
+        throw std::runtime_error(std::string(__func__) +
```
eivindjahren, line 469: I think it would be a good idea to add tests for the new exceptions that are raised to see that they work.

eivindjahren (changes requested):
I had some small suggestions. But mainly I think some tests for the new exceptions would be good :)

tests/rd_tests/test_grav.py
```diff
@@ -134,3 +134,23 @@ def test_create_minimal(self):
...
+                "RPORV",
+            ]
+        ]
+
+        with TestAreaContext("grav_invalid_phase"):
+            write_kws("TEST", kws)
+            init = ResdataFile("TEST.INIT")
+            grav = ResdataGrav(self.grid, init)
+
+            with self.assertRaises(ValueError):
```
eivindjahren: pytest.raises(ValueError, match=...)

lib/resdata/rd_util.cpp
```diff
@@ -51,9 +51,9 @@ const char *rd_get_phase_name(rd_phase_enum phase) {
         return RD_PHASE_NAME_GAS;
         break;
     default:
-        util_abort("%s: phase enum value:%d not recognized \n", __func__,
-                   phase);
-        return NULL;
+        throw std::invalid_argument(
+            std::string(__func__) + ": phase enum value: " +
+            std::to_string(static_cast<int>(phase)) + " not recognized");
```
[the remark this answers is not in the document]
eivindjahren: `#include <stdexcept>`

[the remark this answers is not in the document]
eivindjahren: `#include <string>`

lib/include/resdata/rd_grav.hpp
```diff
@@ -1,41 +1,47 @@
 #ifndef ERT_RD_GRAV_H
 #define ERT_RD_GRAV_H
 
+#include <optional>
```
ajaust, line 4: I cannot comment on commit messages so I comment on this file instead:

Please check the commit messages for consistency. We try to follow the rules described [in this blog post](https://chris.beams.io/git-commit), e.g., headlines/subject lines of 50 chars and body wrapped at 72 chars, no trailing punctuation in headers etc. This helps keeping the messages easily readable in all developer environments (terminal, VS Code etc.) and across projects.

python/resdata/gravimetry/rd_grav.py
```diff
@@ -89,21 +68,21 @@ def add_survey_RPORV(self, survey_name: str, restart_view: ResdataFileView) -> N
...
-           restart_view1 = rst_file.restart_view(sim_time=date)
-           restart_view2 = rst_file.restart_view(report_step=67)
+            import datetime
+            from resdata.resfil import ResdataRestartFile
+            ...
+            ...
+            date = datetime.datetime(year, month, day)
+            rst_file = ResdataFile("CASE.UNRST")
+            restart_view1 = rst_file.restart_view(sim_time=date)
+            restart_view2 = rst_file.restart_view(report_step=67)
```
ajaust, line 78: Nice formatting fix. 👍

lib/resdata/cwrap_pybind.cpp
```diff
@@ -274,40 +274,44 @@ py::object WellState() {
...
+            }
+            return cls;
+            template <> rd_grav_type *from_cwrap<rd_grav_type>(py::handle obj) {
+                if (!py::isinstance(obj, ResdataGrav()))
+                    throw py::type_error(
+                        "Expected ResdataGrav, got " +
+                        static_cast<std::string>(py::repr(obj)));
+
+                return cast_cwrap<rd_grav_type>(obj);
+            }
```
ajaust: Double check that this compiles. To me, it looks like we are missing some closing braces `}`. The tests are currently failing which is probably related to the missing braces.

ajaust (commented):
Nice changes. I had just a minor comment w.r.t. commit message formatting. 

I had a hint on how to fix some C++ issue, but this one is already fixed. There is still a comment for this which I cannot delete, but just ignore it. I will mark it as resolved.

lib/resdata/cwrap_pybind.cpp
ajaust: This seems to be fixed. 🙂

lib/include/resdata/rd_grav.hpp
MagnusSletten: Fixed 👍

lib/resdata/rd_grav.cpp
MagnusSletten: Tests have been added covering them 👍

tests/rd_tests/test_grav.py
```diff
@@ -134,3 +134,111 @@ def test_create_minimal(self):
...
+            grav = ResdataGrav(self.grid, init)
+
+            restart_file = ResdataFile("TEST.UNRST")
+            restart_view = restart_file.restart_view(sim_time=datetime.date(2000, 1, 1))
+
+            grav.new_std_density(1, 0.5)
+            grav.add_std_density(1, 0, 0.5)
+
+            with self.assertRaisesRegex(RuntimeError, "restart file did not contain"):
+                grav.add_survey_RPORV("rporv", restart_view)
```
ajaust: I think the comment of @eivindjahren about `pytest.raises(ValueError, match=...)` disappeared. I believe the comment was about using `pytest.raises` over `self.assertRaisesRegex`. The latter seems to be the approach of the `unittest` package. Using `pytest.raises` should be more canonical since we are using `pytest` (or at least try to use it), see [pytest docs on assertions](https://docs.pytest.org/en/7.1.x/how-to/assert.html).

The assertions using `with self.assertRaisesRegex` should probably become something like
```suggestion
            with pytest.raises(RuntimeError, match=r".* restart file did not contain .*")
                grav.add_survey_RPORV("rporv", restart_view)
```
You need to check this. I don't know this by heart myself.

This will require an `import pytest` in the current file if `pytest` is not imported yet.

ajaust (approved):
From my side it looks good. All my comments are addressed. 👍 

I think one comment by Eivind got lost and was not explicitly addressed by a comment nor changes. I added Please have a look if this is something that should still be changed.

ajaust (approved):
Usage of `pytest.raise` looks good. 👍

lib/resdata/rd_grav_pybind.cpp
MagnusSletten: I chose py::int_ here 👍

eivindjahren (approved):

PR #1219, opened 2026-06-30
Remove util abort in rd_kw

eivindjahren:


lib/resdata/rd_grid.cpp
```diff
@@ -4306,11 +4306,11 @@ double rd_grid_get_property(const rd_grid_type *rd_grid,
             return -1; /* Tried to lookup an inactive cell. */
 
     } else {
-        std::unique_ptr<char> type_name(rd_type_alloc_name(data_type));
+        std::string type_name = rd_type_name(data_type);
         throw std::invalid_argument(fmt::format(
             "rd_grid_get_property: can not lookup type:{} - keyword type must "
             "be numeric.",
-            type_name.get()));
+            type_name.c_str()));
```
ajaust: Is the `c_str()` call required? I would have expected that `fmt::format` accepts `std::string`.

lib/resdata/rd_kw.cpp
```diff
@@ -195,18 +197,18 @@ static int get_columns(const rd_data_type data_type) {
     case (RD_STRING_TYPE):
         return COLUMNS_CHAR; // TODO: Is this correct?
     default:
-        util_abort("%s: invalid rd_type: %s\n", __func__,
-                   rd_type_alloc_name(data_type));
+        throw std::invalid_argument(
+            fmt::format("invalid rd_type: {}", rd_type_name(data_type)));
         return -1;
```
ajaust: Do we still need this `return` statement? It should be unreachable now due to the exception that we throw.

lib/tests/test_rd_kw.cpp
```diff
@@ -0,0 +1,335 @@
...
+    SECTION("icmp_string on non char type") {
+        REQUIRE_THROWS_WITH(rd_kw_icmp_string(int_kw.get(), 0, "X"),
+                            ContainsSubstring("is wrong type"));
+    }
+}
+
+TEST_CASE("rd_kw_iset_string_ptr validates type and length", "[rd_kw]") {
+    SECTION("non alphabetic type") {
+        auto int_kw = make_int_kw("KW", 1);
+        REQUIRE_THROWS_WITH(rd_kw_iset_string_ptr(int_kw.get(), 0, "x"),
```
ajaust, line 100: Whay is `"x"` a non-alphabetic type? I thought `x` would be a valid letter.

lib/tests/test_rd_kw.cpp
```diff
@@ -0,0 +1,335 @@
...
+    REQUIRE_THROWS_WITH(rd_kw_shift_int(float_kw.get(), 1),
+                        ContainsSubstring("is wrong type"));
+
+    auto int_kw = make_int_kw("KW", 3);
+    REQUIRE_THROWS_WITH(rd_kw_scalar_set_float_or_double(int_kw.get(), 1.0),
+                        ContainsSubstring("wrong type"));
+    REQUIRE_THROWS_WITH(rd_kw_scale_float_or_double(int_kw.get(), 1.0),
+                        ContainsSubstring("wrong type"));
+    REQUIRE_THROWS_WITH(rd_kw_shift_float_or_double(int_kw.get(), 1.0),
+                        ContainsSubstring("wrong type"));
```
ajaust, line 126: Is it correct that the expected substrings vary? It is sometimes "wrong type" and other times "is wrong type".

lib/resdata/rd_kw.cpp
ajaust: Can we use the `rd_kw_ptr rd_kw(rd_kw_alloc_empty(), rd_kw_free);` pattern here as well? There a risk of a memory leak in here now since `rd_kw_alloc_data` may throw.

ajaust, line 1461: Might this leak? Not sure if `rd_kw_fread_alloc` might throw, but it feels a bit inconsistent with the other locations where the raw pointer was replaced by `rd_kw_ptr`.

ajaust: This might leak now. I think `rd_kw_alloc_data` and `rd_kw_fread_data` may throw.

ajaust: This might also profit from using `rd_kw_ptr`.

ajaust, line 594: From what I can see this won't throw. We could use `rd_kw_ptr` here for consistency, but not sure if it makes sense.

ajaust: This might profit from `rd_kw_ptr`, but unsure if this may leak, i.e., not sure if `rd_kw_fread_header` throws.

ajaust (commented):
The changes look good overall. I have found locations where we might lack the use `rd_kw_ptr`. This might just have slipped through. For some parts, I think it is relevant because more parts of our code may throw now.

lib/tests/test_rd_kw.cpp
eivindjahren: The problem isn't that the "x" is not alphabetic its that the keyword has type int.

lib/tests/test_rd_kw.cpp
eivindjahren: Oups, that was just an error.

lib/resdata/rd_kw.cpp
eivindjahren: Seems inconsistent, but rd_kw_fread_alloc returns the pointer so this is fine.

eivindjahren: Fixed

eivindjahren: Fixed

eivindjahren: Done

eivindjahren: A bit worried about changing this, but its on my todo list.

eivindjahren: Fixed

lib/resdata/rd_kw.cpp
eivindjahren: Removed it

lib/resdata/rd_grid.cpp
eivindjahren: Not required, I removed it.

lib/resdata/rd_kw.cpp
ajaust: Not crucial, but `rd_kw_alloc_new_shared` seems to still use `rd_kw_ptr`.

ajaust (approved):
Looks good. 🎉 

There is one occasion where the change to `rd_kw_ptr` might have been lost during the fixes. Just have a look if you agree.

PR #1220, opened 2026-06-30
 Fix null keyword segfault after rd_kw_fread_alloc

kaamyashinde:
rd_file_kw loads keyword data lazily from disk. When rd_kw_fread_alloc fails, file_kw->kw is NULL, but the code still validated and registered the keyword — causing a segfault.

This PR skips that work when load fails, and throws std::invalid_argument if an unloaded keyword is written in place. A regression test covers the write path.

### DIFF
--- a/lib/resdata/rd_file_kw.cpp
+++ b/lib/resdata/rd_file_kw.cpp
@@ -164,6 +164,11 @@ bool rd_file_kw_equal(const rd_file_kw_type *kw1, const rd_file_kw_type *kw2) {
 }
 
 static void rd_file_kw_assert_kw(const rd_file_kw_type *file_kw) {
+    if (file_kw->kw == NULL)
+        throw std::runtime_error(
+            "rd_file_kw: keyword could not be loaded from file "
+            "(rd_kw_fread_alloc returned NULL)");
+
     if (!rd_type_is_equal(rd_file_kw_get_data_type(file_kw),
                           rd_kw_get_data_type(file_kw->kw)))
         util_abort("%s: type mismatch between header and file.\n", __func__);
@@ -197,10 +202,8 @@ static void rd_file_kw_load_kw(rd_file_kw_type *file_kw, ERT::FortIO &fortio,
     {
         fortio.fseek(file_kw->file_offset, SEEK_SET);
         file_kw->kw = rd_kw_fread_alloc(fortio);
-        if (file_kw->kw != NULL) {
-            rd_file_kw_assert_kw(file_kw);
-            inv_map_add_kw(inv_map, file_kw, file_kw->kw);
-        }
+        rd_file_kw_assert_kw(file_kw);
+        inv_map_add_kw(inv_map, file_kw, file_kw->kw);
     }
 }
 
@@ -273,9 +276,6 @@ bool rd_file_kw_fskip_data(const rd_file_kw_type *file_kw,
 */
 
 void rd_file_kw_inplace_fwrite(rd_file_kw_type *file_kw, ERT::FortIO &fortio) {
-    if (file_kw->kw == NULL)
-        throw std::invalid_argument(
-            "rd_file_kw: cannot write keyword that is not loaded");
     rd_file_kw_assert_kw(file_kw);
     fortio.fseek(file_kw->file_offset, SEEK_SET);
     rd_kw_fskip_header(fortio);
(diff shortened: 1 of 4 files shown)


lib/resdata/tests/rd_file_view.cpp
```diff
@@ -94,8 +96,29 @@ void test_create_file_kw() {
     rd_file_kw_free(file_kw2);
 }
 
+void test_unloaded_file_kw_is_rejected() {
+    rd_file_kw_type *file_kw = rd_file_kw_alloc0("TEST_KW", RD_INT, 10, 0);
+
```
eivindjahren: We should probably document this somewhere, but we have moved away from this style of test. Could you instead add this to a new test file in `lib/tests` which uses the catch2 testing framework?

lib/resdata/tests/rd_file_view.cpp
```diff
@@ -94,8 +96,29 @@ void test_create_file_kw() {
...
+    bool threw = false;
+    try {
+        rd_file_kw_inplace_fwrite(file_kw, fortio);
+    } catch (const std::invalid_argument &) {
+        threw = true;
+    }
+    test_assert_true(threw);
+
+    rd_file_kw_free(file_kw);
```
eivindjahren: We want to avoid manual memory management like this in the tests as when the tests fail we get a false-positive from the address-sanitizer. You just need to use a unique_ptr:

You just have to declare file_kw like this:

```cpp
std::unique_ptr<rd_file_kw_type, decltype(&rd_file_kw_free)> file_kw(rd_file_kw_alloc0("TEST_KW", RD_INT, 10, 0), &rd_file_kw_free);
```

in order to get the raw pointer, you just call `.get()`:

```cpp
rd_file_kw_inplace_fwrite(file_kw.get(), fortio);
```

lib/resdata/tests/rd_file_view.cpp
```diff
@@ -94,8 +96,29 @@ void test_create_file_kw() {
     rd_file_kw_free(file_kw2);
 }
 
+void test_unloaded_file_kw_is_rejected() {
+    rd_file_kw_type *file_kw = rd_file_kw_alloc0("TEST_KW", RD_INT, 10, 0);
+
+    rd::util::TestArea ta("reject_unloaded");
+    FILE *fp = util_fopen("dummy", "w");
```
eivindjahren: FILE is a part of the c standard libary and has a few caveats with error handling. I suggest using the following c++ standard library methods to write instead:

```cpp
{
    std::ofstream ofs("dummy");
    REQUIRE(ofs); // catch2 uses REQUIRE for assert statements, checks that the file was created.
} // ofs is destructed at the end of this scope and guaranteed to exist after this
ERT::FortIO fortio("dummy", std::ios_base::in, false, true);
```

lib/resdata/rd_file_kw.cpp
```diff
@@ -196,8 +197,10 @@ static void rd_file_kw_load_kw(rd_file_kw_type *file_kw, ERT::FortIO &fortio,
     {
         fortio.fseek(file_kw->file_offset, SEEK_SET);
         file_kw->kw = rd_kw_fread_alloc(fortio);
-        rd_file_kw_assert_kw(file_kw);
-        inv_map_add_kw(inv_map, file_kw, file_kw->kw);
+        if (file_kw->kw != NULL) {
+            rd_file_kw_assert_kw(file_kw);
```
eivindjahren: I think if `rd_kw_fread_alloc` fails, then the file is invalid and we cannot continue. I think it would be best if we added a throw statement inside `rd_file_kw_assert_kw` when file_kw->kw == NULL instead.

lib/tests/test_rd_file_kw.cpp
```diff
@@ -0,0 +1,29 @@
...
+
+    auto filename = (dirname / "dummy").string();
+    {
+        std::ofstream ofs(filename);
+        REQUIRE(ofs);
+    }
+
+    ERT::FortIO fortio(filename, std::ios_base::in, false, true);
+
+    REQUIRE_THROWS_AS(rd_file_kw_inplace_fwrite(file_kw.get(), fortio),
```
eivindjahren, line 27: Test looks good!

eivindjahren (approved):
:+1:

PR #1224, opened 2026-07-01
Small cleanup

eivindjahren:


lib/CMakeLists.txt
```diff
@@ -156,285 +156,57 @@ install(DIRECTORY include/ DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
...
-# When running tests against the build tree (editable install), put the
-# extension directly next to rd_well_state.py so it can be imported as
-# resdata.well._well_state without an install step.
-set_target_properties(
-  _well_state PROPERTIES LIBRARY_OUTPUT_DIRECTORY
-                         ${PROJECT_SOURCE_DIR}/python/resdata/well)
+  install(TARGETS ${module_name}
+          LIBRARY DESTINATION ${${module_name}_install_dir})
+  # When running tests against the build tree (editable install), put the
+  # extension directly next to rd${module_name}.py so it can be imported.
```
ajaust: This might be slightly inaccurate since not all modules have a Python name that is prefixed with `rd_`. For example, there is a `well_info.py`. The same seems to apply to `fortio` and `well_state`. `rd_sum` is an odd case since the `rd_` prefix is actually part of the module name so the comment implies the existence of a `rd_rd_sum.py` Python file. 

Maybe just rephrase it to be less specific, e.g., something like

```text
# When running tests against the build tree (editable install), put the
# extension in the source package directory so it can be imported without
# installing.
```

ajaust (approved):
Looks good. I found only a tiny thing in a comment.

lib/CMakeLists.txt
eivindjahren: Done!

PR #1229, opened 2026-07-06
Set backward-cpp to use backtrace_symbol on clang

eivindjahren:
This was set for macos, however, when clang is used on linux, the same option has to be set.


berland (approved):
LGTM

Issue #1197, opened 2026-06-22
Version 7 roadmap


eivindjahren:
Current main branch of resdata will eventually become version 7.0.0.

Major overarching changes:
----------------------

* Removed all code that is giving a deprecation warning.
* Removed all datatypes in resdata.util.util, this includes
    *  `DoubleVector`
    * `IntVector`
    *  `TimeVector`
    * `StringHash`
    * `DoubleHash`
    * `IntegerHash`
    * `StringList`
    *  `BoolVector`
    * `LookupTable`
    * `ThreadPool`
    The signature for functions that used these types as parameter or return type has changed to use numpy or builtin types.
    See type-hints for specifics. These functions include:
    * `resdata.summary.Summary::keys` (now returns `list[str]` instead of `StringList`)
    * `resdata.summary.Summary::blocked_production` (now returns a `numpy.vector` instead of `DoubleVector`)
    * `resdata.summary.Summary::report_index_list` (now returns a `list[int]` instead of `IntVector`)

* Passing incorrect types will now raise `TypeError` (would sometimes raise `ValueError` before).
* Passing integers that are too large for int32/int64, will raise TypeError rather than wrap.
* bool is not longer automatically converted. This applies to any method that eventually called cwrap.
* `__repr__` results no longer contain address.

Invalid:
```python
grid = Grid(f"{case}.EGRID", apply_mapaxes="yes") # Any object would be converted to bool in version 6
```

Valid:
```python
grid = Grid(f"{case}.EGRID", apply_mapaxes=True) # Version 7 only accepts bool
```

Changes to individual components:
---------------------------------
* Removed the `WellSegment.__len__` function as it broke the expectation that len() would return an int.
* `__repr__` of WellInfo no longer uses "[no name]" when name is "".

* FortIO now requires that the members `FortIO.READ_MODE`, `FortIO.WRITE_MODE`, `FortIO.APPEND_MODE`, `FortIO.READ_AND_WRITE_MODE` is used instead:

invalid:
```python
fort_io = FortIO(summary_path, mode=2)
```

valid:
```python
fort_io = FortIO(summary_path, mode=FortIO.WRITE_MODE)
```
* All dropped `WellConnections` now write a message stderr
* `ResdataFile.iget_restart_sim_days` now raises IndexError when the index is too large rather than returning -1.
* ResdataFile.has_sim_time(t) now returns True when find_sim_time(t) is not None. Before
there was a discrepancy between `has_sim_time` and `find_sim_timeq where
[...]

PR #1234, opened 2026-07-06
Remove all deprecations

eivindjahren:
See https://github.com/equinor/resdata/issues/1197


python/resdata/summary/_date2num.py
```diff
@@ -0,0 +1,36 @@
...
+# implementation could be replaced with:
+#
+#   from matplotlib.dates import date2num
+HOURS_PER_DAY = 24.0
+MINUTES_PER_DAY = 60 * HOURS_PER_DAY
+SECONDS_PER_DAY = 60 * MINUTES_PER_DAY
+MUSECONDS_PER_DAY = 1e6 * SECONDS_PER_DAY
+
+
+def date2num(dt):
```
SAKavli, line 13: Don't we typecheck in resdata for the Python parts?

SAKavli (approved):
🧠

python/resdata/summary/_date2num.py
eivindjahren: Not yet!

PR #1236, opened 2026-07-06
Replace util_abort with throw in rd_file_kw.cpp

### LINKED ISSUE #1207: convert rd_file_kw_assert_kw to throw instead of util_abort
This depends on

* #1206 
* #1205
* #1204  
* #1202
* #1200 

as we need to make sure the thrown exceptions convert to the python side and the calling scope cleans up when exceptions are thrown.

kaamyashinde:
**Issue**
Resolves #1207

**Approach**
Replace util_abort with throw in rd_file_kw.cpp


eivindjahren (approved):
:+1: (Please squash)

PR #1237, opened 2026-07-06
Fix a bug with swapped arguments to well_conn

eivindjahren:
This changes the direction of the well head from dirX (1 converted
from bool true) to dirZ as intented.

Also increases coverage of the well submodule.

### DIFF
--- a/lib/resdata/well_conn.cpp
+++ b/lib/resdata/well_conn.cpp
@@ -259,12 +259,11 @@ well_conn_type *well_conn_alloc_wellhead(const rd_kw_type *iwel_kw,
 
         if (matrix_connection)
             return well_conn_alloc(conn_i, conn_j, conn_k, connection_factor,
-                                   (well_conn_dir_enum)open, well_conn_dirZ,
-                                   header.unit_system);
+                                   well_conn_dirZ, open, header.unit_system);
         else
-            return well_conn_alloc_fracture(
-                conn_i, conn_j, conn_k, connection_factor,
-                (well_conn_dir_enum)open, well_conn_dirZ, header.unit_system);
+            return well_conn_alloc_fracture(conn_i, conn_j, conn_k,
+                                            connection_factor, well_conn_dirZ,
+                                            open, header.unit_system);
     } else
         // The well is completed in this LGR - however the wellhead is in another LGR.
         return NULL;
--- a/lib/resdata/well_ts.cpp
+++ b/lib/resdata/well_ts.cpp
@@ -108,16 +108,6 @@ const char *well_ts_get_name(const well_ts_type *well_ts) {
 void well_ts_add_well(well_ts_type *well_ts, well_state_type *well_state) {
     well_node_type *new_node = well_node_alloc(well_state);
     well_ts->ts.push_back(new_node);
-
-    if (well_ts->ts.size() > 1) {
-        const well_node_type *last_node = well_ts->ts.back();
-        if (new_node->sim_time < last_node->sim_time)
-            // The new node is chronologically before the previous node;
-            // i.e. we must sort the nodes in time. This should probably happen
-            // quite seldom:
-            std::sort(well_ts->ts.begin(), well_ts->ts.end(),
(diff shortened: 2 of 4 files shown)


lib/resdata/well_ts.cpp
```diff
@@ -108,16 +108,6 @@ const char *well_ts_get_name(const well_ts_type *well_ts) {
 void well_ts_add_well(well_ts_type *well_ts, well_state_type *well_state) {
     well_node_type *new_node = well_node_alloc(well_state);
     well_ts->ts.push_back(new_node);
-
-    if (well_ts->ts.size() > 1) {
-        const well_node_type *last_node = well_ts->ts.back();
-        if (new_node->sim_time < last_node->sim_time)
```
SAKavli, line 114: What is meant by a "well node"?

tests/test_well.py
```diff
@@ -704,21 +705,25 @@ def test_that_simulation_time_matches_the_restart_report_date(tmp_path, grid, pr
     assert (sim_date.year, sim_date.month, sim_date.day) == (2020, 6, 29)
 
 
-def test_that_unified_restart_steps_have_increasing_simulation_times(tmp_path, grid):
+def test_that_states_are_in_the_order_of_the_file(tmp_path, grid):
```
SAKavli, line 721: I am a little confused, the commit message suggests the sorting is dead code, aka it should be sorted already, while the asserts here asserts the order in the file is preserved (unsorted).

lib/resdata/well_conn.cpp
```diff
@@ -259,12 +259,11 @@ well_conn_type *well_conn_alloc_wellhead(const rd_kw_type *iwel_kw,
 
         if (matrix_connection)
             return well_conn_alloc(conn_i, conn_j, conn_k, connection_factor,
-                                   (well_conn_dir_enum)open, well_conn_dirZ,
-                                   header.unit_system);
+                                   well_conn_dirZ, open, header.unit_system);
         else
-            return well_conn_alloc_fracture(
-                conn_i, conn_j, conn_k, connection_factor,
-                (well_conn_dir_enum)open, well_conn_dirZ, header.unit_system);
```
SAKavli, line 267: Could you explain this syntax for me?
`(well_conn_dir_enum)open`

tests/test_well.py
```diff
@@ -583,6 +601,38 @@ def test_that_well_head_returns_the_global_connection_head(tmp_path, grid, produ
...
 
 
+def test_that_the_well_head_direction_is_z(tmp_path, grid, producer):
+    path = str(tmp_path / "CASE.X0000")
+    write_restart(path, [producer])
+
+    well_head = WellInfo(grid, path)["OP1"][0].wellHead()
+
+    # The wellhead is a matrix connection (well_conn_alloc). Its direction is
+    # always vertical (Z).
```
SAKavli, line 585: Is this true for horizontal wells?

lib/resdata/well_ts.cpp
eivindjahren: This is not a great naming convention, but its what is used. Often means an element in some datastructure.

tests/test_well.py
eivindjahren: The dead code would sort it, but it never happened so the order is the same as the file. The old test was in effect not testing the right thing.

lib/resdata/well_conn.cpp
eivindjahren: That is a c style cast of a boolean to a well_conn_dir_enum it converst `true` to `1`which then becomes `well_conn_dirX`.

tests/test_well.py
eivindjahren: Yes, but it should be ok as this only concerns the well head.

SAKavli (approved):
Looks good!
If I understood you correctly, we dropped the sorting as it didn't sort in the first place?

eivindjahren:
> Looks good! If I understood you correctly, we dropped the sorting as it didn't sort in the first place?

Correct!

PR #1245, opened 2026-07-13
Add additional tests around file_view behavior

eivindjahren:
These additional tests are required to be able to safely refactor rd_file_view


MagnusSletten:
Can we add a reverse-slice test such as view[::-1]? The current tests only cover positive step values.

MagnusSletten (approved):
Looks good! The negative test can be re-added on the other PR.

PR #1248, opened 2026-07-14
Fix a bug with blockview

### LINKED ISSUE #1247: block_view2 with end_view=None doesn't work as intended
The intention behind `end_view=None` seems to be to read until the end of file, however, it ends up not updating the file_kw in the loop so that the first kw is pushed onto the list every time. Can be reproduced with

```python
with openResdataFile("test-data/local/ECLIPSE/simple/SIMPLE.EGRID") as file:
    view = file.block_view2(None, None, 0)
    print(list(view)) # Should include all keywords, but just contains "FILEHEAD" seven times.
```

eivindjahren:
**Issue**
Resolves #1247

### DIFF
--- a/python/resdata/resfile/rd_file.py
+++ b/python/resdata/resfile/rd_file.py
@@ -209,7 +209,29 @@ class ResdataFile(BaseCClass):
     def free(self):
         _file._free(self)
 
-    def block_view(self, kw, kw_index):
+    def block_view(self, kw: str, kw_index: int) -> ResdataFileView:
+        """A view of the keyword block delimited by ``kw``.
+
+        The returned :class:`ResdataFileView` contains the keywords from the
+        ``kw_index``'th occurrence of ``kw`` (inclusive) up to, but not
+        including, the next occurrence of ``kw``. In other words the file is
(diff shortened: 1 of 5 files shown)


python/resdata/resfile/rd_file.py
```diff
@@ -222,7 +244,44 @@ def block_view(self, kw, kw_index):
             "Index out of range, must be in [0, %d), was %d." % (ls, kw_index)
         )
 
-    def block_view2(self, start_kw, stop_kw, start_index):
+    def block_view2(
```
MagnusSletten, line 247: block_view1 and block_view2 is maybe not the clearest naming scheme?

MagnusSletten (approved):
Looks good! The bug is clearly addressed and the accompanying tests cover the edge cases well. The extra documentation is also nice. 

One small optional point about the python naming scheme. It's not blocking.

python/resdata/resfile/rd_file.py
eivindjahren: 100%, however, for now we should avoid making too large changes to the interface.

PR #1249, opened 2026-07-15
Fix `Summary.dates` not returning correct dates for restarted cases.

eivindjahren:
There is a per-file index and a across file index when there
are restarts. In order to correctly get the report_end, the
offset of the index_node has to be added, and the overlap between
files correctly accounted for.

report_before would return the last_report of the file, but when
files overlap, the next file is chosen in the index. Therefore,
we need it to instead return the last report step indexed.

The incorrect behavior was introduced in https://github.com/equinor/resdata/commit/e89245d605046cff9426d10c88f504a1490c0580 which also
introduced throwing an exception instead of returning
INVALID_MINISTEP_NR=-1. rd_sum_data_iget_report_step still checked
for -1, which is now removed.

I tested this with a history case (HISTORY.DATA) where STARTDAT
was 1 JAN 2017, and TSTEP was set to 4*15 days giving the dates

2017-01-16
2017-01-31
2017-02-15
2017-03-02

which we get from sum.report_dates() both before this change
and after this change.

The restarted prediction case was set to start at step 4 (2017-03-02)
and TSTEP was set to 10*15.

Before this commit, the following dates would be in sum.report_dates:

2017-01-02
2017-01-05
2017-01-10
2017-01-16
2017-01-31
2017-02-15
2017-03-02
2017-03-17
2017-04-01
2017-04-16

which seems to indicate that prediction report steps start before the
last report step of the history. After the changes in the commit we get:

2017-03-17
2017-04-01
2017-04-16
2017-05-01
2017-05-16
2017-05-31
2017-06-15
2017-06-30
2017-07-15
2017-07-30

So prediction starts 15 days after the last step of history as
expected.

I further verified this by looking directly at the file contents:
The setup used non-unifed output so HISTORY.S0004 was at days=60
and PREDICTION.S005 was at day=75, which is 2017-03-02 and 2017-03-17
respectively.

### DIFF
--- a/lib/private-include/detail/resdata/rd_sum_file_data.hpp
+++ b/lib/private-include/detail/resdata/rd_sum_file_data.hpp
@@ -122,6 +122,8 @@ public:
 
     std::pair<int, int> report_range(int report_step) const;
     bool report_step_equal(const rd_sum_file_data &other, bool strict) const;
+
+    /** Returns the last report step strictly before @end_time */
     int report_before(time_t end_time) const;
     int get_time_report(int max_internal_index, time_t *data);
     int get_data_report(int params_index, int max_internal_index, double *data,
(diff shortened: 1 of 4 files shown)


MagnusSletten (approved):
Looks good! The problems with restart-chain report indexing are fixed, and the new behaviour is tested thoroughly.

PR #1250, opened 2026-07-15
Remove dead code and fix compiler warnings

eivindjahren:


MagnusSletten (approved):
Nice! The removed code has no usages within the repository, the project builds successfully, and the build configuration update is standard so this is an easy approval.

PR #1253, opened 2026-07-17
Add validation of grid dimensions, fix loading of GRID files with LGRs and dual porosity, and make save_GRID handle formattedness in the same way as save_EGRID

eivindjahren:
The purpose of this PR was originally to just add tests for lines that will be touched by the rd_file refactor. However, some bugs were discovered along the way.

* save_GRID and save_EGRID did not have the same behavior with regards to formatted files. Arguably ok behavior undocumented behavior, but I have now documented that they both have that behavior.
* rd_grid did not validate that nx, ny, nz were positive leading to problematic behavior.
* reading a GRID file with dual porosity and LGR did not appropriately account for stride. This is now fixed so that written GRID files are read correctly.


lib/resdata/rd_grid.cpp
```diff
@@ -2235,6 +2235,11 @@ rd_grid_alloc_GRDECL_kw(int nx, int ny, int nz, const rd_kw_type *zcorn_kw,
                         const rd_kw_type *actnum_kw,    /* Can be NULL */
                         const rd_kw_type *mapaxes_kw) { /* Can be NULL */
 
+    if (nx < 0 || ny < 0 || nz < 0)
+        throw std::invalid_argument(
```
MagnusSletten, line 2239: Should zero be allowed here?

MagnusSletten (approved):
Looks good to me! The change is nicely described and sensible, the additional tests are solid.  One small question which is not a blocker.

lib/resdata/rd_grid.cpp
eivindjahren: Yes, I believe that is ok as it is just an empty grid.

PR #1258, opened 2026-07-23
Use pybind11 for smspec_node

eivindjahren:
This fixes a bug with `smspec_node::valid_type` where a `nullptr` check was done too late. It also makes the operators of `ResdataSMSPECNode` follow normal conventions for unimplemented types.


lib/resdata/smspec_node_pybind.cpp
```diff
@@ -0,0 +1,136 @@
...
+            "\n"
+            "Many of the summary keywords have an integer stored in the\n"
+            "vector NUMS as an attribute, i.e. the block properties have\n"
+            "the global index of the cell in the nums vector. If the\n"
+            "variable in question makes use of the NUMS value this "
+            "property\n"
+            "will return the value, otherwise it will return None:\n"
+            "\n"
+            "   sum.smspec_node(\"FOPT\").num     => None\n"
+            "   sum.smspec_node(\"BPR:1000\").num => 1000\n"
```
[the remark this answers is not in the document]
eivindjahren: This is not the case, but there is this weird SMSPECS_NUMS_INVALID check. We should not change this behavior, but it could return e.g. 0 if the internal field is not SMSPECS_NUMS_INVALID which the user could create through add_variable

lib/include/resdata/smspec_node.hpp
```diff
@@ -1,17 +1,11 @@
-#ifndef ERT_SMSPEC_NODE_HPP
-#define ERT_SMSPEC_NODE_HPP
-
+#pragma once
 #include <cstdio>
 
 #include <string>
```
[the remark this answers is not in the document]
eivindjahren: Fixed

lib/resdata/smspec_node_pybind.cpp
```diff
@@ -0,0 +1,124 @@
...
+                          "Class can not be instantiated directly!");
+            throw py::error_already_set();
+        }))
+        .def(py::self == py::self)
+        .def(py::self < py::self)
+        .def(py::self > py::self)
+        .def("__hash__",
+             [](const rd::smspec_node &self) {
+                 return py::hash(py::cast(self.get_gen_key1()));
+             })
```
MagnusSletten, line 30: Found locally by copilot:

"Could we avoid hashing get_gen_key1() here? Equality is based on cmp(), and those do not always represent identity using the same fields. For completion nodes, for example, cmp() compares keyword, well name and NUM, while get_gen_key1() contains IJK coordinates derived from the summary’s grid dimensions. Nodes from two summaries can therefore compare equal but have different hashes. The current test only compares nodes within one summary, so it won’t catch this. Could we hash the fields used by cmp() and add a cross-summary test, or make these objects unhashable?"

lib/resdata/rd_sum_pybind.cpp
```diff
@@ -232,11 +232,10 @@ PYBIND11_MODULE(_rd_sum, m) {
...
         "_get_var_node",
         [](py::handle self, std::string lookup_kw) {
-            return reinterpret_cast<std::uintptr_t>(
-                const_cast<rd::smspec_node *>(rd_sum_get_general_var_node(
-                    from_cwrap<rd_sum_type>(self), lookup_kw.c_str())));
+            return const_cast<rd::smspec_node *>(rd_sum_get_general_var_node(
+                from_cwrap<rd_sum_type>(self), lookup_kw.c_str()));
         },
-        py::return_value_policy::reference);
+        py::return_value_policy::reference, py::keep_alive<0, 1>());
```
MagnusSletten, line 238: Do we have a lifetime-test for this? E.g: get node, delete the original summary reference and check that the note is still accessible?

python/resdata/summary/rd_smspec_node.py
```diff
@@ -1,178 +0,0 @@
...
-class ResdataSMSPECNode(BaseCClass):
-    """
-    Small class with some meta information about a summary variable.
-
-    The summary variables have different attributes, like if they
-    represent a total quantity, a rate or a historical quantity. These
-    quantities, in addition to the underlying values like WGNAMES,
-    KEYWORD and NUMS taken from the the SMSPEC file are stored in this
-    structure.
-    """
```
MagnusSletten, line 17: Minor documentation reduction: the class level description is not present in the newer file.

MagnusSletten (commented):
Nice work! A couple of mostly minor comments. The hashing functionality is the most important to double check.

python/resdata/summary/rd_smspec_node.py
eivindjahren: Added!

lib/resdata/smspec_node.cpp
```diff
@@ -798,11 +783,17 @@ rd_smspec_var_type smspec_node::valid_type(const char *keyword,
...
+            return RD_SMSPEC_INVALID_VAR;
+
         return var_type;
     }
 
     if (var_type == RD_SMSPEC_COMPLETION_VAR ||
         var_type == RD_SMSPEC_SEGMENT_VAR) {
+        if (!wgname)
+            return RD_SMSPEC_INVALID_VAR;
+
```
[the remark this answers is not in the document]
eivindjahren: Fixed!

lib/resdata/rd_sum_pybind.cpp
eivindjahren: There is one, but its just a silent assumption in another test. This just re-implements the parent-child relationship set up in `createCReference`.

lib/resdata/smspec_node_pybind.cpp
eivindjahren: Maybe, but I hesitate to change this. It is sufficient though.

PR #1262, opened 2026-08-12
Update deprecations

eivindjahren:


python/resdata/util/util/rng.py
```diff
@@ -1,11 +1,16 @@
...
 from cwrap import BaseCClass
+from typing_extensions import deprecated
 
 from resdata import ResdataPrototype
 from resdata.util.enums import RngAlgTypeEnum, RngInitModeEnum
 
 
+@deprecated(
+    "RandomNumberGenerator is deprecated and will be removed in version 7. "
+    "Use the builtin random module instead."
```
andreas-el, line 12: Should we give an example? Other messages have hints as to how to resolve.

python/resdata/util/util/bool_vector.py
```diff
@@ -105,10 +107,18 @@ class BoolVector(VectorTemplate):
     def __init__(self, default_value=False, initial_size=0):
         super().__init__(default_value, initial_size)
 
+    @deprecated(
+        "resdata vectors are deprecated and will be removed in version 7. All functions returning vectors "
+        "will be replaced with python lists. Convert the value to list with list() or to numpy.array to silence this warning."
+    )
```
andreas-el: Can this be referenced to, similar to how `@stringlist_deprecation` is done?
There is an awful lot of identifcal deprecation text.

python/resdata/util/util/rng.py
eivindjahren: I think its easily googleable.

python/resdata/util/util/bool_vector.py
eivindjahren: Fixed

python/resdata/util/util/rng.py
andreas-el: Fair enough.

andreas-el (approved):
Well done! 🍰

PR #1264, opened 2026-08-14
Fix a regression to add_keywords

eivindjahren:
In 982b02065b3520c742d0ce24b1e62e7704f1a847 the keyword_pattern in add_keywords was assumed to be not None, however, it turns out that this can be None.


berland (commented):
👍🏻

berland (approved):

PR #1266, opened 2026-08-14
Remove code deprecated in version-6.3

eivindjahren:
This removes code marked as deprecated in the version-6.3 branch. This is not exhaustive, it only removes the code that was simple to remove now.


ajaust (approved):
🔪

PR #1270, opened 2026-08-17
Remove TimeVector

eivindjahren:


tests/rd_tests/test_rd_sum.py
```diff
@@ -1690,8 +1713,8 @@ def test_that_time_range_discards_time_of_day_for_explicit_start_and_end():
         datetime.datetime(2000, 5, 1, 0, 0),
     ]
     assert [t.datetime() for t in daily][:2] == [
-        datetime.datetime(2000, 3, 16, 0, 0),
-        datetime.datetime(2000, 3, 26, 0, 0),
+        datetime.datetime(2000, 3, 16, 7, 30),
+        datetime.datetime(2000, 3, 26, 7, 30),
```
ajaust, line 1733: Has the result changed in this commit or was the value in the assertion wrong before? Is the bisectability of the repository affected by this?

python/resdata/summary/rd_sum.py
```diff
@@ -1485,3 +1479,49 @@ def _parse_time_unit(
...
+
+    if time_unit == "y":
+        year += num
+    else:
+        month += num - 1
+        delta_year, new_month = divmod(month, 12)
+        month = new_month + 1
+        year += delta_year
+
+    return datetime.datetime(year, month, day, hour, minute, second)
```
ajaust: Not an issue for the current code because `time_range` does some sanitization by setting the `day` to `1`  if we increment by year or month.

Are we ok with this method failing if the caller input will lead to an invalid date or should we add a safeguard?

### Examples
- Increment to non-existing date (31st of February does not exist)
```
_next_time(datetime.datetime(2000, 1, 31), 1, "m")
```
- Increment to non-existing date (29th of February does not exist in 2001)
```
_next_time(datetime.datetime(2000, 2, 29), 1, "y")
```

ajaust (commented):
Nice.

I have two comment/questions.

python/resdata/summary/rd_sum.py
eivindjahren: Fixed it. It uses python-dateutils, which is already a transient dependency (via pandas), but now it becomes a dependency.

ajaust (approved):

PR #1271, opened 2026-08-18
Remove DoubleVector

eivindjahren:


lib/CMakeLists.txt
```diff
@@ -13,15 +13,15 @@ endif()
 # Make cmath have defines like M_PI
 add_compile_definitions(_USE_MATH_DEFINES)
 
-foreach(type int double long float)
+foreach(type int)
```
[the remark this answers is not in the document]
eivindjahren: These were unused and we don't consider the C API public.

lib/CMakeLists.txt
```diff
@@ -13,15 +13,15 @@ endif()
...
   set(TYPE ${type})
   set(SIGNED_TYPE true)
   configure_file(vector_template.hpp.in include/ert/util/${type}_vector.hpp)
   configure_file(util/vector_template.cpp.in ${type}_vector.cpp)
   list(APPEND opt_srcs ${CMAKE_CURRENT_BINARY_DIR}/${type}_vector.cpp)
 endforeach()
 
-foreach(type bool size_t)
+foreach(type bool)
```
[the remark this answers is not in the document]
eivindjahren: These were unused and we don't consider the C API public.

lib/resdata/rd_sum.cpp
```diff
@@ -710,40 +711,40 @@ rd_sum_ptr rd_sum_alloc_resample(const rd_sum_type *rd_sum, const char *rd_case,
...
-    auto data =
-        make_double_vector(rd_sum_vector_get_size(rd_sum_vector.get()), 0);
+    int num_nodes = rd_smspec_num_nodes(rd_sum->smspec.get());
+    std::vector<double> data{};
 
     int report_step = 0;
     for (auto it = times_begin; it != times_end; ++it, ++report_step) {
         time_t input_t = *it;
         if (input_t < start_time) {
+            data.resize(num_nodes, 0.0);
```
ajaust, line 720: Do we need the overallocation by one elemen or could we resize to `num_nodes-1`? We allocate for `num_nodes` entries, but only write at `[i-1]` for `1 <= i < num_nodes`. This means, we never write the last entry.

lib/resdata/rd_sum_data.cpp
```diff
@@ -734,37 +734,30 @@ void rd_sum_data_fwrite_interp_csv_line(const rd_sum_data_type *data,
...
-  rd_sum_data_get_interp_vector( data, sim_time, keys, results);
-
-*/
-
-void rd_sum_data_get_interp_vector(const rd_sum_data_type *data,
-                                   time_t sim_time,
-                                   const rd_sum_vector_type *keylist,
-                                   double_vector_type *results) {
+/** If the keylist contains invalid indices the corresponding element in the
+  results vector will be zero. */
```
ajaust: Is this correct? I thought the element in the results vector will be `missing_value` if an index is invalid.

ajaust (commented):
Nice! Just some smaller comments from my side.

lib/resdata/rd_sum.cpp
eivindjahren: This is a lot simpler at least as we would otherwise have to deal with num_nodes = 0.

ajaust (approved):

PR #1273, opened 2026-08-19
 VectorTemplate, IntVector and PermutationVector

eivindjahren:
This removes VectorTemplate, IntVector and PermutationVector which is deprecated in version 6.3.


lib/resdata/rd_kw_pybind.cpp
```diff
@@ -266,12 +265,23 @@ PYBIND11_MODULE(_kw, m) {
...
+           py::array_t<int, py::array::c_style | py::array::forcecast> actnum) {
+            if (nx < 0 || ny < 0 || nz < 0)
+                throw std::invalid_argument(
+                    "nx, ny and nz must be non-negative");
+
+            auto required_size = static_cast<py::ssize_t>(nx) * ny * nz;
+            if (actnum.size() < required_size)
+                throw std::invalid_argument(fmt::format(
+                    "actnum has {} elements, but nx*ny*nz={} were expected",
+                    actnum.size(), required_size));
```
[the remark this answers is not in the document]
eivindjahren: This is fine, py::ssize_t is 64 bit.

lib/resdata/fault_block.cpp
```diff
@@ -22,45 +20,30 @@ FaultBlock::FaultBlock(fault_block_layer_type *parent_layer, int block_id)
...
     if (!valid_center) {
         double new_xc = 0;
         double new_yc = 0;
 
-        for (int index = 0; index < int_vector_size(i_list.get()); index++) {
-            int i = int_vector_iget(i_list.get(), index);
-            int j = int_vector_iget(j_list.get(), index);
+        for (size_t index = 0; index < i_list.size(); index++) {
+            int i = i_list[index];
+            int j = j_list[index];
```
ajaust: I am unsure if we are at risk of an out-of-bounds access due to how `add_cell` works.

In `add_cell` we have the following:

```c++
    i_list.push_back(i);
    j_list.push_back(j);
    global_index_list.push_back(rd_grid_get_global_index3(grid, i, j, k));
```
If `i_list.push_back(i)` succeeds, but `j_list.push_back(j)` fails, `j_list` will hold one fewer element than `i_list` and we have an out-of-bounds access. 

Do you think this is possible or will the application always (safely) abort if a memory allocation in `push_back` fails?

ajaust (commented):
What is the purpose of this PR? The subject mentions the removal of `IntVector` and `bool_vector_type`, but the body says 

> This removes VectorTemplate, IntVector and PermutationVector which is deprecated in version 6.3.

which does not mention `bool_vector_type` at all. I also still find the `bool_vector_type`:
```text
$ git branch
  fix_build
  knowledge
  main
  python-trusted-publishing
* remove_bool_vector_type
  remove_double_vector
  remove_time_t_vector
  use-uv-with-lockfile

$ rg bool_vector_type
lib/util/type_vector_functions.cpp
7:int_vector_type *bool_vector_alloc_active_list(const bool_vector_type *mask) {
18:bool_vector_type *int_vector_alloc_mask(const int_vector_type *active_list) {
19:    bool_vector_type *mask = bool_vector_alloc(0, false);

lib/util/string_util.cpp
84:        bool_vector_type *mask = int_vector_alloc_mask(active_list);
116:                                    bool_vector_type *active_mask) {
131:bool_vector_type *string_util_alloc_active_mask(const char *range_string) {
132:    bool_vector_type *mask = bool_vector_alloc(0, false);

lib/util/vector_template.cpp.in
1185:   bool_vector_type   * v2;

lib/util/tests/ert_util_string_util.cpp
29:static void test2(const bool_vector_type *active_mask) {
46:    bool_vector_type *active_mask =

lib/util/tests/ert_util_type_vector_functions.cpp
14:        bool_vector_type *mask = int_vector_alloc_mask(index_list);
29:    bool_vector_type *mask = bool_vector_alloc(0, false);

lib/include/ert/util/type_vector_functions.hpp
11:int_vector_type *bool_vector_alloc_active_list(const bool_vector_type *mask);
12:bool_vector_type *int_vector_alloc_mask(const int_vector_type *active_list);

lib/include/ert/util/string_util.hpp
13:                                    bool_vector_type *active_mask);
14:bool_vector_type *string_util_alloc_active_mask(const char *range_string);
```

eivindjahren:
> What is the purpose of this PR? The subject mentions the removal of `IntVector` and `bool_vector_type`, but the body says
> 
> > This removes VectorTemplate, IntVector and PermutationVector which is deprecated in version 6.3.

I appologize for the confusion the bool_vector_type removal was moved to the "Remove unused utilities" PR in the stack, and I didn't notice the title was out of date.

lib/resdata/fault_block.cpp
eivindjahren: You are right, I think the right course of action is to switch to a list of tuple. I added a commit that I will squash with the fault_block commit which addresses this.

lib/resdata/fault_block.cpp
```diff
@@ -26,8 +26,7 @@ void FaultBlock::add_cell(int i, int j) {
     if (this->is_detached())
         throw std::invalid_argument(
             "Cannot add a cell on a detached fault block");
-    i_list.push_back(i);
-    j_list.push_back(j);
+    index_list.emplace_back(i, j);
     global_index_list.push_back(rd_grid_get_global_index3(grid, i, j, k));
```
ajaust: Not sure if this is a concern, but `global_index_list` may fail to allocate during the `push_back`. In this case `index_list` was extended, but `global_index_list` misses some entry.

ajaust (commented):
Thanks for the explanation. 👍 

I have one more question about some potential inconsistency, but that's it.

lib/resdata/fault_block.cpp
eivindjahren: Fixed it by calculating global_index_list when it is needed.

ajaust (approved):
Looks good. 

I guess we need to remove `IntVector` now from `res-compat` as well (see https://github.com/equinor/res-compat/blob/main/src/ecl/util/util/__init__.py).

PR #1275, opened 2026-08-19
Remove unused utilities

eivindjahren:


lib/util/util.cpp
ajaust, line 794: I think the declaration of this method was removed from the header. If the method is not used internally, we can probably delete it  as well.

ajaust (approved):
Looks good. I found the implementation of one method that you might be able to delete since its declaration was removed from `util.hpp`.

PR #1280, opened 2026-08-25
Avoid some usages of int_vector_type & vector_type

eivindjahren:
The goal of this PR is to make some progress towards not using internal vector types. I needed to resolve some bugs and design issues before that is possible.

This resolves severals bugs related to type conversions and having bad states representable:

1. parser.cpp would alias EOF by not using int as the return type for fgetc.
2. smspec would not always switch between representations of time
3. There was a long standing regression with select_box

### DIFF
--- a/lib/include/resdata/nnc_info.hpp
+++ b/lib/include/resdata/nnc_info.hpp
@@ -1,25 +1,16 @@
-#ifndef ERT_NNC_INFO_H
-#define ERT_NNC_INFO_H
-
-#include <ert/util/type_macros.hpp>
+#pragma once
+#include <cstddef>
+#include <vector>
+#include <memory>
 
 #include <resdata/nnc_vector.hpp>
 
 typedef struct nnc_info_struct nnc_info_type;
 
-#ifdef __cplusplus
-#include <vector>
 const std::vector<int> &
 nnc_info_get_grid_index_list(const nnc_info_type *nnc_info, int lgr_nr);
 const std::vector<int> &
 nnc_info_get_self_grid_index_list(const nnc_info_type *nnc_info);
-#endif
-
-#ifdef __cplusplus
-extern "C" {
-#endif
-
-UTIL_IS_INSTANCE_HEADER(nnc_info);
 
 nnc_info_type *nnc_info_alloc(int lgr_nr);
 void nnc_info_free(nnc_info_type *nnc_info);
@@ -34,17 +25,11 @@ nnc_vector_type *nnc_info_get_vector(const nnc_info_type *nnc_info, int lgr_nr);
 nnc_vector_type *nnc_info_get_self_vector(const nnc_info_type *nnc_info);
 
 int nnc_info_get_lgr_nr(const nnc_info_type *nnc_info);
-int nnc_info_get_size(const nnc_info_type *nnc_info);
+size_t nnc_info_get_size(const nnc_info_type *nnc_info);
 int nnc_info_get_total_size(const nnc_info_type *nnc_info);
 bool nnc_info_equal(const nnc_info_type *nnc_info1,
                     const nnc_info_type *nnc_info2);
 nnc_info_type *nnc_info_alloc_copy(const nnc_info_type *src_info);
 bool nnc_info_has_grid_index_list(const nnc_info_type *nnc_info, int lgr_nr);
 
-#ifdef __cplusplus
-}
-#include <memory>
-
 using nnc_info_ptr = std::unique_ptr<nnc_info_type, decltype(&nnc_info_free)>;
-#endif
-#endif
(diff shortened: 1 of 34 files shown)


lib/resdata/nnc_info.cpp
```diff
@@ -83,76 +61,63 @@ bool nnc_info_equal(const nnc_info_type *nnc_info1,
...
 
             while (true) {
-                nnc_vector_type *vector1 =
-                    nnc_info_get_vector(nnc_info1, lgr_nr);
-                nnc_vector_type *vector2 =
-                    nnc_info_get_vector(nnc_info2, lgr_nr);
+                const nnc_vector_type *vector1 =
+                    nnc_info_get_vector(nnc_info1, static_cast<int>(lgr_nr));
+                const nnc_vector_type *vector2 =
+                    nnc_info_get_vector(nnc_info2, static_cast<int>(lgr_nr));
```
ajaust, line 74: Is there any chance/risk that either `nnc_info1->lgr_index_map.size()` or `nnc_info2->lgr_index_map.size()` are larger than `numeric_limits<int>::max()` which would break the `static_cast` to `int`?

lib/include/resdata/nnc_info.hpp
```diff
@@ -34,17 +25,11 @@ nnc_vector_type *nnc_info_get_vector(const nnc_info_type *nnc_info, int lgr_nr);
 nnc_vector_type *nnc_info_get_self_vector(const nnc_info_type *nnc_info);
 
 int nnc_info_get_lgr_nr(const nnc_info_type *nnc_info);
-int nnc_info_get_size(const nnc_info_type *nnc_info);
+size_t nnc_info_get_size(const nnc_info_type *nnc_info);
```
ajaust, line 28: 1. This method does not seem to be used outside of tests, but I assume it is part of the public interface so we should keep it.
2. There is a code example in a comment using `nnc_info_get_size` https://github.com/equinor/resdata/blob/2020f7614c2f98c8533c565281404b731b28a401/lib/resdata/rd_grid.cpp#L358 that uses `int` in the `for`-loop. This should be changed to `size_t`

lib/resdata/rd_util.cpp
```diff
@@ -733,45 +722,33 @@ bool rd_fmt_file(const char *filename, bool *__fmt_file) {
...
+            std::toupper(static_cast<unsigned char>(month_name_[i])));
+    upper_month[month_len] = '\0';
+
+    static const std::
+        unordered_map<std::string_view, int>
+            month_map =
+                {
+                    {"JAN", 1},  {"FEB", 2},  {"MAR", 3},  {"APR", 4},
+                    {"MAI", 5},  {"MAY", 5},  {"JUN", 6},  {"JUL", 7},
+                    {"JLY", 7}, /* ECLIPSE ambiguous on July. */
```
ajaust: I am wondering if we should keep the comment
```text
/* ECLIPSE ambiguous on July. */
```
I don't understand why it was originally made. My first reaction was that July must have more than one label (`JLY` and `JUL`). However, this applies to other months as well (May, October, December).

Is July somewhat special in the context of Eclipse in a different way than May, October and December (which may simply use Norwegian naming instead of Eclipse naming).

lib/resdata/rd_util.cpp
```diff
@@ -733,45 +722,33 @@ bool rd_fmt_file(const char *filename, bool *__fmt_file) {
...
-    return month_nr;
+static int rd_get_month_nr__(const char *month_name_) {
+    constexpr size_t month_len = 3;
+
+    /* Guard against reading past the end of short/malformed input before
+       touching any of its bytes. */
+    if (strnlen(month_name_, month_len) != month_len)
+        return -1;
+
+    char upper_month[month_len + 1];
```
ajaust: I am just wondering if we could simplify the method by using `std::string` instead of `char[]` for `upper_month`. At the moment there is some manual fiddling  (adding null-byte, slicing to remove the null byte again etc.) which looks like it is adding noise without any benefit.

ajaust (commented):
The change look good. I think it would be great if we could add some tests to the parser if this is not too much work (and out of scope).

lib/resdata/nnc_info.cpp
eivindjahren: Well, this is not making it worse. I think the odds are low as this would require a file that is larger than 64gb I think.

lib/resdata/rd_util.cpp
eivindjahren: I am not sure, I kept it as is. I think that is reasonable.

lib/include/resdata/nnc_info.hpp
eivindjahren: Fixed

ajaust (approved):

PR #1281, opened 2026-08-26
Remove resdataprototype

eivindjahren:


python/resdata/util/util/ctime.py
ajaust, line 3: Seems to be unused now.

python/resdata/__init__.py
```diff
@@ -41,64 +10,6 @@
...
-    Set callback function for util_abort, which is called prior to std::abort()
-    """
-    global _abort_handler
-    _abort_handler = function
-
-    ResdataPrototype.lib.util_set_abort_handler(_c_abort_handler)
-
-
-class ResdataPrototype(Prototype):
-    lib = _dlopen_resdata()
```
ajaust, line 96: I assume that this did some magic w.r.t. loading the library. After removing it, this became an issue for the Windows build.

It looks like we can fix it by updating our cibuildwheel config for window, see https://github.com/ajaust/resdata/pull/1/changes/fa4fc64ffd9e111028716eb44624d244df9e96ea. By telling delvewheel where `libresdata.dll` is instead of skipping it (since we could not locate it earlier), delvewheel seems to be able to do some magic and stuff works, see https://github.com/ajaust/resdata/actions/runs/33078634167/job/98539535465?pr=1 (I wrote the message when py311 @ windows passed successfully assuming that the rest will work).

ajaust (commented):
Good! Some small comments and suggestions for fixing the Windows build.

ajaust (approved):

PR #1283, opened 2026-08-27
Resolve an object lifetime bug in add_polyline

eivindjahren:


lib/geometry/geo_polygon.cpp
```diff
@@ -28,25 +24,20 @@ struct geo_polygon_struct {
...
         polygon = new geo_polygon_struct(std::nullopt);
 
-    UTIL_TYPE_ID_INIT(polygon, GEO_POLYGON_TYPE_ID);
     return polygon;
 }
 
 void geo_polygon_free(geo_polygon_type *polygon) { delete polygon; }
 
 void geo_polygon_free__(void *arg) {
-    geo_polygon_type *polygon = geo_polygon_safe_cast(arg);
```
ajaust, line 48: Typo? The commit message refers to "safe case", but we remove a "save cast".

lib/geometry/geo_polygon.cpp
```diff
@@ -221,65 +179,46 @@ static XLinesStatus xsegments(const std::array<std::array<double, 2>, 4> points,
...
         return 0;
     else {
         double length = 0;
-        double x0 = polygon->xcoord.at(0);
-        double y0 = polygon->ycoord.at(0);
-
-        for (size_t i = 1; i < polygon->xcoord.size(); i++) {
-            double x1 = polygon->xcoord.at(i);
-            double y1 = polygon->ycoord.at(i);
+        auto [x0, y0] = points[0];
```
ajaust: If `length()` is called on an empty Polygon we access the `points` vector even though it is empty.

```C++
rd::Polygon polygon;
std::cout << polygon.length() << std::endl; // Out-of-bounds access to points[0], but points.size() is 0
```

tests/geometry_tests/test_cpolyline_collection.py
```diff
@@ -155,11 +155,10 @@ def test_that_creating_two_polylines_with_the_same_name_raises_key_error():
         collection.createPolyline(name="border")
 
 
-def test_that_a_reference_polyline_can_be_added_to_another_collection():
+def test_that_a_polyline_from_one_collection_can_be_added_to_another_collection():
```
ajaust, line 158: Can we add any tests that would trigger the lifetime bug to ensure that it is not introduced again in the future?

tests/geometry_tests/test_cpolyline.py
```diff
@@ -211,7 +211,7 @@ def test_that_unzip2_returns_separate_x_and_y_lists():
 def test_that_getitem_with_non_integer_index_raises_type_error():
     polyline = CPolyline(init_points=[(0, 0), (1, 1)])
 
-    with pytest.raises(TypeError, match="Index argument must be integer"):
+    with pytest.raises(TypeError):
```
ajaust, line 214: Why can't we match against the error message anymore?

ajaust (commented):
Nice work. I left some comments.

lib/geometry/geo_polygon.cpp
eivindjahren: Fixed

tests/geometry_tests/test_cpolyline.py
eivindjahren: We can, but its the automatically generated one by pybind11 now. I thought it would be best to not depend on that text being exactly the same.

tests/geometry_tests/test_cpolyline_collection.py
eivindjahren: I updated the test to be a regression test.

ajaust (approved):

PR #1284, opened 2026-08-28
Remove copen

eivindjahren:
This removes all usage of cwrap.open and replace it with python code. We go by a conversion to fd for testing purposes only.

Also replaces CFILE-based GRDECL read/write paths (and removes related C/C++/pybind APIs) with python.


python/resdata/grid/rd_grid.py
```diff
@@ -1046,10 +1091,10 @@ def write_grdecl(self, rd_kw, pyfile, special_header=None, default_value=0):
...
         """
-
-        if len(rd_kw) == self.get_num_active() or len(rd_kw) == self.get_global_size():
-            cfile = CFILE(pyfile)
-            _grid._fwrite_grdecl(self, rd_kw, special_header, cfile, default_value)
+        if len(rd_kw) == self.get_global_size():
+            full_kw = rd_kw
+        elif len(rd_kw) == self.get_num_active():
+            full_kw = self._scatter_to_global(rd_kw, default_value)
         else:
```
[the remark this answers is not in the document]
eivindjahren: This is an intended change

python/resdata/resfile/rd_kw.py
```diff
@@ -189,37 +270,71 @@ def read_grdecl(
...
+        if rd_type is not None:
+            if not isinstance(rd_type, ResDataType):
+                raise TypeError("Expected ResDataType, was: %s" % type(rd_type))
+
+            if rd_type not in [ResDataType.RD_FLOAT, ResDataType.RD_INT]:
+                raise ValueError(
+                    "The type:%s is invalid when loading keyword:%s"
+                    % (rd_type.type_name, kw)
+                )
+
```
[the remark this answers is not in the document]
eivindjahren: This exact check was in the previous version of the code.

python/resdata/resfile/rd_kw.py
```diff
@@ -24,14 +24,95 @@
...
+def _parse_grdecl_scalar(token: str, rd_type: ResDataType):
+    """Parses @token as a single scalar value of @rd_type."""
+    if rd_type.is_int():
+        if _GRDECL_INT_TOKEN_RE.match(token):
+            return int(token)
+    else:
+        if _GRDECL_FLOAT_TOKEN_RE.match(token):
+            return float(token)
+    return None
```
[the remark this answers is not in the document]
eivindjahren: This is  pre-existing quirk. It will never effect parsing as the only valid types for parsing are float and int.

python/resdata/resfile/rd_kw.py
```diff
@@ -242,9 +357,56 @@ def fseek_grdecl(cls, fileH, kw, rewind=False):
...
+        at EOF on failure.
+
+        Only the first whitespace-delimited token of a line is ever
+        considered a candidate keyword; if it does not match, the rest of
+        the line is discarded. Lines whose first token starts with "--"
+        are comments and are skipped, as are blank lines.
+        """
+        while True:
+            line_start = fileH.tell()
+            line = fileH.readline()
```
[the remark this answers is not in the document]
eivindjahren: This is an intended change.

python/resdata/resfile/rd_kw.py
```diff
@@ -24,14 +24,95 @@
...
+    match = _GRDECL_MULTIPLIER_RE.match(token)
+    if match:
+        count_str, value_str = match.groups()
+        value = _parse_grdecl_scalar(value_str, rd_type)
+        if value is None:
+            return None
+        count = int(count_str)
+        if count <= 0:
+            return None
+        return [value] * count
```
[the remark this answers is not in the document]
eivindjahren: This is problem existed before this change, and we could try to detect an OOM situation, but we already do that sufficiently through pythons list constructor.

python/resdata/summary/rd_sum.py
```diff
@@ -1399,15 +1401,13 @@ def __repr__(self):
...
         Will dump a csv formatted line of the keywords in keywords, evaluated
-        at the interpolated time. pfile should point to an open Python
-        file handle.
+        at the interpolated time.
         """
-        cfile = CFILE(pfile)
-        ctime = CTime(time)
-        _rd_sum._dump_csv_line(self, ctime.ctime(), keywords, cfile)
+        values = self.get_interp_row(keywords, time, invalid_value=math.nan)
+        pfile.write(",".join("" if math.isnan(v) else "%f" % v for v in values))
```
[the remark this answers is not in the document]
eivindjahren: I think the current behavior is reasonable.

python/resdata/resfile/rd_kw.py
```diff
@@ -987,11 +1171,39 @@ def write_grdecl(self, file):
...
+            columns = self._GRDECL_COLUMNS_BOOL
+        elif data_type.is_char() or data_type.is_string():
+            blocksize = self._GRDECL_BLOCKSIZE_CHAR
+            columns = self._GRDECL_COLUMNS_CHAR
+        else:
+            raise ValueError(
+                "Internal inconsistency: message type keywords should not have data"
+            )
+
+        _grdecl_write_tokens(file, _kw._format_grdecl_tokens(self), blocksize, columns)
```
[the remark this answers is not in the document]
eivindjahren: Unfortunately the streamed version is about 2.1x slower, but I changed the std::vector<std::string> to a version that builds the py::list directly without copying.

lib/resdata/cwrap_pybind.cpp
```diff
@@ -37,6 +49,131 @@ template <typename T> T *cast_cwrap(py::handle obj) {
...
+#else
+    ::lseek(fd, pos, SEEK_SET);
+#endif
+}
+
+/// Validates that @fd is open with access permissions compatible with @mode.
+void check_fd_mode(int fd, const char *mode) {
+#ifdef _WIN32
+    if (_get_osfhandle(fd) == -1)
+        throw py::value_error("Invalid file descriptor: " + std::to_string(fd));
```
ajaust: Why do we not check the `mode` settings on Windows? We seem to only check the that the file descriptor is in a "good" state and nothing more.

lib/resdata/rd_grid_pybind.cpp
```diff
@@ -1,4 +1,5 @@
 #include <optional>
+#include <stdio.h>
```
ajaust: Do you know why we need `stdio.h` in this (and some of the other) binding files? I don't see any obvious use of any types or methods provided by `stdio`. We seem to only rely on standard types (int, `"w"` etc.) and our own new `FdStream` which lives in `cwrap_pybind`.

lib/resdata/rd_kw_pybind.cpp
```diff
@@ -3,6 +3,7 @@
 
 #include <stdexcept>
 #include <optional>
+#include <stdio.h>
```
ajaust: Include of `stdio` (see comment above)

lib/resdata/rd_sum_pybind.cpp
```diff
@@ -2,6 +2,7 @@
 #include <ctime>
 #include <optional>
 #include <stdexcept>
+#include <stdio.h>
```
ajaust: Include of `stdio` (see comment above)

lib/include/resdata/rd_kw.hpp
```diff
@@ -47,6 +50,99 @@ size_t rd_kw_fortio_size(const rd_kw_type *rd_kw);
...
+    } else {
+        arg_x = 0.0;
+        pow_x = 0.0;
+    }
+    return {arg_x, static_cast<int>(pow_x)};
+}
+
+inline size_t format_kw_element_buf(char *buf, size_t buf_size, int value) {
+    int written = std::snprintf(buf, buf_size, " %11d", value);
+    if (written < 0)
```
ajaust, line 76: Do we need to check for truncation? From what I understand, `snprintf` could return `written >= buf_size` which would indicate that something was truncated.

lib/resdata/rd_kw_pybind.cpp
```diff
@@ -64,11 +64,80 @@ class KwIterator {
...
+            size_t len = rd::format_kw_element_buf(buf, sizeof(buf),
+                                                   rd_kw_iget_double(kw, i));
+            tokens[i] = py::str(buf, len);
+        }
+        break;
+    }
+    case RD_BOOL_TYPE:
+        for (int i = 0; i < size; i++)
+            tokens[i] =
+                rd_kw_iget_bool(kw, i) ? py::str("  T") : py::str("  F");
```
ajaust, line 104: Why are we not using `format_kw_element(bool)` here, but we use `format_kw_element` in the other cases.

python/resdata/resfile/rd_kw.py
```diff
@@ -25,14 +25,79 @@
...
+    match = _GRDECL_MULTIPLIER_RE.match(token)
+    if match:
+        count_str, value_str = match.groups()
+        value = _parse_grdecl_scalar(value_str, rd_type)
+        if value is None:
+            return None
+        count = int(count_str)
+        if count <= 0:
+            return None
+        return [value] * count
```
ajaust, line 68: Is it realistic to exhaust the memory here due to large values of `count`? Do we need some kind of safeguard?

python/resdata/resfile/rd_kw.py
```diff
@@ -227,12 +292,45 @@ def read_grdecl(
...
+        if values:
+            result.numpy_view()[:] = values
+        return result
+
+    @staticmethod
+    def _read_grdecl_values(tokenizer: _GrdeclTokenizer, header, rd_type, strict):
+        values = []
+        while True:
+            token = tokenizer.next_token()
+            if token is None or token == "/":
```
ajaust, line 317: What kind special meaning has the `/` character? We stop reading values as soon as we find the character.

python/resdata/summary/rd_sum.py
```diff
@@ -1406,9 +1406,8 @@ def dump_csv_line(self, time: TimeLike, keywords, pfile: TextIOWrapper):
         Will dump a csv formatted line of the keywords in keywords, evaluated
         at the interpolated time.
         """
-        ctime = CTime(time)
-        with synced_fd(pfile) as fd:
-            _rd_sum._dump_csv_line(self, ctime.ctime(), keywords, fd)
+        values = self.get_interp_row(keywords, time, invalid_value=math.nan)
+        pfile.write(",".join("" if math.isnan(v) else "%f" % v for v in values))
```
ajaust, line 1410: Is there any valid use case that would lead to a `nan` value which is now filtered out?

tests/rd_tests/test_grid.py
```diff
@@ -1157,3 +1157,107 @@ def test_rectangular_grid_does_not_have_mapaxes():
...
+
+    transformed = Grid(str(filename), apply_mapaxes=True)
+    untransformed = Grid(str(filename), apply_mapaxes=False)
+
+    assert transformed.get_xyz(global_index=0) != untransformed_reference.get_xyz(
+        global_index=0
+    )
+    assert untransformed.get_xyz(global_index=0) == untransformed_reference.get_xyz(
+        global_index=0
+    )
```
ajaust: This assertion appears trivial. Is there any way that this assertion would ever fail? Can we maybe compare against a hardcoded list of (untransformed) coordinates?

lib/resdata/rd_kw_pybind.cpp
```diff
@@ -64,11 +64,80 @@ class KwIterator {
...
 };
 
+static py::list format_grdecl_tokens(py::object self) {
+    rd_kw_type *kw = from_cwrap<rd_kw_type>(self);
+    int size = rd_kw_get_size(kw);
+    rd_type_enum type = rd_kw_get_type(kw);
+    int string_width =
+        static_cast<int>(rd_type_get_sizeof_iotype(rd_kw_get_data_type(kw)));
+
+    py::list tokens(size);
```
ajaust, line 72: Is it potentially too memory expensive to collect all tokens first before processing further instead of reading token one by one and immediately processing it?

I like the current approach since it is simple to follow, but I have no idea how large the files (how many tokens) can be.

lib/include/resdata/rd_kw.hpp
```diff
@@ -47,6 +50,99 @@ size_t rd_kw_fortio_size(const rd_kw_type *rd_kw);
...
+        throw std::runtime_error(
+            "snprintf failed formatting char/string rd_kw element");
+    return static_cast<size_t>(written);
+}
+
+inline std::string format_kw_element(int value) {
+    char buffer[32];
+    size_t len = format_kw_element_buf(buffer, sizeof(buffer), value);
+    return std::string(buffer, len);
```
ajaust, line 118: This will read past the buffer length if `snprintf` inside `format_kw_element_buf` has truncated the writing. This is connected to the comment above about `snprintf`. Not sure if this is a real concern or if we know that the buffer is always large enough.

ajaust (commented):
Nice changes. I have a few comments and questions.

python/resdata/resfile/rd_kw.py
eivindjahren: It could happen, but I think we don't need to safe guard it in applications where this is supposed to be used. I think its better to just assume there is enough memory or crash.

python/resdata/resfile/rd_kw.py
eivindjahren: Yes, this indicates the end of a keyword in a file of many keywords.

python/resdata/summary/rd_sum.py
eivindjahren: I don't think so. I think we handle this better than we used to in any case.

lib/resdata/rd_kw_pybind.cpp
eivindjahren: Yes, I thought it would be best to strike some trade-off between memory usage and speed here.

lib/include/resdata/rd_kw.hpp
eivindjahren: The assumption here is that the buffer is large enough for the given format specifier. I think we could just assume that.

lib/resdata/rd_grid_pybind.cpp
eivindjahren: Ah, yes, apologize. This is due to clangd with the "import-what-you-use" setting thinking that some symbols in stdio.h are not in cstdio. I think fdopen in this case. I'll remove it in the same commit as FdStream.

lib/resdata/rd_kw_pybind.cpp
eivindjahren: This just was just marginally smaller. I think its safe, we are not going to change this behavior.

lib/resdata/rd_kw_pybind.cpp
eivindjahren: Fixed

lib/resdata/cwrap_pybind.cpp
eivindjahren: Unsure, in any case, this is just an intermediate before translating to python where its the python standard librarys problem.

tests/rd_tests/test_grid.py
eivindjahren: Ah, sorry about that, the test went through some versions and ended up meaningless. I have fixed it like you said.

ajaust (approved):
Looks good. 👍 
There are some performance tests failing, but this might be something that we have to accept. I am sure you know more than I do to judge this. 🙂

PR #1285, opened 2026-08-31
Remove custom arrays

eivindjahren:


lib/resdata/rd_coarse_cell.cpp
```diff
@@ -270,14 +264,14 @@ int rd_coarse_cell_get_active_fracture_index(
 */
 int rd_coarse_cell_iget_active_cell_index(
     const rd_coarse_cell_type *coarse_cell, int index) {
-    return int_vector_iget(coarse_cell->active_cells, index);
+    return coarse_cell->active_cells[index];
```
ajaust: I think `int_vector_iget` was bound-checking internally. In this case, `coarse_cell->active_cells[index];` is not equivalent. Consider `->at(index)` if the bound checks are relevant.

lib/resdata/rd_coarse_cell.cpp
```diff
@@ -270,14 +264,14 @@ int rd_coarse_cell_get_active_fracture_index(
...
 }
 
 int rd_coarse_cell_iget_active_value(const rd_coarse_cell_type *coarse_cell,
                                      int index) {
-    return int_vector_iget(coarse_cell->active_values, index);
+    return coarse_cell->active_values[index];
```
ajaust: I think `int_vector_iget` was bound-checking internally. In this case, `coarse_cell->active_cells[index];` is not equivalent. Consider `->at(index)` if the bound checks are relevant.

lib/resdata/rd_coarse_cell.cpp
```diff
@@ -111,12 +114,20 @@ rd_coarse_cell_type *rd_coarse_cell_alloc() {
     coarse_cell->active_index = -1;
     coarse_cell->active_fracture_index = -1;
     coarse_cell->__cell_list_sorted = false;
-    coarse_cell->cell_list = int_vector_alloc(0, 0);
-    coarse_cell->active_cells = int_vector_alloc(0, 0);
-    coarse_cell->active_values = int_vector_alloc(0, 0);
     return coarse_cell;
 }
 
+static void vector_fprintf(std::vector<int> vector, FILE *stream) {
```
ajaust: Could consider

```suggestion
static void vector_fprintf(const std::vector<int>& vector, FILE *stream) {
```

or similar to avoid a copy of the vector.

lib/resdata/well_rseg_loader.cpp
```diff
@@ -11,57 +14,44 @@
...
+        int relative_index = loader->relative_index_map[index];
+        loader->absolute_index_map[index] = relative_index + rseg_offset;
     }
 
-    loader->rst_view->index_fload_kw(loader->kw, 0, index_map, loader->buffer);
+    loader->rst_view->index_fload_kw(loader->kw, 0, loader->absolute_index_map,
+                                     loader->buffer.data());
 
-    return (double *)loader->buffer;
+    return (double *)loader->buffer.data();
```
ajaust: Do you know if this safe? The buffer is a `std::array` and I do not know if it has the same guarantees, for example, w.r.t. memory alignment, as the previous buffer allocation via `malloc`.

ajaust (commented):
I added some comments.

lib/resdata/rd_coarse_cell.cpp
eivindjahren: Yup, sorry, my mistake. Fixed.

lib/resdata/rd_coarse_cell.cpp
eivindjahren: Yup, sorry, my mistake. Fixed.

lib/resdata/rd_coarse_cell.cpp
eivindjahren: fixed

lib/resdata/well_rseg_loader.cpp
eivindjahren: I changed fload to use std::vector<T, N> which is cast to char* which means aliasing is not a problem. Since we use std::byte and std::memcpy (instead of cast) we avoid alignment problems.

ajaust (approved):

PR #1286, opened 2026-09-01
Replace cwrap.BaseCClass

eivindjahren:
First does some minor clean up in the c code, then replaces cwrap.BaseCClass


lib/resdata/FortIO.cpp
```diff
@@ -481,7 +487,7 @@ bool FortIO::read_at_eof() {
 */
 void FortIO::fwrite_error() {
     if (m_writable)
-        util_unlink(m_filename.c_str());
+        std::filesystem::remove(m_filename);
 }
```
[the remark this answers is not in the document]
ajaust: I guess throwing could be acceptable, but we have to ensure that all callers of `fwrite_error` are fine with this.

python/resdata/_base_c_class.py
```diff
@@ -0,0 +1,51 @@
...
+        return self
+
+    def _ad_str(self):
+        return f"at 0x{self.__c_pointer:x}"
+
+    def _create_repr(self, args=""):
+        return f"{self.__class__.__name__}({args}) {self._ad_str()}"
+
+    def __repr__(self):
+        return self._create_repr()
```
ajaust, line 63: `BaseCClass` had different way to determine equality of objects. It [defines an `__eq__`-operator](https://github.com/equinor/cwrap/blob/45392e7eca66af0cba38b7b5ae409cdd944753cf/src/cwrap/basecclass.py#L110-L117) which returns true if the object the wrapper is pointing to is the same. I believe that two `_BaseCClass` instances pointing to the same C object would be considered to be different because they are different Python objects.

Is this change in behavior intended?

python/resdata/_base_c_class.py
```diff
@@ -0,0 +1,51 @@
...
+            new_obj = cls.__new__(cls)
+            _BaseCClass.__init__(
+                new_obj, c_pointer=c_pointer, parent=parent, is_reference=True
+            )
+            return new_obj
+        else:
+            return None
+
+
+__all__ = ["_BaseCClass"]
```
ajaust, line 96: Im unsure about the Copilot comment about the lack of calling `free` in `__del__`. What do you think?

python/resdata/summary/rd_sum_keyword_vector.py
```diff
@@ -28,6 +27,9 @@ def __getitem__(self, index):
     def __len__(self):
         return _rd_sum_keyword_vector._get_size(self)
 
+    def __iter__(self):
+        return iter([self[i] for i in range(len(self))])
```
ajaust: I don't have good overview about how large/long `self` can be so no sure if Copilot's comment is relevant.

python/resdata/_base_c_class.py
```diff
@@ -0,0 +1,51 @@
...
+        if not c_pointer:
+            raise ValueError("Must have a valid (not null) pointer value!")
+
+        if c_pointer < 0:
+            raise ValueError(
+                "The pointer value is negative! This may be correct, but usually is not!"
+            )
+
+        self.__c_pointer = c_pointer
+        self.__parent = parent
```
ajaust, line 34: Which method/who is supposed to use `__parent`? It is only set in `setParent`, but it seems it isn't read.

ajaust (commented):
Nice work.

I seem to have mostly the same comments and questions as Copilot.

lib/resdata/FortIO.cpp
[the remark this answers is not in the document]
eivindjahren: Yes, I believe they are.

python/resdata/_base_c_class.py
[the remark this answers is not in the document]
eivindjahren: Fixed

python/resdata/summary/rd_sum_keyword_vector.py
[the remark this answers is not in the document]
eivindjahren: Fixed

python/resdata/_base_c_class.py
eivindjahren: Right, I fixed it.

python/resdata/_base_c_class.py
eivindjahren: Yes, we definitely need that. Fixed it.

python/resdata/summary/rd_sum_keyword_vector.py
eivindjahren: Not really relevant, but I fixed it.

python/resdata/_base_c_class.py
eivindjahren: Its not really used by anyone, but it ensures that parent is alive while the held reference is alive.

ajaust: Maybe add a comment about this so that nobody deletes the member thinking it is obsolete.

python/resdata/_base_c_class.py
```diff
@@ -0,0 +1,70 @@
...
+        self.__is_reference = is_reference
```
ajaust, line 35: Sorry to come with something else again.

Is `__init__` enough or do we also need to reimplement [`__new__`](https://github.com/equinor/cwrap/blob/45392e7eca66af0cba38b7b5ae409cdd944753cf/src/cwrap/basecclass.py#L38-L42) as well?

The classes derived from `_BaseCClass` have this initializer

https://github.com/equinor/resdata/blob/0da56e9f84e8e0d8e76e11f41da5c84856ab4c4a/python/resdata/grid/faults/layer.py#L12-L17

and only call `__init__` if the `c_ptr` is defined. In the other case, the members of the base class (`__c_pointer`, `__parent`, `__is_reference`) are undefined (?). I am not sure if this could upset Python's garbage collector.

ajaust (commented):
Nice. One minor thing and something that is related to the `_BaseCClass` that I didn't notice before.

python/resdata/_base_c_class.py
eivindjahren: We need to implement __new__, I have fixed this.

ajaust (approved):

eivindjahren:
I checked for leaks in the python tests and found a couple related to thrown exceptions. This was now possible to detect after changing to PyCapsule.

python/resdata/_base_c_class.py
```diff
@@ -0,0 +1,94 @@
...
+        obj = super().__new__(cls)
+        # Set safe defaults before __init__ runs, so that a __del__ call on a
+        # partially constructed object (e.g. because a subclass __init__
+        # raised before calling super().__init__()) does not itself raise an
+        # AttributeError - it will just see a None c_pointer and skip free().
+        obj.__c_pointer = None
+        obj.__parent = None
+        obj.__is_reference = False
+        return obj
+
```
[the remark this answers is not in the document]
eivindjahren: We don't do this.

python/resdata/_base_c_class.py
```diff
@@ -0,0 +1,94 @@
...
+# PyCapsule_GetPointer extracts the raw void* address wrapped by an (unnamed)
+# capsule. Used only for equality/hashing/repr - never to reconstruct a typed
+# pointer for use by C++ code (that goes through cast_cwrap in C++ instead).
+_pycapsule_get_pointer = ctypes.pythonapi.PyCapsule_GetPointer
+_pycapsule_get_pointer.restype = ctypes.c_void_p
+_pycapsule_get_pointer.argtypes = (ctypes.py_object, ctypes.c_char_p)
+
+
+def _capsule_address(capsule):
+    return _pycapsule_get_pointer(capsule, None)
```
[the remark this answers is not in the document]
eivindjahren: Fixed

PR #1287, opened 2026-09-03
Validate `lgr_nr`

ajaust:
**Approach**
Check if `lgr_nr` is negative or too large and reject it. I use `1 000 000` as upper bound, but I don't really know if this is a good upper bound or not. I expect that we would not have more local grid refinements than that.

I think we could also fix it more thoroughly if we would make `lgr_index_map` a `std::unordered_map`, since the comment also mentions

> Cases have popped up where the series of GRIDHEAD(4) values from lgr to lgr have holes :-(

If this is worth it, we could go foor `std::unordered_map` instead. I tried to keep it simple for now since there are other changes lined up that will resolve the issue


eivindjahren (approved):
👍

PR #1288, opened 2026-09-03
Validate RPORV size against active cell count

ajaust:
add_survey_RPORV read data from the RPORV keyword without checking that its size matches the grid's active cell count. Reject the restart file with a clear error if the sizes differ, and add tests covering both oversized and undersized RPORV keywords.


eivindjahren (approved):
:+1:

PR #1289, opened 2026-09-04
Backport improved validation to version 6.3

ajaust:
Recently improved validation of `lgr_nr` and `rporv` backported to version 6.3.

I need feedback if returning `NULL`  instead of using `util_abort` in `rd_grav_survey_alloc_RPORV` is a good idea. It follows the same pattern (return `NULL`) used in `rd_grav_survey_alloc_FIP`.

### DIFF
--- a/lib/resdata/rd_grav.cpp
+++ b/lib/resdata/rd_grav.cpp
@@ -455,18 +455,26 @@ static rd_grav_survey_type *
 rd_grav_survey_alloc_RPORV(rd_grav_type *rd_grav,
                            const rd_file_view_type *restart_file,
                            const char *name) {
-    rd_grav_survey_type *survey =
-        rd_grav_survey_alloc_empty(rd_grav, name, GRAV_CALC_RPORV);
-
-    if (rd_file_view_has_kw(restart_file, RPORV_KW)) {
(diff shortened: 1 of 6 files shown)


lib/resdata/rd_grav.cpp
```diff
@@ -455,19 +455,27 @@ static rd_grav_survey_type *
...
+        rd_file_view_iget_named_kw(restart_file, RPORV_KW, 0);
+    const int active_size = rd_grav->grid_cache->size();
+    const int rporv_size = rd_kw_get_size(rporv_kw);
+    if (rporv_size != active_size) {
+        fprintf(stderr,
+                "%s: %s keyword has %d elements, but the grid has %d "
+                "active cells\n",
+                __func__, RPORV_KW, rporv_size, active_size);
+        return NULL;
+    }
```
ajaust, line 472: Is this reasonable (similar approach as in `rd_grav_survey_alloc_FIP`) or should we rather fail hard using `util_abort`?

eivindjahren: I think this is a reasonable approach.

eivindjahren (approved):

PR #1291, opened 2026-09-10
Migrate build backend to scikit build core

ajaust:
I had a lot of stale builds using the current build setup. This was partially due do the fact that build artifacts where placed in the source tree. I decided to migrate to `scikit-build-core` since the new release of `scikit-build` wrapping around `scikit-build-core` is not release yet so `scikit-build` still lacks some of the features.

Changes/improvements:
- `uv pip install -e .` works
- Fewer python package files (everything moved into `pyproject.toml`). Some extra scripts to set up conan, dynamic metadata etc. moved to `_custom_build`.
- `scikit-build-core` has experimental support for autorebuilds of C/C++ code. This is currently deactivated, but we could give it a try.
- All build artifacts should end up in `_skbuild`. Deleting this directory should remove all old/stale objects and library files.
- Potentially more PEP517 compliant than the old setu,p but I did not check this.


_custom_build/conan.py
```diff
@@ -0,0 +1,73 @@
...
+
+import json
+import os
+import re
+import shutil
+import subprocess
+from pathlib import Path
+
+
+def _detect_ca_cert() -> None:
```
eivindjahren, line 11: I must admit I have forgotten why we need this. Do you know? If so it would be nice to document.

eivindjahren (approved):
I just had a comment about the cert code that was moved. If we can track down why we need it, that would be nice.

eivindjahren:
> scikit-build-core has experimental support for autorebuilds of C/C++ code. This is currently deactivated, but we could give it a try.

How do I try this?

ajaust:
I think adding
```toml
[tool.scikit-build.editable]
# Enables the automatic rebuild on initial import
rebuild = true 
```
to `pyproject.toml` should enable the automatic rebuilds or  pass
```text
-Ceditable.rebuild=true
```
as option to the `uv pip install` command.

Link to the full docs: https://scikit-build-core.readthedocs.io/en/stable/configuration/editable.html

_custom_build/conan.py
ajaust: This method comes from 6296766f0 and tries to simplify handling of packet inspection in corporate networks for Linux/RHEL by adding the machine's certificates to conan's certificate store. It does not have any effect on MacOS since the path `/etc/pki/tls/cert.pem` does not exist. I pushed a commit removing the method. Could you test on your machine if it affects your `resdata` build with `_detect_ca_cert` removed? If this does not have any effect for you, I think we can just remove the method.

ajaust: I just rebase on main and wait for the tests to pass. I will investigate the necessity of `_detect_ca_cert` independently of this PR.

PR #1293, opened 2026-09-15
Add aarch64 linux

pinkwah:
`ubuntu-24.04-arm` is the current latest non-preview github runner image.[^1] There isn't a `ubuntu-latest-arm` sadly.

This PR is for mostly for running Ert on ARM-macOS via Docker w/ Linux, where the best option is to not emulate but run on aarch64 directly.

[^1]: https://github.com/actions/runner-images#available-images


ajaust:
Nice. I think if we put your PR on top of https://github.com/equinor/resdata/pull/1291, the failing checks will work (I have branch using #1291 and your PR [here](https://github.com/ajaust/resdata/pull/4)). I will have a look what is left to get #1291 merged so your change can get merged as well.

ajaust (approved):
Thanks! I just rebased, updated the README. Will merge once the tests are all green.

ajaust (commented):
The tests time out on aarch64. I need to investigate what is going on there.

